PR #2, opened 2021-11-25
feat(layouts): add basic structure for layouts

ismay:
Based on specs here: https://docs.google.com/document/d/1ahBGLrSwHCPaE_jB-_Dmwf4ww6ve_FYqs_eUHg215kA. Follows all the requirements.

- I've taken special care to not use any hardcoded values unless required in the specs (i.e. the sidebar width), everything is completely fluid. So the header expands with content, the footer expands with content, the sidebar expands vertically (and becomes scrollable), and the content area is scrollable in both directions if content overflows. That means that components rendered in the layout can control their own dimensions without having to know about the layout.
- Not using any js for the styling, just for toggling the classes that control the styling.
- I'd initially split up the layout component into composable components, but that seemed counterproductive. First off, I don't think there's a benefit to it, as it doesn't need to be composable. Plus, it would only _seem_ like the components are independent when in fact they are completely coupled. It also increases the api surface area, so a single component with multiple slots for children seemed like the most stable, sane solution (see the multiple slots example from this paragraph: https://reactjs.org/docs/composition-vs-inheritance.html#containment).


src/App.js
```diff
@@ -1,35 +1,24 @@
-import { DataQuery } from '@dhis2/app-runtime'
-import i18n from '@dhis2/d2-i18n'
+import { CssVariables } from '@dhis2/ui'
 import React from 'react'
-import classes from './App.module.css'
+import './App.css'
+import { Layout, Header, Main, Sidebar, Footer } from './layout/index.js'
```
Mohammer5, line 4: I think it'd be better if these names were aligned with the design docs (regardless of internally used elements):
* `Header -> SelectorToolbar`
* `Main -> DataWorkspace / Workspace`
* `Sidebar -> RightPanel`
* `Footer -> BottomBar`

src/App.js
```diff
@@ -1,35 +1,24 @@
...
+        </Header>
+        <Main>
+            <h1>Main</h1>
+        </Main>
+        <Sidebar>
+            <h1>Sidebar</h1>
+        </Sidebar>
+        <Footer>
+            <h1>Footer</h1>
+        </Footer>
```
Mohammer5, line 20: I thought about this a bit. I think, until we have sub-grids, it makes more sense to put the bottom bar into the workspace area. On the one hand, the bottom bar is functionally tightly coupled to the workspace (it's basically the submit buttons), on the other hand is the "style behavior" (the animation when the right-hand panel opens/closes) identical/simultaneously, applying that effect on one container only seems reasonable to me as it's slightly more maintainable and easier to reason about.

ismay:
@Mohammer5 I see you already reviewed parts of this. I'm currently still sketching, I'll let you know when it's ready for review.

src/app.css
```diff
@@ -0,0 +1,3 @@
+body {
+    background: var(--colors-grey100);
```
ismay, line 2: Background color from the specs.

src/app.js
```diff
@@ -0,0 +1,22 @@
...
+    return (
+        <>
+            <CssVariables colors />
+            <Layout
+                header=""
+                main=""
+                sidebar=""
+                footer=""
+                showSidebar
+                showFooter
```
ismay, line 16: State for this can live in App, and setters can be passed down.

src/App.js
ismay: I started out with that, but tbh I found it rather cumbersome. It introduces all kinds of awkward long names, and doesn't immediately make sense when working in the app. I can imagine these names would maybe make sense for the components rendered _in_ the layout, but coupling the layout slots to names from the design seems a bit excessive to me. Doesn't really make the code easier to read or work on for me, the opposite in fact.

src/App.js
ismay: I've just finished the layout component. So with the current layout I've gone for an order in the html that makes sense for the html, and I'm just using grid for positioning/animating. That's also what I've always seen recommended as you don't have to care about source order with grid. So I don't think we have to overthink this, the simplest order seems to me to be just header, content, aside, footer (all on the same level). Then we can just lay that out in css however we want.

ismay:
> @Mohammer5 I see you already reviewed parts of this. I'm currently still sketching, I'll let you know when it's ready for review.

Ok, ready for review now!

src/App.js
ismay: But that's my opinion of course, so let me know if you think it's not a valid line of reasoning.

src/layout/layout.module.css
```diff
@@ -0,0 +1,40 @@
+.wrapper {
+    height: 100%;
+    display: grid;
+    grid-template-rows: auto 1fr;
+    grid-template-columns: 1fr 0;
+    grid-template-areas:
+        'header header'
+        'main   sidebar';
+    transition: grid-template-columns 200ms ease-out;
```
Mohammer5, line 9: Currently this works in FF only, other browser have not implemented the spec yet (see this stackoverflow reply).

@cooper-joe I've talked to @ismay about this and our perspective is to ignore it as it'll be implemented at some point in the future. We don't know when though. If this is crucial functionality, then @ismay could change it and implement a more complicated solution to get this to work.

The current implementation is like this:
```
.layout (display: grid)
  > .header
  > .workspace
  > .rightHandSidebar
  > .bottomBar
```

The alternative solution would look similar to this:
```
.layout (display: flex | grid)
  > header
  > main (display: flex)
    > .workspace (display: flex; flex-direction: columns; flex-grow: 1)
      > .data (flex-grow: 1)
      > .bottomBar (fixed height)
    > .rightHandPanned (fixed width with transition from 0 to fixed and vice versa)
```

ismay: One note is that it _might_ be possible to get the animation with a more complex solution. In trying to get there whilst working on this PR though, I found that it led to other regressions.

ismay: So if it's absolutely crucial we could give it another shot, see if we've overlooked a solution. But if it's not crucial then the existing solution would be a simple and robust way to achieve the specced layout (with the only exception being the lack of animations).

cooper-joe: Animation is a supporting element here, so I don't think it warrants sacrificing the simple, robust layout. The design, user flow, and concept are all still valid without the animation, we just lose a little bit of _smoothness_. But, on balance, I'm OK with this solution.

src/layout/index.js
```diff
@@ -0,0 +1 @@
+export { default as Layout } from './layout.js'
```
Mohammer5, line 1: With #3 being merged, I think we should move `src/layout` to `src/app/layout` as it's being used by the app wrapper only, it's not a shared component (if it was, then it should be in `src/shared/layout`)

Mohammer5 (changes requested):
I guess we're good to go then in terms of functionality, one small change request

src/layout/layout.module.css
ismay: All right, I'll remove the animation and keep the layout as is then. By the way @cooper-joe, there are still other options if we want to make the transition less jarring. We could fade the elements in and out for example. Or use GPU scaling for the size transition. We could talk about that after this PR, I think we could make something work.

PR #3, opened 2021-11-26
feat(file hierarchy): copy & slightly modify file hierarchy from approvals app

Mohammer5:
I've copied the file structure from the approvals app.

* made some minor modifications
  * `approvals-app/src/auth -> data-entry-app/src/app/auth`
    * `auth` is only being used by the app
  * `approvals-app/src/navigation -> data-entry-app/src/shared/navigation`
    * It's shared code, so I don't see why we shouldn't put it there
* copied some things that we'll probably need as well (and can easily delete when not needed)
  * `src/test-utils`
  * `src/shared/error-message`
  * `src/app/auth`
* Some functionality that I copied has to be changed, so I simply removed the approvals-app functionality and left `@TODO` comments

@ismay and I have different opinions on how we should nest.
I've put `auth` into the `app` folder as it's only being used there, from my perspective this makes file tree navigation easier. I assume that @ismay would prefer to have `auth` on the root level and keep everything as flat as possible. I think this is a highly subjective discussion with no right or wrong and I don't mind "bending to the will of the majority" if others see it the same way


Birkbjo:
> I've put auth into the app folder as it's only being used there, from my perspective this makes file tree navigation easier. I assume that @ismay would prefer to have auth on the root level and keep everything as flat as possible. 
I don't really have strong opinions on this, but agree that if this is the only place it is used (and will be as far as we can predict), it makes sense to nest it.

src/shared/navigation/read-query-params.js
```diff
@@ -0,0 +1,11 @@
+import { parse } from 'query-string'
```
Birkbjo, line 1: I think https://www.npmjs.com/package/use-query-params is better for our use-case in this app.

Mohammer5: I think @HendrikThePendric considered it but then switch to this one as the values are needed for constructing/manipulating the central state, which is outside of the react scope. I can investigate this a bit further and will get back to you

PR #5, opened 2021-12-01
feat(layouts): add layout component

ismay:


src/app/auth/use-is-authorized.js
```diff
@@ -1,4 +1,4 @@
-export const useIsAuthorized = () => {
+const useIsAuthorized = () => {
     // @TODO(auth): determine how to fetch/store authorities
```
ismay, line 2: We can take this from the scheduler

Mohammer5 (approved):

PR #6, opened 2021-12-08
feat(context selection): add global state & selector bar

Mohammer5:
This PR adds the context selection bar.
There are some ToDos left which can be addressed post-merge as they're about small details like auto-selection or "default" attribute option combos.

The google docs are up-to-date, so they can be referred to when it comes to behavior. I haven't written any tests or documentation yet, so that's the only source of truth right now. If there's a discrepancy between the code and @cooper-joe's google docs, then that's a bug.

### The code

We've agreed that modules that only have one export should use expose that unit as default export. I like putting the `export` keyword to the unit that's being exported which is why I chose to use the `function` keyword as `export default function ...` is possible and `export default const ....` is not.

I purposefully kept the nesting-depth to 2 so it doesn't get to complex / complicated. Code that's shared between code is hoisted just as much so that every part the needs the code has access to it. If that causes the shared code to be on the root level, it's put into the `src/shared` folder instead to keep the root level as unpolluted as possible.

I copied some code from the approvals-app and didn't pay too much attention to that code as it seems to work already (e.g. the period selector).

**I've tried to keep things self-contained.**
* The selector bar consists of up to 5 items.
* No item "touches" any other item
* If an item needs to react to changes of another item, it "subscribes" to the param in the context selection (which uses the query params)

### PR Follow-up todos (I'll create a jira issue once merged):

* [ ] Discuss all possible texts with @cooper-joe 
* [ ] jest tests
* [ ] developer documentation
* [ ] Data sets
  * [ ] How to handle pages / large lists?
    * Do we want to paginate at all?
* [ ] org unit selector
  * [ ] Only load org units that are connected to the data set & the user has access to


src/app/app.js
```diff
@@ -1,21 +1,31 @@
...
+import { QueryParamProvider } from 'use-query-params'
 import './app.css'
+import { ContextSelection } from '../context-selection/index.js'
 import { Layout } from './layout/index.js'
 
+/**
+ * "use-query-params" requires a router. It suggests react-router-dom in the
+ * docs, so that's why I chose that one. Unfortunately it doesn't work with
+ * react-router-dom@6, so I have do go with react-router-dom@^5
+ */
```
ismay, line 13: You could go with `reach-router` instead, `use-query-params` is compatible with that as well. From the reach docs:

> Reach Router and it’s sibling project React Router are merging as React Router v6. In other words, Reach Router v2 and React Router v6 are the same. There is more information on the maintainers website.
>
> https://reach.tech/router/

So if you can't use the latest react-router due to compatibility issues, I'd look into reach-router, of which you should be to use the latest version. Might result in us having to make less changes later.

README.md
```diff
@@ -1,45 +1,6 @@
...
-
-## Learn More
-
-You can learn more about the platform in the [DHIS2 Application Platform Documentation](https://platform.dhis2.nu/).
-
-You can learn more about the runtime in the [DHIS2 Application Runtime Documentation](https://runtime.dhis2.nu/).
-
-To learn React, check out the [React documentation](https://reactjs.org/).
+In order to provide deep linking, the global state has been implemented with
+`use-query-params`.
```
ismay, line 6: To me the phrasing feels a little unclear, what do you think of:

> We're using `use-query-params` to store part of the app's state in the url query parameters. This allows users to bookmark and share specific states of the app.

That way it's clear why we're doing this.

src/context-selection/README.md
```diff
@@ -0,0 +1,48 @@
+# Context selection
```
ismay, line 1: Having markdown files throughout the code seems a little unconventional to me. Doesn't have to be a bad thing of course. But it seems to me this readme is doing two things:

1. Describing the "contracts", i.e. "this module exports this, does that, etc."
2. Offer some explanation on how things work

I think I'd prefer 1. to be addressed with just unit tests. They should also describe the contracts, plus they will break if anything changes.

And for 2. I think I'd prefer comments in the code. If the code is complex enough that it's hard to comprehend then a comment closeby should help. Having it in the code also makes it easier to update it if the code changes.

For documentation like this I worry that it would get out of date faster than the above solutions would.

src/context-selection/TODOS.md
```diff
@@ -0,0 +1,5 @@
+-   [ ] data set selector
```
ismay, line 1: I think this could go in a PR description or an issue.

src/context-selection/category-option-combo-selector-bar-item/categories-menu.js
```diff
@@ -0,0 +1,76 @@
...
+            </Button>
+        </div>
+    )
+}
+
+CategoriesMenu.propTypes = {
+    close: PropTypes.func.isRequired,
+    // No specific shape due to dynamic keys,
+    // only `{ [categoryId]: categoryOptionId }`
+    selected: PropTypes.object.isRequired,
```
ismay, line 74: If you want you could enforce the categoryOptionId type with:

```
selected: PropTypes.objectOf(PropTypes.string),
 ```

File: src/context-selection/context-selection.js [hunk 6]
```diff
@@ -1,3 +1,32 @@
...
+import { OrgUnitSetSelectorBarItem } from './org-unit-selector-bar-item/index.js'
+import { PeriodSelectorBarItem } from './period-selector-bar-item/index.js'
+import { SectionFilterSelectorBarItem } from './section-filter-selector-bar-item/index.js'
+import { useContextSelection } from './use-context-selection.js'
 
-export default ContextSelection
+export default function ContextSelector() {
+    const [, setSelectionContext] = useContextSelection()
+
+    const onClearSelectionClick = () => {
```
ismay, line 13: Maybe you could create a hook `useClearContextSelection` and colocate that with the definition of the query param schema. Means code outside of that module doesn't need to know about the schema.

src/context-selection/category-option-combo-selector-bar-item/category-option-combo-selector-bar-item.js
```diff
@@ -0,0 +1,94 @@
...
+            nextSelected
+        ).map(([key, value]) => `${key}:${value}`)
+
+        setSelected(nextSelected)
+        setCategoryOptionComboSelection(nextCategoryOptionComboSelection)
+    }
+
+    useDeselectOnDataSetChange(deselectAll)
+    useDeselectOnOrgUnitChange(deselectAll)
+    useDeselectOnPeriodChange(deselectAll)
```
ismay, line 58: Could this not be replaced with something like:

```
useEffect(() => {
  if (selectedNotCleared || categoryOptionComboSelectionNotCleared) {
    setSelected({})
    setCategoryOptionComboSelection([])
  }
}, [dataSetId, periodId, orgUnitId])
```

Current abstraction makes things harder to follow in my opinion.

src/context-selection/category-option-combo-selector-bar-item/category-option-combo-selector-bar-item.js
```diff
@@ -0,0 +1,94 @@
...
+        categoryCombination?.isDefault
+    ) {
+        return null
+    }
+
+    const label = categoryCombination
+        ? categoryCombination?.displayName
+        : errorCategoryCombination
+        ? i18n.t('Category option combination: An error occurred')
+        : i18n.t('Loading categories...')
```
ismay, line 75: Nested ternaries aren't very readable in my opinion. They're terse, but when they're nested it gets easy to misread them.

src/context-selection/category-option-combo-selector-bar-item/category-option-combo-selector-bar-item.js
```diff
@@ -0,0 +1,94 @@
...
+    useDeselectOnOrgUnitChange(deselectAll)
+    useDeselectOnPeriodChange(deselectAll)
+
+    if (
+        !dataSetId ||
+        !periodId ||
+        !orgUnitId ||
+        loadingCategoryCombination ||
+        !categoryCombination?.categories.length ||
+        categoryCombination?.isDefault
```
ismay, line 66: Maybe you could have a helper function that accepts the variables here and returns a boolean, and use that value here. Would be easy to unit test and mean you don't have to test all these states with the entire component. Plus it'd make this part more readable.

src/context-selection/category-option-combo-selector-bar-item/use-selected-category-option-by-category-id.js
```diff
@@ -0,0 +1,9 @@
+import { useCategoryOptionComboSelection } from '../use-context-selection.js'
+
+export default function useSelectedCategoryOptionByCategoryId(id) {
+    const [categoryOptionIds] = useCategoryOptionComboSelection()
+    const selectedValue =
+        categoryOptionIds.find((curId) => curId.startsWith(`${id}:`)) || ''
```
ismay, line 6: Would this not suffice:

```
curId.startsWith(id)
```

src/context-selection/data-set-selector-bar-item/data-set-selector-bar-item.js
```diff
@@ -0,0 +1,49 @@
...
+            {selectableDataSets && (
+                <MenuSelect
+                    values={selectableDataSets}
+                    selected={dataSetId}
+                    onChange={({ selected }) => {
+                        setDataSetId(selected)
+                        setDataSetOpen(false)
+                    }}
+                />
+            )}
```
ismay, line 46: I can imagine you could get some funky output here with stale while revalidate. Previous values won't always be cleared, so it might be possible that you have an error and stale data for example. To prevent that from happening you'd want separate returns for loading, error and data, in that order.

src/context-selection/data-set-selector-bar-item/use-selectable-data-sets.js
```diff
@@ -0,0 +1,29 @@
...
+        params: {
+            paging: false,
+        },
+    },
+}
+
+export default function useSelectableDataSets() {
+    const { loading, error, data } = useDataQuery(QUERY_SELECTABLE_DATA_SETS)
+
+    // @TODO: How to handle pages / large lists?
```
ismay, line 15: If we're going to go in the direction I'm suspecting, we could use this: https://react-query.tanstack.com/guides/paginated-queries

src/context-selection/period-selector-bar-item/year-navigator.js
```diff
@@ -0,0 +1,31 @@
+import { Button, IconArrowRight24, IconArrowLeft24 } from '@dhis2/ui'
+import PropTypes from 'prop-types'
+import React from 'react'
+import classes from './year-navigator.module.css'
+
+// To avoid users from navigating too far back
+const startYear = 1970
```
ismay, line 7: Plus, before unix nothing existed anyway.

src/context-selection/period-selector-bar-item/period-selector-bar-item.js
```diff
@@ -0,0 +1,90 @@
...
+                        onChange={({ selected }) => {
+                            setPeriodId(selected)
+                            setPeriodOpen(false)
+                        }}
+                        periodType={dataSetPeriodType}
+                        year={year}
+                    />
+                </>
+            ) : (
+                <div />
```
ismay, line 86: Would null not suffice here as well?

src/context-selection/use-deselect/use-deselect-on-coc-change.js
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react'
+import { useCategoryOptionComboSelection } from '../use-context-selection.js'
+
+export default function useDeselectOnCocChange(deselectAll) {
+    const [initialExecution, setInitialExecution] = useState(true)
+    const [categoryOptionComboSelection] = useCategoryOptionComboSelection()
+
+    useEffect(() => {
+        if (!initialExecution) {
```
ismay, line 9: Don't you want the opposite here? `if (initialExecution)`. Same for the other hooks on this level.

ismay:
Sorry for the large amount of comments. I tried to keep it confined to things that have a functional impact (or where it seemed like it to me at least), or where I felt it would impact readability. Let me know what you think, hope it helps.

src/context-selection/use-deselect/use-deselect-on-coc-change.js
Mohammer5: The initial/default value is `true`

src/app/app.js
Mohammer5: These projects merged and are being released as react-router v6, which is exactly the one that didn't work

src/context-selection/category-option-combo-selector-bar-item/use-category-combination.js
```diff
@@ -0,0 +1,39 @@
...
+
+    const categoryCombination = data?.dataSet.categoryCombo
+
+    return {
+        calledCategoryCombination: called,
+        loadingCategoryCombination: loading,
+        errorCategoryCombination: error,
+        refetchCategoryCombination: refetch,
+        categoryCombination,
+    }
```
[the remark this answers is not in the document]
Mohammer5: :ok_hand:

src/context-selection/period-selector-bar-item/period-selector-bar-item.js
Mohammer5: No.. Component requires a child

src/context-selection/category-option-combo-selector-bar-item/category-option-combo-selector-bar-item.js
Mohammer5: We talked about this. I changed it a bit, but not as proposed. Feel free to propose another suggestion, I'll mark all other comments talking about this as resolved

src/context-selection/data-set-selector-bar-item/use-selectable-data-sets.js
Mohammer5: This is about whether we want to handle pagination at all.
Maybe large lists is something that's rather unusual in the wild and therefore we can omit pagination.

src/context-selection/README.md
Mohammer5: Agreed, can be done entirely with tests actually

src/context-selection/data-set-selector-bar-item/data-set-selector-bar-item.js
Mohammer5: I'll talk to @cooper-joe about it

ismay:
Rebased on development and did some minor housekeeping.

- Removed unused snapshots
- Added a test wrapper in e257130 that can be used as a wrapper for testing hooks/components that use react-query. It supplies the required queryClient via a provider, but also disables automatic retries and error logging.

src/context-selection/data-set-selector-bar-item/data-set-selector-bar-item.js
Mohammer5: @cooper-joe and I talked about this and agreed on the following:

>  for now it's best to simply not show any data while loading new data / an error occurred. We can still think about how to improve in in the future (either generally or specifically for this app)

ismay: Ok, so then this should be changed right? Because currently it will show data whilst revalidating/loading

ismay: Ah I see you've addressed that with renderMenu. All right

src/context-selection/use-context-selection.js
```diff
@@ -0,0 +1,57 @@
...
+
+export function useOrgUnitId() {
+    return useQueryParam('orgUnitId', PARAMS_SCHEMA.orgUnitId)
+}
+
+export function usePeriodId() {
+    return useQueryParam('periodId', PARAMS_SCHEMA.periodId)
+}
+
+export function useCategoryOptionComboSelection() {
```
Birkbjo, line 30: I think we should call this `attributeOptionCombo`. This is the term we are using when talking about the `categoryOptionCombo` when it's on the dataSet. It's also the property-name in the API when using this value to get/set dataValues. 
It also prevents confusion with `categoryOptionCombos` on dataElements.

Birkbjo (commented):
Looks very nice! Just a minor comment, but it may result in quite a few name-changes if we decide to implement it... So might not be worth it.

src/context-selection/use-context-selection.js
Mohammer5: Done

Birkbjo (approved):

PR #10, opened 2022-01-11
feat: retry mutations and optimistic updates

ismay:
- Enabled retry for cell mutations, which will retry in order once online
- Added optimistic updates for cell datavalue mutation

Follow-up:

- Implement optimistic updates for comments
- Cached mutations should only be removed after a server response for that mutation (i.e. success or server error). Otherwise they should persist (check the retry setting)
- Organize the data fetching hooks (so that the query doesn't change without the optimistic update adapting to that change, should probably colocate them)
- Could integrate the isPaused state for the mutation hook
- Keep only the last mutation for each item
- Add a global indicator for the amount of pending mutations
- Persist the query cache with the optimistic updates and the mutation cache
- Ensure that the optimistic updates won't be overwritten until the related mutation has settled

Docs:

- https://tkdodo.eu/blog/effective-react-query-keys
- https://react-query.tanstack.com/guides/optimistic-updates
- https://react-query.tanstack.com/guides/query-keys
- https://react-query.tanstack.com/reference/QueryClient#queryclientsetqueriesdata
- https://react-query.tanstack.com/guides/query-invalidation
- https://react-query.tanstack.com/guides/invalidations-from-mutations
- https://react-query.tanstack.com/guides/filters#query-filters
- https://react-query.tanstack.com/graphql
- https://github.com/tannerlinsley/react-query/discussions/1626#discussioncomment-938629
- https://github.com/tannerlinsley/react-query/discussions/1500#discussioncomment-229672
- https://github.com/tannerlinsley/react-query/discussions/2597#discussioncomment-1249036
- https://github.com/tannerlinsley/react-query/discussions/2734#discussioncomment-1398792
- https://github.com/tannerlinsley/react-query/discussions/1325
- https://github.com/tannerlinsley/react-query/discussions/1304

### DIFF
--- a/src/data-workspace/use-data-value-mutation.js
+++ b/src/data-workspace/use-data-value-mutation.js
@@ -38,35 +38,42 @@ export const useDataValueMutation = () => {
                 queryClient.getQueryData(dataValueQueryKey)
 
             // Optimistically update to the new value
-            queryClient.setQueryData(dataValueQueryKey, (oldDataValues) => {
-                const newDataValues = { ...oldDataValues }
+            queryClient.setQueryData(dataValueQueryKey, () => {
+                const newDataValues =
+                    previousDataValues.dataValues.dataValues.map(
+                        (dataValue) => {
+                            const {
+                                categoryOptionCombo,
+                                dataElement,
+                                orgUnit,
+                                period,
+                            } = dataValue
+                            const { co, de, ou, pe, value } = newDataValue
+                            const match =
+                                categoryOptionCombo === co &&
+                                dataElement === de &&
+                                orgUnit === ou &&
+                                period === pe
 
-                newDataValues.dataValues.dataValues =
-                    oldDataValues.dataValues.dataValues.map((dataValue) => {
-                        const {
-                            categoryOptionCombo,
-                            dataElement,
-                            orgUnit,
-                            period,
-                        } = dataValue
-                        const { co, de, ou, pe, value } = newDataValue
-                        const match =
-                            categoryOptionCombo === co &&
-                            dataElement === de &&
-                            orgUnit === ou &&
-                            period === pe
+                            if (!match) {
+                                return dataValue
+                            }
 
-                        if (!match) {
-                            return dataValue
+                            return {
+                                ...dataValue,
+                                value,
+                            }
                         }
+                    )
 
-                        return {
-                            ...dataValue,
-                            value,
-                        }
-                    })
-
-                return newDataValues
+                // Ensure we don't mutate previousDataValues
+                return {
+                    ...previousDataValues,
+                    dataValues: {
+                        ...previousDataValues.dataValues,
+                        dataValues: newDataValues,
+                    },
+                }
             })
 
             return { previousDataValues, dataValueQueryKey }
--- a/src/data-workspace/use-data-value-mutation.js
+++ b/src/data-workspace/use-data-value-mutation.js
@@ -1,0 +1,88 @@
+import { useDataEngine } from '@dhis2/app-runtime'
+import { useQueryClient, useMutation } from 'react-query'
+import { useContextSelection } from '../context-selection/index.js'
+import { dataValueQuery, useAttributeOptionCombo } from './data-workspace.js'
+
+const DATA_VALUE_MUTATION = {
+    resource: 'dataValues',
+    type: 'create',
+    params: ({ ...params }) => ({ ...params }),
+}
+
+export const useDataValueMutation = () => {
+    const queryClient = useQueryClient()
+    const [{ dataSetId, orgUnitId, periodId }] = useContextSelection()
+    const attributeOptionComboId = useAttributeOptionCombo()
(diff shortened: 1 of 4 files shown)


src/data-workspace/use-data-value-mutation.js
```diff
@@ -0,0 +1,88 @@
...
+
+            // Snapshot the previous value
+            const previousDataValues =
+                queryClient.getQueryData(dataValueQueryKey)
+
+            // Optimistically update to the new value
+            queryClient.setQueryData(dataValueQueryKey, (oldDataValues) => {
+                const newDataValues = { ...oldDataValues }
+
+                newDataValues.dataValues.dataValues =
```
Mohammer5, line 44: I suppose it doesn't really matter, but this will also update the `oldDataValues.dataValues` property as the referencing the same object

ismay: Yeah I was wondering the same. I'll double check it, because it could affect the snapshot.

ismay: Good thing you commented on this, I'd glossed over it, but this would have caused a bug. The oldDataValues and previousDataValues are both references to the same object. So mutating oldDataValues would have mutated the snapshot and thus broken the rollback. Addressed it with cfdc56c

Mohammer5 (approved):
Code looks good to me, would like @Birkbjo or @KaiVandivier to take look as well though

src/data-workspace/use-data-value-mutation.js
```diff
@@ -0,0 +1,95 @@
...
+                        }
+
+                        return {
+                            ...dataValue,
+                            value,
+                        }
+                    })
+
+                return newDataValues
+            })
```
Mohammer5, line 77:
```suggestion
            queryClient.setQueryData(dataValueQueryKey, (prevSnapshot) => {
                const updatedDataValues =
                    prevSnapshot.dataValues.dataValues.map((dataValue) => {
                        const {
                            categoryOptionCombo,
                            dataElement,
                            orgUnit,
                            period,
                        } = dataValue
                        const { co, de, ou, pe, value } = newDataValue
                        const match =
                            categoryOptionCombo === co &&
                            dataElement === de &&
                            orgUnit === ou &&
                            period === pe

                        if (!match) {
                            return dataValue
                        }

                        return {
                            ...dataValue,
                            value,
                        }
                    })

                return {
                    ...prevSnapshot,
                    dataValues: {
                        ...oldDataValues.dataValues,
                        dataValues: updatedDataValues,
                    },
                }
            })
```

Maybe this simplifies it a bit and reduces the possibility of accidentally overwriting the old snapshot?

ismay: Adapted this a little, see 9f1c5bc321289d6e59f316de76d8c39221d3e7d4

Birkbjo (approved):
Code LGTM to me!

I'm a bit unsure of the behavior and regards to cached mutations though. Do we actually need to rollback the value? And I guess this doesn't happen when you are offline, only when the server rejects or someting?

src/data-workspace/use-data-value-mutation.js
```diff
@@ -0,0 +1,95 @@
...
+        onError: (err, newDataValue, context) => {
+            queryClient.setQueryData(
+                context.dataValueQueryKey,
+                context.previousDataValues
+            )
+        },
+        // Always refetch after error or success
+        // eslint-disable-next-line max-params
+        onSettled: (newDataValue, error, variables, context) => {
+            queryClient.invalidateQueries(context.dataValueQueryKey)
```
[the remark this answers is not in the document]
ismay: So with regards to the refetch & rollback after an error, that's to address this:

> In most of these failure cases, you can just trigger a refetch for your optimistic queries to revert them to their true server state. In some circumstances though, refetching may not work correctly and the mutation error could represent some type of server issue that won't make it possible to refetch. In this event, you can instead choose to rollback your update.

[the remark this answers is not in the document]
ismay: In a similar vein, just keeping the optimistic update on success is a possibility. Syncing with the server after success is just a way to ensure that our data is 100% consistent with the backend, as for example our optimistic updates for new dataValues aren't completely realistic (they lack certain bits of data that we can't predict, like creation time, etc.).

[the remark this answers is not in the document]
ismay: Oh and to answer:

> And I guess this doesn't happen when you are offline, only when the server rejects or someting?

That's correct, mutations will only succeed or fail once the user is online.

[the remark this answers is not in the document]
Birkbjo: I see. I actually think that we don't want to rollback, as we should also integrate this with validations. So to prevent confusing scenarios (eg. multiple mutations in a row, that rollbacks - like you said), I think that showing errors in a good way and let the user fix the error is a better solution.
But yeah, we could try it out and remove it later.

PR #18, opened 2022-01-15
feat: sidebar

mediremi:
Relates to [TECH-856](https://jira.dhis2.org/browse/TECH-856) (sidebar for data values)
Closes [TECH-1198](https://dhis2.atlassian.net/browse/TECH-1198) (adding sidebar), [TECH-857](https://jira.dhis2.org/browse/TECH-857) (contextual help) and [TECH-1129](https://dhis2.atlassian.net/browse/TECH-1129) (Upgrade react-router-dom to v6)

Some todos are still open. I would like to address these in a followup task so others can start using the sidebar already.
The most important things are handling "comments", "limits" and "followup". Except for optimistic updates, these tasks are somewhat trivial and shouldn't take too long. They're not blocking any other work though, which is why I'd like to pull these tasks out of this PR.

## Implemented:

**Unrelated to sidebar**
- [x] Upgrades react router to v6
- [x] Hides periods in the period selector that span into the future

**Sidebar stuff**
- [x] Technical docs
- [x] Options button in top bar
- [x] Show contextual help in sidebar when selected in options dropdown
- [x] Close sidebar when user clicks close button
- [x] [Pass current item to sidebar](https://github.com/dhis2/data-entry-app/blob/b50292a8ee2aec4350b7271d95bc26aeea09b4cc/src/sidebar/sidebar.js#L12)
- [x] Toggle sidebar
   - [x] when user presses shift+enter
- [x] [Remove `CustomDataProvider` from data details sidebar view](https://github.com/dhis2/data-entry-app/blob/b50292a8ee2aec4350b7271d95bc26aeea09b4cc/src/sidebar/data-details/data-details.js#L10-L36)
- [x] Audit log
  + [x] Use `/api/dataEntry/dataValueContext` introduced in https://github.com/dhis2/dhis2-core/pull/10552 (returns audit records and data value history)
  + [x] [Fetch audit log when audit log unit expanded](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/audit-log.js#L11-L17)
  + [x] [Find better way of rendering entries that does not depend on `dangerouslySetInnerHTML`](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/audit-log.js#L19-L59)

## To Dos in followup tasks:
- [ ] Cypress tests
- [ ] Toggle sidebar
   - [ ] when user double clicks on cell -> Skip for now, ask @cooper-joe: Double click normally selects the text in an input, do we really want to disable that behavior?
- [ ] Comments
  + [x] Handled through the regular data value endpoint
  + [ ] [Fetch comment when comment unit expanded](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/comment.js#L16-L22)
  + [ ] [Update comment when user saves changes](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/comment.js#L24-L25)
- [ ] Limits
  + Endpoint for min-max values/limits has not been implemented yet? 
  + [ ] [Fetch limits when limits unit expanded](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/limits.js#L18-L24)
  + [ ] [Update limits when user saves changes](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/limits.js#L26-L27)
  + [ ] [Allow user to delete limits](https://github.com/dhis2/data-entry-app/blob/ce2df741a673ab5031cdf65f4328ac8070375596/src/sidebar/limits.js#L28)


src/sidebar/audit-log.js
```diff
@@ -0,0 +1,129 @@
...
+    return (
+        <>
+            <span
+                className={styles.entryMessage}
+                dangerouslySetInnerHTML={{ __html: renderMessage(entry) }}
+            ></span>
+            <span className={styles.entryTimeAgo}>{timeAgo}</span>
+        </>
+    )
+}
```
mediremi, line 59: Couldn't find a better way to do this :grimacing:

src/app/app.js
```diff
@@ -16,17 +17,45 @@ import useQueryClient from './use-query-client.js'
...
+    const currentItem = {
+        id: 'aC392Jk3200',
+        name: 'Malaria treated at PHU without ACT',
+        code: 'DE2398742',
+        lastUpdated: {
+            at: new Date('2022-01-14'),
+            userDisplayName: 'Mwangi Babatunde',
+        },
+        markedForFollowup: false,
+        type: 'numerical',
```
mediremi, line 30: This is how I imagined the schema of data items would be based on the requirements in the design spec.

src/sidebar/sidebar.js
```diff
@@ -0,0 +1,60 @@
...
+            at: new Date('2022-01-01'),
+        },
+        {
+            displayName: 'Firstname Lastname',
+            changeType: 'DELETE',
+            oldValue: '15',
+            at: new Date('2021-10-01'),
+        },
+    ],
+}
```
mediremi, line 34: Like the data item schema, this is how I imagined the schemas for the various data item related resources based on the requirements of the design spec.

src/sidebar/audit-log.js
mediremi: that's right :100:

Birkbjo: Guess this is an argument to look into supporting [react-i18next](https://react.i18next.com/latest/trans-component)

src/sidebar/audit-log.js
```diff
@@ -0,0 +1,129 @@
...
+
+const Entry = ({ entry }) => {
+    const timeAgo = moment(entry.at).fromNow()
+
+    return (
+        <>
+            <span
+                className={styles.entryMessage}
+                dangerouslySetInnerHTML={{ __html: renderMessage(entry) }}
```
Birkbjo, line 54: If we cannot find a better solution for translations, it's probably worth a comment to explain why this is needed, and that this is safe since all values are actually sanitized - `i18next` sanitizes all values by default.

mediremi: Ah just realised that the default case is vulnerable to XSS. I'll fix that

src/sidebar/title.js
```diff
@@ -0,0 +1,24 @@
...
+            title={i18n.t('Close details sidebar')}
+            onClick={onClose}
+        >
+            <IconCross24 color={colors.grey700} />
+        </button>
+    </header>
+)
+
+Title.propTypes = {
+    onClose: PropTypes.func.isRequired,
```
Mohammer5, line 21: This is really a non-issue and I just want to write down what's on my mind right now..
I thought about the names of these callbacks recently. I think this should be called `onClick` or `close` (leaning towards the latter):

* When reading through the layout or sidebar component's code, `onClose` implies (to me) that this component reacts to when the sidebar is being closed (sth like: `onClose(() => doSth())`)
* `close` implies that this actually triggers some kind of action
* `onClick` leaves the decision about what should happen to the `Sidebar` (but one could argue that there's a cross icon in this component, so that separation does not make any sense, which is why I'm leaning towards the `close` option)
* `onCloseClick` is kind of the middle ground between the two above, inside the `<Layout />` component would do something like this: `sidebar={<Sidebar onCloseClick={hideSidebar} />`

src/sidebar/comment.js
```diff
@@ -0,0 +1,130 @@
...
+                            <Button
+                                small
+                                primary
+                                type="submit"
+                                loading={submitting}
+                            >
+                                {submitting
+                                    ? i18n.t('Saving...')
+                                    : i18n.t('Save comment')}
+                            </Button>
```
Mohammer5, line 88: Should this button be disabled while loading?

src/sidebar/comment.js
```diff
@@ -0,0 +1,130 @@
...
+                            <Button
+                                small
+                                secondary
+                                onClick={() => setEditing(false)}
+                            >
+                                {i18n.t('Cancel')}
+                            </Button>
```
Mohammer5, line 95: This one should def. be disabled while loading is true: Even if we cancelled the outgoing request, there's a chance it hit the server already and we'd be simply prevent receiving the response.

src/sidebar/limits.js
```diff
@@ -0,0 +1,225 @@
...
+}
+
+Limits.propTypes = {
+    itemId: PropTypes.string.isRequired,
+}
+
+const LimitsUnit = ({ itemId, itemType }) => (
+    <ToggleableUnit
+        title={i18n.t('Minimum and maximum limits')}
+        disabled={itemType !== 'numerical'}
```
mediremi, line 214: `numerical` isn't actually a valid item type: see https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java for list of possible types

src/sidebar/audit-log.js
KaiVandivier: I'm not sure of a better way than `dangerouslySetInnerHTML` to get the different styles within the text, but I'm wondering if `span` and `class` would get translated and become invalid HTML -- what do you think about:
```
i18n.t(
    '{{displayName}} {{htmlOpenTag}}updated{{htmlCloseTag}} to {{newValue}} (was {{oldValue}})',
    {
        displayName,
        htmlOpenTag: `<span class="${styles.entryType}">`,
        htmlCloseTag: '</span>'
        newValue,
        oldValue,
    }
)
```

Is there such thing as RTL html? The order of the tags matters I guess -- could be `htmlTagLeft` and `htmlTagRight` as keys maybe

src/sidebar/comment.js
mediremi: Setting `loading=true` disables `@dhis2/ui` buttons: https://github.com/dhis2/ui/blob/afc6eb5ef84a9796330b7eeac8addd2dd337f786/components/button/src/button/button.js#L60

src/sidebar/audit-log.js
mediremi: Yeah either solution relies on translators not accidentally messing with the HTML and I'm not sure which one is less error prone.

src/sidebar/comment.js
mediremi: Fixed in commit https://github.com/dhis2/data-entry-app/pull/18/commits/b3278b944710e7442a1aa88ab3756a74667a89c5

src/sidebar/audit-log.js
Mohammer5: I think we should just not use any html whatsoever until it's actually supported, unless this is really important.
If I remember correctly, that's the strategy we chose in the sms config app with @cooper-joe

src/data-workspace/use-close-right-hand-panel-on-selection-change.js
```diff
@@ -0,0 +1,19 @@
...
+        () => {
+            rightHandPanel.id && rightHandPanel.hide()
+        },
+        // contextSelection is part of the dependency array as we need to hide
+        // the sidebar when the selection changes. There's no need to use the
+        // context selection in the useEffect though
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+        [searchParams]
+    )
+}
```
HendrikThePendric, line 19: Probably you would want to add the rightHandPanel to the dependency array here, or work around this in another way. In the current code, you could end up with a stale reference to `rightHandPanel`.

d2.config.js
```diff
@@ -2,7 +2,7 @@ const config = {
     type: 'app',
     title: 'Data Entry',
     pwa: {
-        enabled: process.env.REACT_APP_NODE_ENV !== 'test',
+        enabled: process.env.REACT_APP_NODE_ENV !== 'production',
```
Mohammer5, line 5: PWA caused some weird behavior from time to time during dev, we could revert this change before merging

src/context-selection/attribute-option-combo-selector-bar-item/use-selected.js
```diff
@@ -2,14 +2,39 @@ import { useCallback } from 'react'
...
+        )
+
+        if (
+            JSON.stringify(nextAocSelection) ===
+            JSON.stringify(attributeOptionComboSelection)
+        ) {
+            return
+        }
+
+        return setAttributeOptionComboSelection(nextAocSelection)
```
Mohammer5, line 32: Only deselect options that are not available in the new selection

Birkbjo, line 32: I find it a bit confusing to return a void/undefined. I would expect just 
```javascript
setAttributeOptionComboSelection(nextAocSelection)
```.
and no return in that case. I started to look into the return value of `setAttributeOptionComboSelection`.

File: src/context-selection/attribute-option-combo-selector-bar-item/use-selected.js [hunk 12]
```diff
@@ -2,14 +2,39 @@ import { useCallback } from 'react'
...
+                    )
+                }
+            )
+        )
+
+        if (
+            JSON.stringify(nextAocSelection) ===
+            JSON.stringify(attributeOptionComboSelection)
+        ) {
+            return
```
Birkbjo, line 29: Is this needed? What happens if you just set the same selection. Is something refetched or re-rendered? I would think setting the same selection wouldn't really do anything?

Mohammer5: You're right, also with the comment above, I could change the two return statements to:

```js
        if (
            JSON.stringify(nextAocSelection) !==
            JSON.stringify(attributeOptionComboSelection)
        ) {
            setAttributeOptionComboSelection(nextAocSelection)
        }
```

Mohammer5: You're right. I can completely omit the condition and the two return statements (which also addresses your comment above)

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -43,7 +65,75 @@ export function EntryFieldInput({ fieldname, dataElement: de, setSyncStatus }) {
...
+    const sharedProps = {
+        fieldname,
+        dataValueParams,
+        setSyncStatus,
+        onFocus,
+        onKeyDown,
+    }
+
+    return (
+        <>
```
Birkbjo, line 127: What's the need for the fragment here, this can just return the `InputComponent` directly?

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -8,19 +11,38 @@ import {
...
     TrueOnlyCheckbox,
 } from '../inputs/index.js'
+import { useDataValueSet } from '../use-data-value-set.js'
 import { useDataValueParams } from './use-data-value-params.js'
 import { VALUE_TYPES } from './value-types.js'
 
-export function EntryFieldInput({ fieldname, dataElement: de, setSyncStatus }) {
-    const [deId, cocId] = fieldname.split('.')
-    const dataValueParams = useDataValueParams({ deId, cocId })
+function useCreateCurrentItem({ de, coc, dataValueSet }) {
```
Birkbjo, line 18: Why is this a hook? This can just be a regular function.

HendrikThePendric (commented):
Great work 💪 . This PR contains way too many changes to review line-by-line in a reliable way. My overall impression is very positive though. I left a few comments in the code about some specific points. And also have two more general ones:
- During functional testing I encountered a bug in the context selection UI, which can be reproduced like this:
    - Make complete selection, for example ART monthly summary + Sierra Leone + any period + any implementing partner
    - Now choose another dataset to see the UI having a really tough time, or just change the year to see it just having a bad day
- This PR introduces two providers. Perhaps we should consider adding some sort of global state management solution? 

I'm only commenting, not requesting changes because I'm off tomorrow and don't want to block things.

src/right-hand-panel/use-right-hand-panel-context.js
```diff
@@ -0,0 +1,6 @@
+import { useContext } from 'react'
+import RightHandPanelContext from './right-hand-panel-context.js'
+
+export default function useRightHandPanelContext() {
```
Birkbjo, line 4: Let's create some hooks to just expose the `set`-function. It seems like this is used in every `InputField`, and thus all field's are re-rendered whenever the panel is shown. That is not necessary.


```javascript
export function useShowRightHandPanel() {
    const { show } = useContext(RightHandPanelContext)
    return useMemo((id) => {
        show(id)
    }, [show])
}

export function useHideRightHandPanel() {
    const { hide } = useContext(RightHandPanelContext)
    return useMemo(() => {
        hide()
    }, [hide])
}
```

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -43,7 +65,75 @@ export function EntryFieldInput({ fieldname, dataElement: de, setSyncStatus }) {
...
+            event.preventDefault()
+            focusNext()
+        } else if (key === 'ArrowUp') {
+            event.preventDefault()
+            focusPrev()
+        }
+    }
+
+    const onFocus = () => {
+        rightHandPanel.hide()
```
Birkbjo, line 115: As a user, I actually find this annoying. If I want to see the sidebar-info, I most likely want to explore that information for other values/fields as well.
I'm not sure if this is in the spec, but if so I think it's something we should discuss.

src/context-selection/contextual-help/cell-reference.js
```diff
@@ -0,0 +1,49 @@
...
+        />
+        <Divider />
+        <CellExample
+            name={i18n.t('Invalid value, not saved')}
+            state="INVALID"
+        />
+        <Divider />
+        <CellExample name={i18n.t('Locked, not editable')} state="LOCKED" />
+    </ToggleableUnit>
+)
```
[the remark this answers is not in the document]
Mohammer5: I think that "legend" is the most arrucate term. The "things" described by a legend are called "symbols", so how about renaming the component itself to `<CellsLegend />` and the individual examples `<CellsLegendSymbol />`?

src/context-selection/attribute-option-combo-selector-bar-item/use-selected.js
```diff
@@ -2,14 +2,39 @@ import { useCallback } from 'react'
...
+                        (category) =>
+                            category.id === categoryId &&
+                            category.categoryOptions.find(
+                                (categoryOption) =>
+                                    categoryOption.id === optionId
+                            )
+                    )
+                }
+            )
+        )
```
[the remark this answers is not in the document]
Mohammer5: I could create a `filterObject` util, which internally does both the `.entries` and `.fromEntries` call:

```js
        const nextAocSelection = filterObject(
            ({ key: categoryId, value: optionId }) => {
                return relevantCategoriesWithOptions.find((category) =>
                    category.id === categoryId &&
                    category.categoryOptions.find(
                        (categoryOption) => categoryOption.id === optionId
                    ))
            },
            attributeOptionComboSelection
        )
```

That might be a lot more readable?

src/shared/sidebar/toggleable-unit.js
```diff
@@ -0,0 +1,55 @@
+import i18n from '@dhis2/d2-i18n'
+import { IconChevronUp24, IconChevronDown24, colors } from '@dhis2/ui'
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import React, { useState } from 'react'
+import styles from './toggleable-unit.module.css'
+
+const ToggleableUnit = ({ title, children, initiallyOpen, disabled }) => {
```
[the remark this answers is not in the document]
Mohammer5: Would `ExpandableUnit` be ok for you?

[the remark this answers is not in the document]
Birkbjo: > Would `ExpandableUnit` be ok for you?

Sounds good!

src/data-workspace/data-details-sidebar/limits.js
```diff
@@ -0,0 +1,226 @@
...
+const query = {
+    limits: {
+        resource: 'limits',
+        id: ({ id }) => id,
+    },
+}
+
+// TODO
+const updateMutation = {}
+const deleteMutation = {}
```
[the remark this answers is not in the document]
Mohammer5: https://dhis2.atlassian.net/browse/TECH-1199

src/data-workspace/data-details-sidebar/limits.js
```diff
@@ -0,0 +1,226 @@
...
+    IconInfo16,
+    colors,
+} from '@dhis2/ui'
+import PropTypes from 'prop-types'
+import React, { useEffect, useState } from 'react'
+import { ToggleableUnit } from '../../shared/index.js'
+import styles from './limits.module.css'
+import LoadingError from './loading-error.js'
+
+// TODO
```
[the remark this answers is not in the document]
Mohammer5: Not done yet
https://dhis2.atlassian.net/browse/TECH-1199

HendrikThePendric (approved):
Code is looking great and the bug I was seeing before seems to be gone. You just need to clean up some unused imports to fix the lint errors

Birkbjo (approved):

PR #11, opened 2022-01-12
feat: data entry cell add ons

KaiVandivier:

PR #19, opened 2022-01-17
feat: data workspace foundation

Birkbjo:


Mohammer5:
Is it too early to say something about file structure, etc? Do you want to tackle that once the functionality is complete?
Or is this close to what's going to be proposed and can be discussed already (could do it on slack too, might be easier)?

Birkbjo:
> Is it too early to say something about file structure, etc? Do you want to tackle that once the functionality is complete? Or is this close to what's going to be proposed and can be discussed already (could do it on slack too, might be easier)?

Well, we are going to merge Kai's #11 into this, so that moves the data-entry cells stuff. 
Other than that, it's mostly what I'm thinking yeah. We could move the `metadata` stuff out as well, as other features probably will use it. That would include `utils`, `selectors` and `metadata-context`

PR #23, opened 2022-01-21
feat: section styles

KaiVandivier:
Adds spacing between entry form elements (filter input and tables) and fixes some section header styles, including a too-short filter input

### DIFF
--- a/src/data-workspace/data-workspace.module.css
+++ b/src/data-workspace/data-workspace.module.css
@@ -1,4 +1,8 @@
 .wrapper {
+    display: flex;
+    flex-direction: column;
+    align-items: start;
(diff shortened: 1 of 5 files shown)


Birkbjo (approved):
LGTM.
Btw, `style`-commit-type is usually to refer to code-style. 
`style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)`
From: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines

KaiVandivier:
Oh, thanks 😅 I'll change the commit messages. Would that be a `feat` instead?

PR #26, opened 2022-01-24
feat: invalid cell tooltip

KaiVandivier:
Adds a tooltip to cells when the type of the value in the cell does not match the data element's `valueType`, as described in the [design specifications](https://docs.google.com/document/d/1NDMM-FAGCdM41n8ShnQuS6snvKvdSWNW-rky23cio_I/edit#heading=h.6ajrvb1c1ck)

### DIFF
--- a/src/data-workspace/data-entry-cell/data-entry-cell.js
+++ b/src/data-workspace/data-entry-cell/data-entry-cell.js
@@ -104,12 +104,17 @@ export function DataEntryCell({ dataElement: de, categoryOptionCombo: coc }) {
 
     return (
         <td className={styles.dataEntryCell}>
-            <ValidationTooltip invalid={meta.invalid} error={meta.error}>
+            <ValidationTooltip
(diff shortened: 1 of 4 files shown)


Mohammer5 (approved):
The tooltip itself seems to work.
I noticed one odd thing though: https://pr-26--dhis2-data-entry.netlify.app/#/?dataSetId=BfMAe6Itzgt&orgUnitId=ImspTQPwCqd&periodId=202112

When entering "a" as value for "BCG Doses given", "Fixed" ">1y", the app seems to process the value and then remove it instead of setting the cell to invalid (didn't happen in other cells though)

KaiVandivier:
For posterity, as discussed in DMs: A validator function isn't implemented yet for "BCG doses given"'s `INTEGER` value type, so it isn't actually getting validated and the app attempts to post that value -- future validator implementations will fix it! :)

PR #29, opened 2022-01-26
feat: add react-query cache persistence

ismay:


Birkbjo:
> Note that the last commit is a somewhat naive implementation for debugging purposes.

Is this the reason why this is not using `persistQueryClient` and `createWebStoragePersistor ` as well? Ref https://react-query.tanstack.com/plugins/persistQueryClient

Also seems like it should be pretty straightforward to use `IndexedDB`: https://github.com/tannerlinsley/react-query/discussions/3198
Might not be necessary for dataValues, but we have to use IndexedDB for `metadata`. But we will probably not use the react-query integration there as we need indexing etc.

ismay:
> > Note that the last commit is a somewhat naive implementation for debugging purposes.
> 
> Is this the reason why this is not using `persistQueryClient` and `createWebStoragePersistor ` as well? Ref https://react-query.tanstack.com/plugins/persistQueryClient

Yeah exactly. The first commit is using the recommended approach, but for debugging I wanted a little more control over the process.

> Also seems like it should be pretty straightforward to use `IndexedDB`: [tannerlinsley/react-query#3198](https://github.com/tannerlinsley/react-query/discussions/3198) Might not be necessary for dataValues, but we have to use IndexedDB for `metadata`. But we will probably not use the react-query integration there as we need indexing etc.

Yeah the storage mechanism is easy to swap out. I'm not sure why we wouldn't use RQ's integration for the metadata though. Is there a reason you prefer having the data searchable in persistent storage over just loading it all into memory on boot? To me the latter seems simpler to implement at the cost of some cpu overhead on boot.

Birkbjo:
> Yeah the storage mechanism is easy to swap out. I'm not sure why we wouldn't use RQ's integration for the metadata though. Could you elaborate on the reason you prefer having the data searchable in persistent storage over just loading it all into memory on boot? To me the latter seems simpler to implement at the cost of some overhead on boot (and memory, but since that's in the order of MB's that shouldn't be a problem).

Yeah, true. Might have been too eager with the "need", but it's definitely something we should monitor. At first I thought we would have to serialize to json and store the JSON-string in IndexedDB (to overcome storage limit in localStorage).  We should investigate how serialisation and deseraizliation on big databases performs though - and especially on lower hardware. But I agree that this might not be as bad as I first envisioned.

ismay:
> > Yeah the storage mechanism is easy to swap out. I'm not sure why we wouldn't use RQ's integration for the metadata though. Could you elaborate on the reason you prefer having the data searchable in persistent storage over just loading it all into memory on boot? To me the latter seems simpler to implement at the cost of some overhead on boot (and memory, but since that's in the order of MB's that shouldn't be a problem).
> 
> Yeah, true. Might have been too eager with the "need", but it's definitely something we should monitor. At first I thought we would have to serialize to json and store the JSON-string in IndexedDB (to overcome storage limit in localStorage). We should investigate how serialisation and deseraizliation on big databases performs though - and especially on lower hardware. But I agree that this might not be as bad as I first envisioned.

I think an optimization that would be a happy medium between both approaches is using indexeddb's object store, which means we won't have to serialize ourselves. Indexeddb will serialize with native code (C++ from what I read), instead of us having to do it in js with JSON.stringify. I believe I saw discussion of a persister using this approach in react-query's discussions. That means we could use RQ's abstractions with an idbobjectstore for ease of use, have a fast mechanism for storing and retrieving and just keep it all in memory. I think that'd be both relatively simple in terms of the code, and fast as well.

ismay:
Or maybe that was what you meant to do already? I'm not sure. In that case, the only change would be that we'd store it in idb through RQ instead of handling it ourselves. Which means we don't have to bother with the low level details of the implementation.

Birkbjo:
> That means we could use RQ's abstractions with an idbobjectstore for ease of use, have a fast mechanism for storing and retrieving and just keep it all in memory.

> Or maybe that was what you meant to do already? I'm not sure. In that case, the only change would be that we'd store it in idb through RQ instead of handling it ourselves. Which means we don't have to bother with the low level details of the implementation.

Yeah, exactly what I meant 😄

Birkbjo:
> Cache persistence for mutations with an offline server is still acting weird, but the app is no longer crashing during hydration. We can also look into replacing the localstorage persister with an idb one after this.

This might also be due to the service-worker responding to API-request as well maybe?

Birkbjo (approved):

ismay:
> > Cache persistence for mutations with an offline server is still acting weird, but the app is no longer crashing during hydration. We can also look into replacing the localstorage persister with an idb one after this.
> 
> This might also be due to the service-worker responding to API-request as well maybe?

Yeah that could be. The whole thing is fairly flaky at the moment, with both a service worker and a dev environment on localhost that is proxying making it a fair amount more unpredictable it seems.

PR #30, opened 2022-01-26
fix(query-client): disable refetch on focus

Birkbjo:
In the spirit of reducing network chattiness, I've disabled `refetchOnWindowFocus`.

It's also mostly because it's annoying to see all the network traffic in the network\-tab when debugging.

We might want to disable `refetchOnReconnect` as well, but that might have some more implications, and it would be interesting to see how that behaves with the optimistic updates.

### DIFF
--- a/src/app/use-query-client.js
+++ b/src/app/use-query-client.js
@@ -15,6 +15,7 @@ const useQueryClient = () => {
         defaultOptions: {
             queries: {
                 queryFn,
+                refetchOnWindowFocus: false,
             },
         },
     })
--- a/src/data-workspace/data-workspace.js
+++ b/src/data-workspace/data-workspace.js
@@ -155,7 +155,6 @@ export const DataWorkspace = () => {
 
     useQuery([metadataQuery], {
         staleTime: 60 * 24 * 1000,
-        refetchOnWindowFocus: false,
         onSuccess: (data) => setMetadata(data.metadata),
     })
 


ismay (commented):
Looks good. We could now remove this same setting from the data-workspace.js right?

> In the spirit of reducing network chattiness, I've disabled refetchOnWindowFocus.
> It's also mostly because it's annoying to see all the network traffic in the network-tab when debugging.

To comply with our goal of reducing network overhead in production this seems like a good change. The requests during dev. don't really matter to me if the setting is useful in production. If that's the only goal I feel an env var like NODE_ENV would be more suited.

> We might want to disable refetchOnReconnect as well, but that might have some more implications, and it would be interesting to see how that behaves with the optimistic updates.

It shouldn't affect the optimistic updates. Since the query that is updated optimistically is disabled whilst there are any pending mutations it shouldn't refetch if it has active optimistic updates.

Birkbjo:
> Looks good. We could now remove this same setting from the data-workspace.js right?

Yes, forgot about that - will fix.

> To comply with our goal of reducing network overhead in production this seems like a good change. The requests during dev. don't really matter to me if the setting is useful in production. If that's the only goal I feel an env var like NODE_ENV would be more suited.

I may have worded myself wrong here. What I ment was that it was the network-request being annoying triggered the removal at first - but of course if we would want this in production we could look into using an env-var... But we don't so I disabled it across the board.

> It shouldn't affect the optimistic updates. Since the query that is updated optimistically is disabled whilst there are any pending mutations it shouldn't refetch if it has active optimistic updates.

Great, so this is probably not sending request either way - since most likely a user will have edit a field before going online again.

ismay (approved):

ismay:
Thanks for the fix, looks good to me!

> I may have worded myself wrong here. What I meant was that it was the network-request being annoying that triggered the removal at first - but of course if we would want this in production we could look into using an env-var... But we don't so I disabled it across the board.

Ah ok I see what you mean :+1:

> Great, so this is probably not sending request either way - since most likely a user will have edit a field before going online again.

Yeah exactly.

PR #32, opened 2022-01-28
feat: add more input types and validators

KaiVandivier:
Here is another round of validators and input types based on different data types. There are a lot (that are described in the 'Data Types' section of [the design spec](https://docs.google.com/document/d/1NDMM-FAGCdM41n8ShnQuS6snvKvdSWNW-rky23cio_I/edit#heading=h.g44lloa2cs97)), and several of them require special handling which makes it challenging to make the inputs cleanly modular, but an effort is being made.

Accounting and notes about these data types and completion status is tracked in this document: [Value types: inputs and validators](https://docs.google.com/document/d/1o_GuiwpN0eXzj6Uv2QHt0hOl2-ixU_BOTLMHVkNih6Y/edit)

Known issues:
* Focus handling on `Select` components (and therefore the data syncing flow) is dysfunctional; a custom component will have to be created


src/data-workspace/data-entry-cell/data-entry-cell.js
```diff
@@ -10,26 +10,54 @@ import {
...
+
+/** Grey triangle in bottom left of cell */
+const CommentIndicator = ({ isComment }) => {
+    return (
+        <div className={styles.bottomLeftIndicator}>
+            {isComment && <div className={styles.bottomLeftTriangle} />}
+        </div>
+    )
+}
+CommentIndicator.propTypes = { isComment: PropTypes.bool }
```
KaiVandivier, line 42: Refactored out here to simplify main component JSX

src/data-workspace/data-entry-cell/data-entry-cell.module.css
```diff
@@ -12,41 +12,35 @@
...
-    height: 100%;
-    background: none;
-    border: none;
-    padding: 8px 16px 8px 8px;
-}
-.input:disabled {
-    background: var(--colors-grey300);
-}
-.input:hover:not(:disabled) {
+.cellInnerWrapper:hover:not(.disabled) {
```
KaiVandivier, line 18: Styles in here are refactored to work for different input types that might be rendered

src/data-workspace/data-entry-cell/inputs.js
```diff
@@ -0,0 +1,309 @@
...
+        type: 'radio',
+        value: 'false',
+        subscription: { value: true },
+    })
+    // Used for the 'clear' button, but works
+    const clearField = useField(name, {
+        type: 'radio',
+        value: '',
+        subscription: { value: true },
+    })
```
KaiVandivier, line 165: For radios, each button needs a Final Form field

src/data-workspace/data-entry-cell/inputs.js
```diff
@@ -0,0 +1,309 @@
...
+                    }}
+                    // probably not used, but included as back-up to onClick:
+                    onBlur={() => {
+                        handleBlur()
+                        clearField.input.onBlur()
+                    }}
+                >
+                    {i18n.t('Clear')}
+                </Button>
+            )}
```
KaiVandivier, line 228: This 'clear' button is similar to a radio in that it sets this field's value to `''` when clicked

src/data-workspace/data-entry-cell/value-types.js
```diff
@@ -0,0 +1,127 @@
...
+} from './validators.js'
+
+/** Use an OptionSet if relevant; else use input type below */
+export function getInputByDataElement(de) {
+    if (de.optionSetValue) {
+        return OptionSet
+    } else {
+        return VALUE_TYPES[de.valueType].Input
+    }
+}
```
KaiVandivier, line 36: Match the appropriate input for the data element

src/data-workspace/data-entry-cell/value-types.js
```diff
@@ -0,0 +1,127 @@
...
+
+export const VALUE_TYPES = Object.freeze({
```
KaiVandivier, line 38: Dictionary of inputs and validators for value types

src/data-workspace/data-workspace.js
```diff
@@ -55,7 +56,9 @@ const metadataQuery = {
             // set element
             'dataSets:fields':
                 'id,displayFormName,formType,dataSetElements[dataElement,categoryCombo],categoryCombo,sections~pluck',
-            'dataElements:fields': 'id,displayFormName,categoryCombo,valueType',
+            'dataElements:fields':
+                'id,displayFormName,categoryCombo,valueType,optionSetValue,optionSet',
+            'optionSets:fields': 'id,options[id,name,sortOrder]',
```
KaiVandivier, line 61: need `optionSet` data for some data elements

src/data-workspace/data-entry-cell/inputs.js
```diff
@@ -0,0 +1,309 @@
...
+
+// Adapt UI components to Final Form's callbacks
+const convertCallbackSignatures = (props) => ({
+    ...props,
+    onChange: (_, e) => props.onChange(e),
+    onFocus: (_, e) => props.onFocus(e),
+    onBlur: (_, e) => props.onBlur(e),
+})
+
+export const withAddlProps = (Component, addlProps) => {
```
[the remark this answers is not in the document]
KaiVandivier: "Additional" -- I'll write it out 😁

src/data-workspace/data-entry-cell/inputs.js
```diff
@@ -0,0 +1,309 @@
...
+}) => {
+    const { input, meta } = useField(name, {
+        type: 'checkbox',
+        subscription: { value: true, dirty: true, valid: true },
+    })
+
+    // todo: checking then unchecking the box will send a single unnecessary POST
+    const handleBlur = () => {
+        // For 'True only', can only send 'true' (or '1') or ''
+        const value = input.checked ? 'true' : ''
```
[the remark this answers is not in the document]
KaiVandivier: Hah yeah, kinda quirky

Mohammer5:
Could you make this a draft PR? Then we know when it's ready for review

KaiVandivier:
It's pretty much ready - just trying to debug one thing 🤔 gonna post in slack

package.json
```diff
@@ -54,6 +54,7 @@
         "react-final-form": "^6.5.7",
         "react-query": "4.0.0-alpha.10",
         "react-router-dom": "^5",
+        "tabbable": "^5.2.1",
```
KaiVandivier, line 57: Used for keyboard navigation among workspace

package.json
```diff
@@ -41,7 +41,7 @@
         "start-server-and-test": "^1.14.0"
     },
     "dependencies": {
-        "@dhis2/app-runtime": "^3.2.6",
+        "@dhis2/app-runtime": "^3.3.0",
```
KaiVandivier, line 44: Adds ability for mutations to `dataValues` and `dataValues/file` to use `multipart/form-data` body type

src/data-workspace/data-entry-cell/data-entry-cell.js
```diff
@@ -10,20 +10,57 @@ import {
...
```
KaiVandivier, line 42: Refactored from component below

src/data-workspace/data-entry-cell/file-inputs.js
```diff
@@ -0,0 +1,143 @@
...
+            enabled:
+                !!input.value &&
+                meta.pristine &&
+                typeof input.value === 'string',
+        }
+    )
+    const { mutate: uploadFile } = useDataValueMutation(
+        MUTATION_TYPES.FILE_UPLOAD
+    )
+    const { mutate: deleteFile } = useDataValueMutation(MUTATION_TYPES.DELETE)
```
KaiVandivier, line 47: Files need special mutations, different from other `dataValues`

src/data-workspace/data-entry-cell/use-data-value-mutation.js
```diff
@@ -5,6 +5,35 @@ import { useAttributeOptionCombo } from '../use-attribute-option-combo.js'
...
+    DEFAULT: 'DEFAULT',
+    FILE_UPLOAD: 'FILE_UPLOAD',
+    DELETE: 'DELETE',
+}
+const mutationsByType = {
+    DEFAULT: DATA_VALUE_MUTATION,
+    FILE_UPLOAD: UPLOAD_FILE_MUTATION,
+    DELETE: DELETE_VALUE_MUTATION,
+}
+
```
KaiVandivier, line 36: Adds mutation types for file uploads and deletions

KaiVandivier:
Still experiencing this bug described [here](https://dhis2.slack.com/archives/C02NHB92S2H/p1646404472014619)

src/data-workspace/data-entry-cell/value-types.js
```diff
@@ -0,0 +1,126 @@
...
+} from './validators.js'
+
+/** Use an OptionSet if relevant; else use input type below */
+export function getInputByDataElement(de) {
+    if (de.optionSetValue) {
+        return OptionSet
+    } else {
+        return VALUE_TYPES[de.valueType].Input
+    }
+}
```
Mohammer5, line 37: The pattern I can see here is a pattern I have seen quite often in dhis2 now and I think we should try to avoid it as it's historically always became a mess, I'd also say that it's an anti-pattern to the react way (at least that's my experience.

The approach here is that there's a config (`VALUE_TYPES`) and the data entry cell can dynamically get what it needs. I'd call that the config based approach.

IMO the "correct" react-style approach would be to have components like `TextInput`, `LongTextInput`, etc.
The `getInputByDataElement` function would turn into a component that would have to be used like this: `<DataEntryCellInput isOptionSetValue={!!de.optionSetValue} valueType={de.valueType} />`, which renders one of the components below.

It'd be more boilerplate code, but it's the most maintainable style imo, as you can easily handle exceptions that will have to be added eventually at an unknown time (might be during QA testing, might be due to backend changes).
It's also easier to reason about the code that way, despite it having more LoCs.
Especially as the contents of the `value-types.js` are being used by the `data-entry-cell.js` module, I think a configuration based approach just leads to code that's harder to understand in the long run

Mohammer5 (changes requested):

src/data-workspace/data-entry-cell/data-entry-cell.js
```diff
@@ -48,64 +86,60 @@ export function DataEntryCell({ dataElement: de, categoryOptionCombo: coc }) {
...
 
-    const onBlur = (event) => {
-        // If this value has changed, sync it to server if valid
-        if (meta.dirty && input.value !== lastSyncedValue && meta.valid) {
-            syncData()
-        }
-        // Also invoke FinalForm's `onBlur`
-        input.onBlur(event)
+    const syncData = (value) => {
+        // Here's where an error state could be set: ('onError')
```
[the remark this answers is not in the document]
Mohammer5: Needs to be [uppercase](https://medium.com/imdoneio/the-imdone-pitch-feedback-welcome-386430accf01) to be consistent with other TODO comments

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -0,0 +1,57 @@
...
+            return <FileResourceInput {...sharedProps} image />
+        }
+        case VALUE_TYPES.LONG_TEXT: {
+            return <LongText {...sharedProps} />
+        }
+        case VALUE_TYPES.TRUE_ONLY: {
+            return <TrueOnlyCheckbox {...sharedProps} />
+        }
+        default: {
+            return <BasicInput {...sharedProps} valueType={de.valueType} />
```
Mohammer5, line 42: Maybe the name `GenericInput` is better? Dunno & doesn't matter anyway

src/data-workspace/data-entry-cell/use-data-value-params.js
```diff
@@ -0,0 +1,44 @@
...
+        dataEntryContext
+
+    const attributeComboId = getDataSetById(metadataFetch.data, dataSetId)
+        .categoryCombo.id
+    const isDefaultAttributeCombo = getCategoryComboById(
+        metadataFetch.data,
+        attributeComboId
+    ).isDefault
+
+    const dvParams = {
```
Mohammer5, line 26: Personal opinion: I'm against abbreviations. It just saves a bit if writing time l, but that shouldn't matter. It makes code harder to read though, so I don't think we should do it unless we're dealing with very long names

Mohammer5 (approved):
This looks good! I like that the files are much smaller now!

PR #33, opened 2022-01-31
feat(section): support for renderAsTabs

Birkbjo:
Support for rendering dataSets with `renderAsTabs`, to actually render tabs. 

I've integrated this with the `section`-selector from the `context-selector`, so that if a tab is changed, the context-selection is updated (and query-param), and visa versa. 

The `section.js`-file is starting to grow with multiple components, so I wonder if we should split out to `section-form`-file, and maybe but into a nested folder? 

TODO

- [x] Look into `renderHorizontally` in schema. Note that this is called `Render vertically` (thus inverted) in the Maintenance app, so you would think `renderHorizontally` would default to true - but that does not seem to be the case.

  -  This property is not actually related to tabs - it's related to multi-org units.


Mohammer5 (approved):
Code looks good.

---

> The section.js-file is starting to grow with multiple components, so I wonder if we should split out to section-form-file, and maybe but into a nested folder?

In regards to this, I generally find it hard to read the files with multiple contents, especially when it's components. I like the structure I've set up in the `src/context-selection` folder quite a bit, but haven't said anything yet as that's mostly preference and the highest priority is to get things done, refactoring code can be done post-release. I think the most important aspect is that things are consistent. The approach I went for was: max 2 level nesting and everthing has it's own file. Then it's quite clear where to find what and test-colocation is also trivial / easy to reason about.

src/context-selection/menu-select/menu-select.js
```diff
@@ -24,9 +24,10 @@ MenuSelect.propTypes = {
     values: PropTypes.arrayOf(
         PropTypes.shape({
             label: PropTypes.string.isRequired,
-            value: PropTypes.string.isRequired,
+            value: PropTypes.string,
```
Birkbjo, line 27: Using `undefined` for eg. "All sections", so this should be allowed.

src/context-selection/menu-select/menu-select.js
```diff
@@ -9,7 +9,7 @@ export default function MenuSelect({ values, selected, onChange }) {
             <Menu>
                 {values.map(({ value, label }) => (
                     <MenuItem
-                        key={value}
+                        key={value || label}
```
Birkbjo, line 12: Wondered if we should add an optional `key` -property to the `values`-object, but this was less code...

src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
```diff
@@ -18,6 +18,14 @@ export default function SectionFilterSelectorBarItem() {
 
     useOnDependentParamsChange(deselect)
 
+    useEffect(() => {
```
Birkbjo, line 21: This could have been done in the "TabbedSections"-component, but doing it here colocates the code with the rest of the selection, and we're able to preselect the first section in a much earlier stage.

Birkbjo:
> In regards to this, I generally find it hard to read the files with multiple contents, especially when it's components. I like the structure I've set up in the `src/context-selection` folder quite a bit, but haven't said anything yet as that's mostly preference and the highest priority is to get things done, refactoring code can be done post-release. I think the most important aspect is that things are consistent. The approach I went for was: max 2 level nesting and everthing has it's own file. Then it's quite clear where to find what and test-colocation is also trivial / easy to reason about.

Agree that it's very much preference, but I find your structure pretty well organised as well. But I do find it a bit heavy on the number of files. Eg. instead of having separate files for `compute-max-year.js` and `get-most-recent-completed-year.js`, I would've just created an utils, or `timeUtils` or something... One file for just importing a function from the other file is just too much separation and navigation in my opinion.

Generally I find it's OK and quicker to navigate if some very coupled components live together - eg. `SectionForms` and `FormSection` (don't love the inverted name there btw). The first just renders multiple of the other. I have to confess that some of it is also just laziness - more work to create more files - and especially if it needs styling - then you have to create `css`-files as well...

src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
```diff
@@ -34,10 +42,14 @@ export default function SectionFilterSelectorBarItem() {
...
-        ...sectionOptions,
-    ]
+    const selectableOptions = sectionOptions.concat({
+        value: undefined,
+        label: i18n.t('Show all sections'),
+    })
+
+    if (dataSetSectionsInfo.data?.renderAsTabs) {
+        selectableOptions.shift()
+    }
```
Mohammer5, line 52:
```suggestion
    const selectableOptions = sectionOptions.concat(
        !dataSetSectionsInfo.data?.renderAsTabs
            ? { value: undefined, label: i18n.t('Show all sections') }
            : undefined
    )
```

Mohammer5 (approved):
Just one suggestion I don't really care about, so just take it or leave it.
Regardless, I'll approve this PR and you can merge either without or with implementing the suggestion

src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
Birkbjo: I actually had it reversed at first - and removed the option by using `unshift`. I also thought about justing a let and concating if `renderAsTabs=false`. 
Anyhow, I do not like concatenating `undefined` - then the options will have an undefined element in the array, and depending on how the child-component that consumes this handles that, you can easily get errors. 

If however you are against mutating the array, which don't really matter here since we are creating a new one everytime either way - I could do it like this

```js
const selectableOptions = dataSetSectionInfo.data?.renderAsTabs ? sectionOptions : sectionOptions.concat({ value: undefined, label: i18n.t('Show all sections') })
```

Mohammer5: Ah, I was under the impression concat won't append anything when the value is undefined, which doesn't make any sense of course.. Nevermind my comment!

src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
```diff
@@ -18,6 +18,15 @@ export default function SectionFilterSelectorBarItem() {
...
+    // select first section if renderAsTabs === true
+    useEffect(() => {
+        const dataSet = dataSetSectionsInfo.data
+        const sections = dataSet?.sections
+        if (dataSet?.renderAsTabs && !sectionFilter && sections?.length) {
+            setSectionFilter(sections[0].id)
+        }
+    }, [dataSetSectionsInfo, sectionFilter])
+
     const shouldComponentRenderNull = useShouldComponentRenderNull()
```
KaiVandivier, line 30: I think some of the logic in the `useEffect` is pretty similar to the logic in `useShouldComponentRenderNull`; maybe the latter hook could be moved above the useEffect so that the value could be used there? This seems to work just fine though

KaiVandivier (changes requested):
My only request is that I think the new `section` dir should be called `section-form`; other than that, it looks good!

I think sharing the section selection logic between the context selector and the tabs makes sense.

KaiVandivier (approved):

PR #35, opened 2022-02-03
fix: use the existing query provider

ismay:
In the interest of keeping things as simple as possible I'm reusing the existing queryClient (from the app-runtime). Otherwise we'd have two layers of queryClients, which seems like a recipe for unpredictability.


src/app/use-configure-query-client.js
```diff
@@ -1,14 +1,15 @@
...
+import { useQueryClient } from 'react-query'
 import { createWebStoragePersistor } from 'react-query/createWebStoragePersistor-experimental'
 import { persistQueryClient } from 'react-query/persistQueryClient-experimental'
 import { dataValueSetQuery } from '../data-workspace/data-workspace.js'
 
 // Persisted data will be garbage collected after this time
 const MAX_CACHE_AGE = 1000 * 60 * 60 * 24 * 31 // One month
 
-const useQueryClient = () => {
+const useConfigureQueryClient = () => {
```
Mohammer5, line 10:
```suggestion
const useConfiguredQueryClient = () => {
```

Mohammer5 (commented):
Name is the only issue I could find

src/app/use-configure-query-client.js
ismay: It's not yet already configured though, the configuration happens in this hook. Basically configureQueryClient, and then useConfigureQueryClient to comply with the rules of hooks.

ismay: If it would return the configured queryClient I would prefer your name as well. Anyway, let me know what you think.

Mohammer5: Ah, sorry, I didn't get that.. My bad

Mohammer5 (approved):

src/app/use-configure-query-client.js
ismay: No worries, thanks for the review!

PR #36, opened 2022-02-07
feat: offline support with react-query v4

ismay:


src/context-selection/attribute-option-combo-selector-bar-item/use-category-combination.js
```diff
@@ -27,6 +27,6 @@ export default function useCategoryCombination() {
         called: !isIdle,
         loading,
         error,
-        data: data?.dataSet.categoryCombo,
+        data: data?.dataSet?.categoryCombo,
```
Mohammer5, line 30: How come this change is required, this should always be in the response?

src/context-selection/data-set-selector-bar-item/use-data-set.js
```diff
@@ -4,7 +4,6 @@ import { useDataSetId } from '../use-context-selection/index.js'
 const QUERY_DATA_SET = {
     dataSet: {
         resource: 'dataSets',
-        id: ({ id }) => id,
```
Mohammer5, line 7: What happened here?

src/context-selection/attribute-option-combo-selector-bar-item/use-category-combination.js
ismay: Yeah I was just sketching. I'd added guards everywhere because I didn't want to take the time to figure out which hooks were all right and which weren't.

src/context-selection/data-set-selector-bar-item/use-data-set.js
ismay: I've removed this whilst sketching as well, to silence an error that occured during the idb serialization. Anything in the query that can't be serialized will throw a datacloneerror.

ismay: Also, this pattern isn't necessary with RQ, but I've not pushed the changes where I've addressed that yet.

PR #38, opened 2022-02-10
feat: move to react-query alpha

ismay:
Todo:

- [x] fix e2e tests
- [x] throttle persist fn
- [x] ensure metadata works
- [x] ensure cache persistence & offline works

Notes

- Almost all queries now use the service worker as the authoritative cache whilst offline. Only queries that are optimistically updated use react-query's cache as the sole source of data when offline (otherwise we'd lose the optimistic updates). Note that react-query still does the SWR for all traffic that goes through its hooks whilst online.
- I had to change the format of our querykeys to allow idb to cache it. IDB doesn't allow storing anything that can't be serialized/deserialized (so functions for example), and our queryKeys contained callbacks in some cases. Since react-query doesn't need queries to be static (it's designed to pick up changes to the querykey) I've changed the format to: `useQuery(['resource', { id, params }])`. So for example: `useQuery(['dataSets', { id: dataSetId }])`. This will also return the data without nesting. This allows idb to persist everything without errors and it's simpler as well. See the changed code for examples. (I could have also used something like JSON.stringify / JSON.parse, or some other method to get rid of the non-serializable parts but then we'd lose a lot of the performance benefits of using idb).
- The exported prop-types I've copied to where they were used, as otherwise eslint can't lint them (eslint doesn't lint proptypes if they're not explicitly defined).
- I've removed some of the useMemo hooks during debugging. Some were defined after a return, which is against the rules of hooks. Others I'll look into after this PR. I'm aiming to condense some of the queries to a single query with variable selectors, to reduce network traffic. At that point I'll also look where we need to add back the useMemo. Total number of requests seems to already have gone down significantly btw.
- For hooks that are selecting a subset of the returned data I moved to using react-query's select callback. This is what react-query expects us to do and allows them to optimize for us.

Later/unrelated:

- combine queries that use fields to request partial data and use selectors instead (can reduce the amount of queries that way, reduce traffic by a lot. Would require centralizing data fetching hooks a little more)
- some hooks return an object wrapped in an array, seems overly complex if all that's used is the single object
- we're mixing default and named exports, we should add a linting rule and make it consistent
- don't use mutation key, but just use fuzzy querykey matching
- maybe centralize endpoint names
- are all exported metadata selectors used?
- revisit disabling SW during dev
- does disabling the dataValues query work with loading other forms? Do we even need to do that now that the network mode is set to online?


src/context-selection/attribute-option-combo-selector-bar-item/use-category-combination.js
```diff
@@ -46,47 +35,60 @@ const isOptionWithinPeriod = ({
 export default function useCategoryCombination() {
     const [periodId] = usePeriodId()
     const [dataSetId] = useDataSetId()
-    const queryKey = [QUERY_CATEGORY_COMBINATION, { id: dataSetId }]
+    const params = {
+        fields: [
+            'categoryCombo[isDefault,displayName, categories[id, displayName, categoryOptions[id, displayName]]]',
```
ismay, line 40: We're missing startDate and endData here. Is that correct?

Mohammer5: Yes

ismay: Yes as in it's ok for these to be removed?

src/context-selection/attribute-option-combo-selector-bar-item/use-category-combination.js
```diff
@@ -46,47 +35,60 @@ const isOptionWithinPeriod = ({
...
+        enabled: !!dataSetId && !!periodId,
+        select: (data) => {
+            if (!data || !periodId) {
+                return data?.categoryCombo
+            }
 
-    const categoryCombo = useMemo(() => {
-        if (!data || !periodId) {
-            return data?.dataSet.categoryCombo
-        }
```
ismay, line 62: I've removed this condition as this _should_ never happen now we've moved to using the selector. Because the query is disabled if there's no periodId, and if there's no data yet the selector won't be called.

src/context-selection/attribute-option-combo-selector-bar-item/use-category-combination.js
Mohammer5: No, they are required (until we have the new endpoint)

src/context-selection/attribute-option-combo-selector-bar-item/use-category-combination.js
Mohammer5: :+1:

src/app/persister.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Creates an Indexed DB persister
+ * @see https://react-query.tanstack.com/plugins/persistQueryClient#building-a-persistor
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
+ */
+
+const createIDBPersister = (idbValidKey = 'reactQuery') => {
+    // Throttle persisting by a second
+    const persistClient = throttle((client) => set(idbValidKey, client), 1000)
```
Mohammer5, line 12: Could you add an explanation for why this is necessary?

src/data-workspace/data-entry-cell/use-data-value-mutation.js
```diff
@@ -78,20 +69,19 @@ export const useDataValueMutation = () => {
...
+                        dataValue.orgUnit === newDataValue.ou &&
+                        dataValue.period === newDataValue.pe
+                )
                 const isNewValue = matchIndex === -1
 
                 // If the field was previously empty the dataValue won't exist yet
                 if (isNewValue) {
                     const formattedNewDataValue = {
-                        attributeOptionCombo: attributeOptionComboId,
+                        attributeOptionCombo,
```
Mohammer5, line 84: Was this a bug?

Mohammer5 (changes requested):
Just the missing `startDate` and `endDate` have to be added, just left two other questions

src/app/persister.js
ismay: Added comment in adf5f2846b05022886718924eecdd31e46b8800c. This is just to prevent writing to the cache more than once per second. When multiple queries are fetching you could potentially get a ton of calls to the persist fn. Which seems unnecessarily taxing. With this throttle we'll write on the leading and trailing edge so we should always write the last call to persist to the indexeddb.

Worst case scenario, the app crashes within the timeout of a second and less than a second of data could be lost. That seems like an ok tradeoff to me.

src/data-workspace/data-entry-cell/use-data-value-mutation.js
ismay: No just a rename, see here: https://github.com/dhis2/data-entry-app/pull/38/files#diff-f4d186e2e137850289b147c9ad12b5e536639e953c7b7e1dce531547b219cf1cL52

ismay:
> Just the missing `startDate` and `endDate` have to be added, just left two other questions

All right, I think I've addressed them all now. Let me know if I've missed something.

Also, because of the changed filter I need to update the mock data, but running `yarn cypress:capture` seems to fail all the tests. I'm using `https://debug.dhis2.org/dev` as the `dhis2BaseUrl` in `cypress.env.json`. Is that not correct?

Mohammer5 (approved):

PR #42, opened 2022-02-24
feat: load organisation unit offline levels

Mohammer5:
* [x] Blocked by dhis2/ui#993
* [x] Blocked by dhis2/app-platform#706


src/context-selection/org-unit-selector-bar-item/load-offline-level.js
```diff
@@ -0,0 +1,57 @@
...
+    const { children: childrenSize } = orgData.orgUnit
+    const nextOfflineLevels = offlineLevels - 1
+
+    let children
+    if (childrenSize) {
+        const orgChildren = await dataEngine.query(QUERY_ORG_CHILDREN_FROM_UI, {
+            variables,
+        })
+        children = orgChildren.orgUnit.children
+    }
```
Mohammer5, line 41:
```suggestion
    if (!childrenSize) {
        return
    }

    const orgChildren = await dataEngine.query(QUERY_ORG_CHILDREN_FROM_UI, {
        variables,
    })
    const children = orgChildren.orgUnit.children
```

src/context-selection/org-unit-selector-bar-item/use-load-offline-levels.js
```diff
@@ -0,0 +1,43 @@
...
+                        }
+
+                        return loadOfflineLevel(payload)
+                    })
+                )
+                    .then(() => setData(true))
+                    .catch(setError)
+                    .finally(() => setLoading(false))
+            })()
+    }, [dataEngine, offlineLevelsToLoadData, setLoading, setData])
```
mediremi, line 40:
```suggestion
    const loadOfflineLevels = async () => {
        setLoading(true)
        try {
            await Promise.all(
                offlineLevelsToLoadData.map((offlineLevelToLoad) => {
                    const payload = {
                        dataEngine,
                        ...offlineLevelToLoad,
                    }

                    return loadOfflineLevel(payload)
                })
            )
            setData(true)
        } catch (error) {
            setError(error)
        }
        setLoading(false)
    }

    useEffect(() => {
        if (offlineLevelsToLoadData) {
            loadOfflineLevels()
        }
    }, [dataEngine, offlineLevelsToLoadData, setLoading, setData])
```

src/context-selection/org-unit-selector-bar-item/use-load-offline-levels.js
```diff
@@ -0,0 +1,43 @@
...
+
+    return { loading, error, data }
```
mediremi, line 42: Since `data` is always just a boolean representing if offline levels have been successfully loaded, maybe this makes more sense:

```suggestion
    return { fetching, error, offlineLevelsLoaded }
```

src/context-selection/org-unit-selector-bar-item/get-offline-levels-to-load.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {String} data.userOrganisationUnits.id
+ * @param {Integer} data.userOrganisationUnits.level
+ *
+ * @returns {Array.<{ id: String, offlineLevels: Int }>}
+ */
+export default function getOfflineLevelsToLoad({
+    organisationUnitLevels,
+    userOrganisationUnits,
+}) {
+    return userOrganisationUnits.reduce((acc, userOrgUnit) => {
```
ismay, line 16: I'd replace this with a map and a filter. Reduce is horrible for readability in my opinion.

src/context-selection/org-unit-selector-bar-item/load-offline-level.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default async function loadOfflineLevel({
+    dataEngine,
+    id,
+    offlineLevels,
+}) {
+    const variables = { id }
+    const orgData = await dataEngine.query(QUERY_ORG_UNIT_FROM_UI, {
+        variables,
+    })
```
ismay, line 31: Why the data engine usage here?

src/context-selection/org-unit-selector-bar-item/use-load-offline-levels.js
```diff
@@ -0,0 +1,43 @@
...
+import { useEffect, useState } from 'react'
+import loadOfflineLevel from './load-offline-level.js'
+import useOfflineLevelsToLoad from './use-offline-levels-to-load.js'
+import useOrganisationUnitLevels from './use-organisation-unit-levels.js'
+
+/**
+ * As the service worker caches request responses,
+ * it should suffice to simply perform all request the org unit tree would
+ * perform as well in advance
+ */
```
ismay, line 11: Since we're already prefetching content with react query (the custom forms) I think it would also make sense to use react-query to prefetch here as well. That means the SW can be confined to just simple caching and means we aren't mixing cache strategies across the app.

src/context-selection/org-unit-selector-bar-item/use-organisation-unit-levels.js
```diff
@@ -0,0 +1,16 @@
+import { useQuery } from 'react-query'
+
+export default function useOrganisationUnitLevels() {
```
ismay, line 3: I think we should reorganize the query structuring a little for this PR. You could simplify things a lot with dependent queries and a query key factory. Probably best if we plan a short google meet for that, easier to explain it that way.

src/context-selection/org-unit-selector-bar-item/load-offline-level.js
Mohammer5: To make sure the request is identical to the one that the UI library does.
The service worker is supposed to cache this request and I thought that the easiest way to keep this app and the component in sync would be do duplicate what the org unit tree does in the library.

src/context-selection/org-unit-selector-bar-item/use-organisation-unit-levels.js
Mohammer5: Refactored as discuss

src/context-selection/org-unit-selector-bar-item/get-offline-levels-to-load.js
Mohammer5: Done

src/context-selection/org-unit-selector-bar-item/use-load-offline-levels.js
Mohammer5: the hook just returns `true` now. I actually wanted it to return nothing as the return value wasn't being used, but it's required for the jest test to work with `@testing-library/react-hooks`'s `withForNextUpdate`

src/context-selection/org-unit-selector-bar-item/use-load-offline-levels.js
Mohammer5: I've done it slightly differently, I moved the function out of the hook

ismay (approved):

Mohammer5:
@mediremi @ismay for some reason I can't test this with cypress. When I let cypress run on the cli, loading offline units consistently crashes (not when using the UI).

I'd say that this is sufficiently tested with jest, unless you disagree. Then I'll try to get this working with cypress

ismay:
> @mediremi @ismay for some reason I can't test this with cypress. When I let cypress run on the cli, loading offline units consistently crashes (not when using the UI).

Hmm that's strange. Would be good to create an issue for that so we don't forget to take a look later on.

Mohammer5:
> Would be good to create an issue for that so we don't forget to take a look later on.

:+1: https://dhis2.atlassian.net/browse/TECH-1073

PR #43, opened 2022-02-24
feat: prefetch custom forms

ismay:
- [x] Add hook that allows fetching custom form by id
- [x] Persist the prefetched data


src/app/query-client/use-query-client.js
```diff
@@ -10,12 +10,13 @@ const useQueryClient = () => {
...
-        const [resource, { params, id }] = queryKey
+        const resource = queryKey[0]
+        const options = queryKey.length > 1 ? queryKey[1] : {}
         const appRuntimeQuery = {
             [resource]: {
                 resource,
-                id,
-                params,
+                id: options.id,
+                params: options.params,
```
Mohammer5, line 19:
```js
const [resource, options] = queryKey
const appRuntime = {
  [resource]: {
    resource,
    id: options?.id,
    params: options?.params,
  },
}
```

ismay: Ah yeah of course. Good suggestion

Mohammer5 (approved):
LGTM

PR #46, opened 2022-03-02
feat(metadata): initital dataSetMetadata support

Birkbjo:

PR #47, opened 2022-03-03
feat: memoize selectors

ismay:
This adds memoization to our selectors in a manner that's easy to use with react-query's select, and shares its cache globally. So computations will only run once across the entire app as long as input stays the same. End goal being easy integration of memoization with our data fetching hooks.

Since reselect has a cache size of 1 it won't work with selectors that take parameters like an id. Usage with an id will invalidate the cache for other ids. So I'm using a small wrapper called re-reselect to create a reselect cache per id. Reselect and re-reselect are small. Will add just 2 to 3 kb to the bundle when minified, in total.

I've checked the memoization and currently it breaks only here: https://github.com/dhis2/data-entry-app/blob/development/src/data-workspace/category-combo-table.js#L84. Which is because a new array is created there every render, so that's expected behaviour. I can fix that after this branch.

I think it's probably best to rebase this on @Birk's work here: https://github.com/dhis2/data-entry-app/pull/46. Or I can help with rebasing that branch on this. We can see what's the easiest. I've kept all fn signatures intact, so it should be an easy rebase.

- ~~Renamed get -> select~~
- ~~Removed unused exports~~
- Removed unused selectors (see below)
- Added jsdoc param docs for all selectors without clear params
- Grouped metadata selectors to make file easier to scan (to my eyes at least, hope that generalizes to everyone)

Removed (unused) selectors:

- getDataElementById
- getSectionsByDataSetId
- getCategoryCombosByDataElements
- getCategoryCombosByDataSetId
- getCategoryOptionById
- getCategoryOptionsByCoCId
- getCategoryOptionComboById
- getSortedCategoryOptionCombosByCategoryComboId

### DIFF
--- a/src/custom-forms/use-custom-forms.js
+++ b/src/custom-forms/use-custom-forms.js
@@ -1,15 +1,20 @@
 import { useQuery } from 'react-query'
+import { createSelector } from 'reselect'
 import keys from './query-key-factory.js'
+
+const getDataSets = (data) => data.dataSets
+const getCustomForms = createSelector(getDataSets, (dataSets) =>
+    dataSets
+        .filter((dataSet) => dataSet.formType === 'CUSTOM')
+        .map((dataSet) => ({
+            id: dataSet.dataEntryForm.id,
+            version: dataSet.version,
+        }))
+)
 
 const useCustomForms = () => {
     return useQuery(keys.all, {
-        select: (data) =>
-            data.dataSets
-                .filter((dataSet) => dataSet.formType === 'CUSTOM')
-                .map((dataSet) => ({
-                    id: dataSet.dataEntryForm.id,
-                    version: dataSet.version,
-                })),
+        select: getCustomForms,
     })
 }
 
--- a/src/metadata/selectors.js
+++ b/src/metadata/selectors.js
@@ -1,229 +1,213 @@
-import { cartesian } from '../shared/utils.js'
+import { createCachedSelector } from 're-reselect'
+import { createSelector } from 'reselect'
 
-export const getDataElements = (metadata) => metadata.dataElements
+// Helper to group array items by an identifier
 
-export const getDataElementById = (metadata, id) =>
-    getDataElements(metadata)[id]
+const groupBy = (input, getIdentifier) =>
+    input.reduce((grouped, item) => {
+        const identifier = getIdentifier(item)
 
-export const getOptionSets = (metadata) => metadata.optionSets
+        grouped[identifier] = grouped[identifier] ?? []
+        grouped[identifier].push(item)
 
-export const getOptionSetById = (metadata, id) => getOptionSets(metadata)[id]
+        return grouped
+    }, {})
+
+// Simple input selectors
 
 export const getCategories = (metadata) => metadata.categories
-
 export const getCategoryCombos = (metadata) => metadata.categoryCombos
-
-export const getCategoryOptions = (metadata) => metadata.categoryOptions
-
 export const getCategoryOptionCombos = (metadata) =>
     metadata.categoryOptionCombos
-
+export const getCategoryOptions = (metadata) => metadata.categoryOptions
+export const getDataElements = (metadata) => metadata.dataElements
 export const getDataSets = (metadata) => metadata.dataSets
-export const getDataSetById = (metadata, dataSetId) =>
-    getDataSets(metadata)?.[dataSetId]
-
 export const getSections = (metadata) => metadata.sections
+export const getOptionSets = (metadata) => metadata.optionSets
 
-export const getSectionsByDataSetId = (metadata, dataSetId) => {
-    const sections = getSections(metadata)
-    const dataSet = getDataSetById(metadata, dataSetId)
+// Select by id
 
-    return dataSet.sections.map((sId) => sections[sId])
-}
-
-export const getDataElementsByDataSetId = (metadata, dataSetId) => {
-    const dataElements = getDataSetById(
-        metadata,
-        dataSetId
-    )?.dataSetElements?.map((dse) => {
-        const dataElement = getDataElementById(metadata, dse.dataElement.id)
-        // catcombo for a dataelement can be overriden per dataSet
-        // in that case, dataSetElement.categoryCombo takes precedence
-        const dataElementCategoryCombo =
-            dse.categoryCombo ?? dataElement.categoryCombo
-        return {
-            ...dataElement,
-            categoryCombo: dataElementCategoryCombo,
-        }
-    })
-
-    return dataElements
-}
-
-export const getDataElementsBySection = (metadata, dataSetId, sectionId) => {
-    const section = getSections(metadata)[sectionId]
-    // need dataSetId to get dataElements from the dataSet, as there can be catCombo overrides
-    const dataSetElements = getDataElementsByDataSetId(metadata, dataSetId)
-
-    return section.dataElements.map((deId) =>
-        dataSetElements.find((dse) => dse.id === deId)
-    )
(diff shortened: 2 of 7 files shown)


src/metadata/selectors.js
```diff
@@ -1,225 +1,181 @@
...
-    const allOptionsLists = categories.map((cat) => cat.categoryOptions)
-    // compute the  combination of category-catgoryOption- this basically computes the categoryCombinations
-    // result is a matrix of categoryOptions for each column in the categoryCombo
-    const optionCombinationOrder = cartesian(allOptionsLists)
-    return optionCombinationOrder.map((opts) =>
-        getCoCByCategoryOptions(metadata, catComboId, opts)
-    )
-}
+
+export const selectDataElementsByCatComboInOrder = createSelector(
```
Birkbjo, line 103: I think "group" was a pretty integral part of what this function does, and it should not be removed from the name. I think something like `selectGroupedDataElements...` or something would better reflect the intention and result of this function.

I can see why you've renamed the functions, but in general I would be cautious about refactorings like this. A lot of these files are "hot" and are being worked on by others, and renaming them or refactoring stuff creates a lot of merge conflicts to resolve. It's so easy to get into code and think "this could've been done a better way", or "lets follow this naming convention", but I think we should try to resist this urge, especially when there are multiple people working on a project simultaneously.

src/metadata/selectors.js
```diff
@@ -1,225 +1,181 @@
...
+        )
+    }
+)
+
+/**
+ * Tries to find the categoryOptionCombo with the given categoryOptions within a category
+ * combination.
+ */
+
+export const selectCoCByCategoryOptions = createCachedSelector(
```
Birkbjo, line 157: The signature of the functions are not really obvious anymore. So I think we should make an effort to comment with JSDOC or something.

ismay: Yeah that sounds good.

src/metadata/selectors.js
Birkbjo: Yeah, I totally understand the rationale, and again not a problem. I think we just have a different perspective on stuff like this. I'm very much a "if it ain't broke, don't fix it kind of guy", and don't really touch code that I don't have to - which is probably just something I've learned through working on the Maintenance app. But there's a strong argument for "fixing broken windows" as well, and do what you think it best - but there's a fine balance. I do think this is particular refactor is justified though.

> For example, everything in selectors.js was exported, but a fair amount of exports were unused.

I really don't see this is a problem. I exported all functions because we will probably need to use the other ones down the line. The selectors are basically an API to the metadata, and I don't see why we should "open"/export this on a case by case basis as they are used in the app.

src/metadata/selectors.js
```diff
@@ -1,225 +1,181 @@
...
-    metadata,
-    catComboId
-) => {
-    const categories = getCategoriesByCategoryComboId(metadata, catComboId)
-
-    // get options for each category, these should be in sorted order
-    const allOptionsLists = categories.map((cat) => cat.categoryOptions)
-    // compute the  combination of category-catgoryOption- this basically computes the categoryCombinations
-    // result is a matrix of categoryOptions for each column in the categoryCombo
-    const optionCombinationOrder = cartesian(allOptionsLists)
```
Birkbjo, line 190: Wheres the equal of this selector? The cartesian product is integral to rendering the dataValues in correct order.

src/metadata/selectors.js
```diff
@@ -1,225 +1,181 @@
...
  */
-export const getCoCByCategoryOptions = (
-    metadata,
-    catComboId,
-    categoryOptionIds
-) => {
-    const cocs = getCategoryOptionCombosByCategoryComboId(metadata, catComboId)
-    const sorted = [...categoryOptionIds].sort() //sort to check for equality
-
-    for (const coc of cocs) {
```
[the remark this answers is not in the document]
ismay: 😋

src/metadata/selectors.js
ismay: I'll resolve it, let me know if there's still stuff that needs to be addressed!

src/metadata/selectors.js
ismay: I'll resolve this for now. We can also talk tomorrow and figure out the specifics if there's anything that still needs addressing. Or let me know here in the meantime, good to talk about the coordination of these types of changes.

src/custom-forms/use-custom-forms.js
```diff
@@ -1,15 +1,20 @@
 import { useQuery } from 'react-query'
+import { createSelector } from 'reselect'
 import keys from './query-key-factory.js'
 
+const getDataSets = (data) => data.dataSets
+const getCustomForms = createSelector(getDataSets, (dataSets) =>
```
Mohammer5, line 6: Why use `createSelector` here? it's just being used by the hook once data has been retrieved, not in shared hooks

Mohammer5 (commented):
I guess it'd make sense to add jest tests for the selectors at this point. There's no need to test the memoization, but testing whether they return what's expected makes sense, especially now that we don't have the deadline time pressure anymore

ismay:
> I guess it'd make sense to add jest tests for the selectors at this point. There's no need to test the memoization, but testing whether they return what's expected makes sense, especially now that we don't have the deadline time pressure anymore

I agree. @Birkbjo, shall we add the tests on your branch? Then we can know with certainty that these changes won't break any of the existing functionality.

Mohammer5:
> I agree. @Birkbjo, shall we add the tests on your branch? Then we can know with certainty that these changes won't break any of the existing functionality.

That sounds like a good idea!

Birkbjo:
> I agree. @Birkbjo, shall we add the tests on your branch? Then we can know with certainty that these changes won't break any of the existing functionality. 

The new DataSetMetadata branch? There's not too many changes there, but I can work to merge those into this when this is merged. And then we can add tests there before we merge that branch, sure!

Birkbjo (approved):

src/metadata/selectors.js
```diff
@@ -1,229 +1,214 @@
...
-            sorted.length === sortedCatOptions.length &&
-            sortedCatOptions.every((id, ind) => id === sorted[ind])
-        ) {
-            return coc
+export const getCoCByCategoryOptions = createCachedSelector(
+    getCategoryOptionCombosByCategoryComboId,
+    (_, categoryComboId) => categoryComboId,
+    (_, __, categoryOptionIds) => categoryOptionIds,
+    (categoryOptionCombos, categoryComboId, categoryOptionIds) => {
+        // Sort since we only care about the contents being identical, not the order
```
mediremi, line 195: Instead of sorting arrays, we could use `Set`s? (i.e. `const targetIds = new Set(categoryOptionIds)`)

ismay: Do you know if that's safe to use with our browser support? I've not looked into that recently

mediremi: Even IE11 supports it according to [this MDN table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#browser_compatibility) so should be alright imo

ismay: Ah ok nice. Yeah should be fine then

ismay: After taking a look at the implementation that used Set I went in a slightly different direction. Because we don't need the sorting at all actually.

Since we don't care about the order, it's sufficient to check if the arrays have the same length and that all ids in targetIds are included in currentIds. That yielded a solution that I think is simple enough and quite readable. I can imagine the Set might be more performant, but I felt readability trumped performance here.

PR #48, opened 2022-03-14
test(metadata-selectors): add tests for selectors

ismay:
Writing tests based on https://github.com/dhis2/data-entry-app/pull/46, to ensure that the changes in https://github.com/dhis2/data-entry-app/pull/47 haven't broken any existing functionality.

Todo:

- [x] getSectionById
- [x] getCategoriesByCategoryComboId
- [x] getCategoryOptionsByCategoryId
- [x] getCategoryOptionCombosByCategoryComboId
- [x] getDataElementsByDataSetId
- [x] getDataElementsBySection
- [x] getGroupedDataElementsByCatComboInOrder
- [x] getGroupedDataElementsByCatCombo
- [x] getCoCByCategoryOptions

### DIFF
--- a/src/metadata/selectors.test.js
+++ b/src/metadata/selectors.test.js
@@ -1,0 +1,492 @@
+import {
+    getCategories,
+    getCategoriesByCategoryComboId,
+    getCategoryById,
+    getCategoryComboById,
+    getCategoryCombos,
+    getCategoryOptionCombos,
+    getCategoryOptionCombosByCategoryComboId,
+    getCategoryOptions,
+    getCategoryOptionsByCategoryId,
+    getCoCByCategoryOptions,
+    getDataElements,
+    getDataElementsByDataSetId,
+    getDataElementsBySection,
+    getDataSetById,
+    getDataSets,
+    getSectionById,
+    getSections,
+    groupDataElementsByCatCombo,
+    groupDataElementsByCatComboInOrder,
+} from './selectors.js'
+
+describe('simple selectors', () => {
+    describe('getCategories', () => {
+        it('returns the expected data', () => {
+            const expected = 'expected'
+            const data = { categories: expected }
+
+            expect(getCategories(data)).toBe(expected)
+        })
+    })
+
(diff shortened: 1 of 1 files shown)


ismay:
By the way, I think the tests could be improved. Some of the wording is a little vague for example. But I think that we could tweak that afterwards.

src/metadata/selectors.test.js
```diff
@@ -0,0 +1,492 @@
...
+            const data = {
+                dataSets: {
+                    [dataSetId]: {
+                        sections: [expected],
+                    },
+                },
+            }
+
+            expect(getSectionById(data, dataSetId, sectionId)).toBe(expected)
+        })
```
Mohammer5, line 149: This is really a non-issue.. Lately I think that it's better to separate the data and the arguments instead of having variables for "shared" ids. I guess the reason why you went for `.toBe` is to ensure that the returned value points to the same reference, but that's not expressed by the test's description, maybe another test case could test for that or the test's description changed to something like `it('returns a reference to the expected data')`, but I don't know if that makes things better or not, maybe it's confusing this way. But as I said, this is a non-issue and more about exchanging thoughts

```suggestion
        it('returns the expected data', () => {
            const data = {
                dataSets: {
                    dataSetId: {
                        sections: [{ id: 'sectionId' }],
                    },
                },
            }
            const expected = { id: 'sectionId' }

            expect(getSectionById(data, 'dataSetId', 'sectionId')).toEqual(expected)
        })
```

ismay: Hey, thanks for the review. I was talking to Birk about how we wanted to approach this, and we settled on something slightly different. I've rebased the tests on development here: https://github.com/dhis2/data-entry-app/pull/49. We can merge that in, rebase my memoization branch on that, and then after that rebase Birk's branch.

That way we can be sure that behaviour on dev. won't be broken by my memoization changes.

If you don't mind, let me know what you think of: https://github.com/dhis2/data-entry-app/pull/49.

PR #50, opened 2022-03-29
feat:  dataset metadata endpoint

Birkbjo:
This implements the new dataSet endpoint `/dataSetMetadata`. This replaces the need for the `/metadata` with selective filters. This has resulted in some changes in the structure of the metadata, most notably that `categoryOptionCombos` do not exist on the root-object, these are now under their respective `categoryCombo`. Note that `attribute`-categoryCombos do not have the `categoryOptionCombos`-property, since they are not needed - see below.

The `/dataValueSets`- API also now supports `attributeCombo` and `attributeOptions`-parameters. This eliminates the need for the client to know the `attributeOptionCombo`, which simplifies our code and reduces network usage, since we don't need to know about all the `attributeOptionCombos` for matching the options. The `/dataValues`-API already supports this with `cc` and `cp`-parameters. 

~**Note: `/api/dataSetMetadata` is soon to be moved to `/api/dataEntry/metadata`. Should we await merging this until that happens?**~ This is merged and integrated in this PR.


src/metadata/use-metadata.js
```diff
@@ -1,34 +1,31 @@
...
-                    'id,skipTotal,categories~pluck,categoryOptionCombos~pluck,isDefault',
-                'categories:fields': 'id,displayFormName,categoryOptions~pluck',
-                'categoryOptions:fields':
-                    'id,displayFormName,categoryOptionCombos~pluck,categoryOptionGroups~pluck,isDefault',
-                'categoryOptionCombos:fields':
-                    'id,categoryOptions~pluck,categoryCombo,name',
-            },
-        },
-    ]
+const selectorFunction = createSelector(
```
Birkbjo, line 5: createSelector seems to be needed, or else this will be re-run for every component that uses the `useMetdata`-hook. Ie. reference to the "selector"-function does not seem to be shared across component instances.

src/data-workspace/category-combo-table.js
```diff
@@ -47,7 +47,7 @@ export const CategoryComboTable = ({
 
     if (
         computedCategoryOptions.length !==
-        categoryCombo.categoryOptionCombos.length
+        (categoryCombo.categoryOptionCombos?.length || 1)
```
Mohammer5, line 50: This is a [magic number](https://eslint.org/docs/rules/no-magic-numbers). My preference would be to avoid that and use a descriptive name for what this value actually stands for (+ maybe a comment, because I'm confused by the `1`; if the variable name is not expressive enough)

src/data-workspace/section-form/section.js
```diff
@@ -38,7 +38,7 @@ export const SectionFormSection = ({ section, globalFilterText }) => {
 
     const maxColumnsInSection = Math.max(
         ...groupedDataElements.map(
-            (grp) => grp.categoryCombo.categoryOptionCombos.length
+            (grp) => grp.categoryCombo.categoryOptionCombos?.length || 1
```
Mohammer5, line 41: Same as in comment above

src/data-workspace/use-attribute-option-combo.js
```diff
@@ -1,43 +1,52 @@
...
 // should not need this when api support CC and CP instead of cocId
 export const useAttributeOptionCombo = () => {
     const { isLoading, isError, data } = useMetadata()
+    const attributeMetadata = useAttributeMetadata()
     const [{ dataSetId, attributeOptionComboSelection }] = useContextSelection()
 
     const cocId = useMemo(() => {
-        if (!isLoading && !isError && dataSetId) {
+        const hasData = data && attributeMetadata.data
+        if (hasData && dataSetId) {
```
Mohammer5, line 19: Why remove the `!isLoading && !isError && ` part? At some point we'll add a UI so the user can refetch metadata? Or is that irrelevant for this part and we want to show stale data?

src/data-workspace/use-attribute-option-combo.js
```diff
@@ -1,43 +1,52 @@
...
         }
 
         return null
-    }, [dataSetId, attributeOptionComboSelection, data, isLoading, isError])
+    }, [
+        dataSetId,
+        attributeOptionComboSelection,
+        data,
+        isLoading,
+        isError,
```
Mohammer5, line 47: Not being used inside the hook anymore

src/metadata/use-metadata.js
```diff
@@ -1,34 +1,31 @@
...
-        select: (data) => hashArraysInObject(data),
-    })
+const queryOpts = {
+    refetchOnMount: false,
+    select: selectorFunction,
+}
+
+export const useMetadata = () => {
+    const metadataQuery = useQuery(queryKey, queryOpts)
+    return metadataQuery
```
Mohammer5, line 30: Just a suggestion, take it or leave it. To me it's more parseable this way:

```suggestion
export const useMetadata = () => {
    const metadataQuery = useQuery(queryKey, {
        refetchOnMount: false,
        select: selectorFunction,
    })
    return metadataQuery
```

src/metadata/use-metadata.js
```diff
@@ -1,34 +1,31 @@
...
+const selectorFunction = createSelector(
+    (data) => data,
+    (data) => {
+        // bug in API includes duplicate "default"-catCombo
+        // one has categoryOptionCombos, and the other is missing properties
+        // delete the one that is missing properties, so we ensure that the correct one is used
+        const filtered = data.categoryCombos.filter(
+            (cc) => !(cc.isDefault && !cc.categoryOptionCombos)
+        )
+        const filteredData = { ...data, categoryCombos: filtered }
```
Mohammer5, line 14:
```suggestion
        const categoryCombos = data.categoryCombos.filter(
            (cc) => !(cc.isDefault && !cc.categoryOptionCombos)
        )
        const filteredData = { ...data, categoryCombos }
```

src/metadata/use-metadata.js
```diff
@@ -1,34 +1,31 @@
...
+        const result = hashArraysInObject(filteredData)
+        console.log(result)
```
Mohammer5, line 16: I suppose @Birkbjo is not done cleaning this up yet, just adding this comment so I'll remember it

src/data-workspace/category-combo-table.js
Birkbjo: Yeah I think I will remove this. Some categoryCombos are not retuned with the `categoryOptionCombos`from the API. This should just be attribute-combos so it shouldn't affect this part of the app. But a bug in the API made the default one also not return it, and made the app crash, so this was just a simple fix. I fixed the default-problem another way in `use-metadata.js`

src/data-workspace/section-form/section.js
Birkbjo: I think it should stay here though. This is basically just defensive coding, and means if no group has the `categoryOptionCombos` due to some weird bug, we still get one column. This could however probably be moved to outside of the map, So its `Math.max(...) || 1`.

src/data-workspace/use-attribute-option-combo.js
Birkbjo: Basically because it's not needed, we only care about the data. `isLoading` is only true in case of "hard" loading, eg. no cached data, so "first" load. Therefore data is undefined as well. If we want the behaviour you are talking about we would want to use `isRefetching`.

src/metadata/use-metadata.js
Birkbjo: This was done to ensure the stability of the object. Mostly due to testing due to some weird behaviour, and I left it as is.

src/data-workspace/category-combo-table.js
Mohammer5: With the fix that Lars did last work, could this `1` be removed completely then?

src/data-workspace/section-form/section.js
Mohammer5: I see, maybe then this expresses the intention a bit better:

```js
    // Never less than 1
    const maxColumnsInSection = Math.max(
        ...groupedDataElements.map(
            (grp) => grp.categoryCombo.categoryOptionCombos?.length
        ),
        1
    )
```

Mohammer5: There's a `ignoreDefaultValues` option in the magic number rule set, so I'd say you're right, this does make sense here

src/data-workspace/use-attribute-option-combo.js
```diff
@@ -1,43 +1,31 @@
...
-    getDataSetById,
-    getCategoryComboById,
-} from '../metadata/selectors.js'
+import { getDataSetById, getCategoryComboById } from '../metadata/selectors.js'
 
-// TODO: this should probably be handled by useContextSelection-hook
-// should not need this when api support CC and CP instead of cocId
-export const useAttributeOptionCombo = () => {
-    const { isLoading, isError, data } = useMetadata()
+export const useAttributeParams = () => {
```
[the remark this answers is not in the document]
Mohammer5: I think we should move it into the `context-selection` namespace as it returns a property called `validSelection`

Mohammer5 (commented):
I just realized I have to review first before you can re-request review, so just ignore this comment

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
```diff
@@ -0,0 +1,38 @@
...
+        const selectedOptions = Object.values(attributeOptionComboSelection)
+        // all categories in catCombo should have a selection
+        const validSelection =
+            categoryCombo.categories.length === selectedOptions.length
+
+        return {
+            attributeCombo: catComboId,
+            attributeOptions: selectedOptions,
+            validSelection,
+        }
```
Mohammer5, line 30: I think the comparison and `validSelection` return value belong to the `useIsValidSelection` hook, not into this one

src/context-selection/use-context-selection/use-is-valid-selection.js
```diff
@@ -0,0 +1,12 @@
+import { useAttributeParams } from '../attribute-option-combo-selector-bar-item/index.js'
+import { useContextSelection } from '../use-context-selection/index.js'
+
+export function useIsValidSelection() {
+    const [selection] = useContextSelection()
+    const { validSelection } = useAttributeParams()
```
Mohammer5, line 6: Should be determined in this hook, not in `useAttributeParams`

src/context-selection/use-context-selection/use-context-selection.js
```diff
@@ -27,6 +27,14 @@ export function usePeriodId() {
     return useQueryParam('periodId', PARAMS_SCHEMA.periodId)
 }
 
+/**
+ *
+ * @returns the selected attribute-options in the shape of
+ * ```{
+ *   categoryId: categoryOptionId
+ * }```
```
Mohammer5, line 35:
```suggestion
 * ```{
 *   [categoryId: String]: String // categoryOptionId
 * }```
```

Mohammer5 (changes requested):
This looks very good already :muscle:! I left a couple of comments and have one general remark:

I think we can also call the `useMetadata` hook in the `app.js` and show the loading spinner / error there. We'll need the metadata in the context selection anyway. Then we could ignore `isLoading` and `isError` returned by the `useMetadata` hook in all other components as they wouldn't get rendered (in `SectionFormSection`, `CategoryComboTable`, `DataWorkspace`, `DefaultForm` & `useDataValueParams`)

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
Birkbjo: I'm not sure I agree. That would mean we have to duplicate this code in `useIsValidSelection`-hook. We already have access to the categoryCombo in this hook, so I choose to expose `validSelection`, and use that in the `useIsValidSelection`-hook instead. If not we have to introduce metadata and the selectors in the other hook.

src/metadata/use-metadata.js
```diff
@@ -1,34 +1,24 @@
...
+
+const queryKey = [`/dataEntry/metadata`]
 
-    return useQuery(queryKey, {
-        select: (data) => hashArraysInObject(data),
-    })
+const queryOpts = {
+    refetchOnMount: false,
+    select: selectorFunction,
+    staleTime: 24 * 60 * 1000,
```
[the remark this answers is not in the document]
Birkbjo: Yeah, good catch! We will probably adjust this in the future as well.

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
Mohammer5: > That would mean we have to duplicate this code in useIsValidSelection-hook

You're talking about the code to get the `categoryCombo`, right? I don't think it's an issue. If you prefer to keep the code dry, you could extract that into its own hook or (cached?) selector.

The reason why I prefer changing this is because I think the conceptual model is more important that a bit of code duplication or creating an extra hook or selector for this. Conceptually `validSelection` is just somewhat related to "attribute params", but not derived from it, even though we have all the values necessary for the `validSelection` calculation. `useAttributeParams` does not indicate that it will return `validSelection`, plus it does two things at the same time, which is an odd and normally discouraged code pattern (despite being a very mild issue in this case)

> If not we have to introduce metadata and the selectors in the other hook.

I don't see the issue with that, we're talking about 5 or 6 lines of code, right? And the data itself comes from memoized selectors.

src/data-workspace/category-combo-table.js
```diff
@@ -47,7 +47,7 @@ export const CategoryComboTable = ({
 
     if (
         computedCategoryOptions.length !==
-        categoryCombo.categoryOptionCombos.length
+        categoryCombo.categoryOptionCombos?.length
```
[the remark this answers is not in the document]
Birkbjo: Yeah, it shouldn't really happen in practice. But it might happen if the category is somehow an `attribute`-categoryCombo.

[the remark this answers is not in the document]
Mohammer5: Alright, I'll create a jira issue for that so that it gets fixed on the BE side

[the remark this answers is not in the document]
Birkbjo: Nono, it's deliberate. But this is just to prevent the app from crashing if a attribute-combo would be selected, but that shouldn't really happen. This is mostly a relic from before the API was fixed, and default-catcombo did not return the `categoryOptionCombos`.

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
Birkbjo: > Conceptually validSelection is just somewhat related to "attribute params", but not derived from it 
It actually is this. The `validSelection` here refers to if the `attrubute-params` is valid, not the entire context-selection. 

I can change this though, as it was mostly because I implemented this way of checking if it's valid just for attribute-params before I implemented the `useIsValidSelection`-hook, so I just figured I could re-use that code...

>  plus it does two things at the same time, which is an odd and normally discouraged code pattern

I can see that it comes off that way now, since we have the `useIsValidSelection`-hook, but previously I used this together with getting the `attributeCombo` and `attributeOptions`, which made this a very ergonomic API. 

I don't think custom hooks necessarily should "just do one thing", or that this case even fulfills that - it's just a helper., or some metadata about the data you return, if you will. Especially when it returns an object either way, it's no problem imo to return a helper property like this, when you have all the orchestration to calculate that prop around already.

Mohammer5: > not the entire context-selection

Ah, that's my bad then! I thought it's vor the entire context as the other hook was called `useIsValidSelection`. I saw that you already changed the `useIsValidHook` to include the other parameters as well :muscle: 

> I don't think custom hooks necessarily should "just do one thing", or that this case even fulfills that - it's just a helper, or some metadata about the data you return, if you will.

So, considering that I thought `validSelection` refers to the entire selection being valid, I would've disagreed with this, but it doesn't matter for this PR anymore as I misunderstood the purpose of the value. IMO even helpers should never do more than the name suggests. Whether the attribute params are valid or not could be seen as meta info of the params themselves, so I'd say the hooks still does "one thing" in this case, so it's fine as long as the other hook also considers the other context values.

---

I'll resolve this conversation as there's now a clear distinction between the return value of `useIsValidSelection` and the `validSelection` prop returned by the `useAttributeParams` hook.

src/data-workspace/category-combo-table.js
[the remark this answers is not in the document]
Mohammer5: I see! Could you add a comment to the code explaining what you just explained to me?

src/metadata/use-metadata.js
```diff
@@ -1,34 +1,24 @@
...
-                    'id,skipTotal,categories~pluck,categoryOptionCombos~pluck,isDefault',
-                'categories:fields': 'id,displayFormName,categoryOptions~pluck',
-                'categoryOptions:fields':
-                    'id,displayFormName,categoryOptionCombos~pluck,categoryOptionGroups~pluck,isDefault',
-                'categoryOptionCombos:fields':
-                    'id,categoryOptions~pluck,categoryCombo,name',
-            },
-        },
-    ]
+const selectorFunction = createSelector(
```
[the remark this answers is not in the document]
Mohammer5: I thought about this a bit more and I think you're right as in that this different from the selectors in the `selectors.js`.

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
Birkbjo: > IMO even helpers should never do more than the name suggests. Whether the attribute params are valid or not could be seen as meta info of the params themselves, so I'd say the hooks still does "one thing" in this case, so it's fine as long as the other hook also considers the other context values. 

When I mentioned "helper", I referred to the `isValid`-property like you suggest here. Not the entire hook. See now that is not so clear :)

src/data-workspace/category-combo-table.js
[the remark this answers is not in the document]
Birkbjo: It is kind of mentioned for the selector here: https://github.com/dhis2/data-entry-app/pull/50/files#diff-0f4a3b4e7a1af6e32767e349e7bdc6c920e1c9606a48dce49d0d00db87155bcfR45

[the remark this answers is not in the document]
Mohammer5: Ah, nice

Birkbjo:
> I think we can also call the `useMetadata` hook in the `app.js` and show the loading spinner / error there. We'll need the metadata in the context selection anyway. Then we could ignore `isLoading` and `isError` returned by the `useMetadata` hook in all other components as they wouldn't get rendered (in `SectionFormSection`, `CategoryComboTable`, `DataWorkspace`, `DefaultForm` & `useDataValueParams`)

I made a new component `LoadApp` to handle this, as I didn't want to block rendering of the `Router`, `CssVariables` etc that are initialized in `app.js`.

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
```diff
@@ -15,10 +15,16 @@ export const useAttributeParams = () => {
         const categoryCombo = selectors.getCategoryComboById(data, catComboId)
 
         const selectedOptions = Object.values(attributeOptionComboSelection)
+
+        if (categoryCombo.isDefault) {
+            // we don't need to supply attributeCombo/options when it's default
+            return {
+                validSelection: true,
+            }
```
Mohammer5, line 23: I'd argue that - despite not being necessary - the returned shape should always be the same.
The values for `attributeCombo` and `attributeOptions` are not undefined here, and I think we should change it for the default return value as well as then components won't have to do "null checks" as long as they rely on the `validSelection`

```suggestion
            return {
                attributeCombo: '',
                attributeOptions: [],
                validSelection: true,
            }
```

Mohammer5 (changes requested):
Btw: The `24 * 60 * 1000` is still present.
Regarding the other comment with the suggestion I made, what was your decision there? Leave it as is?

src/context-selection/attribute-option-combo-selector-bar-item/use-attribute-params.js
Birkbjo: When undefined these will be omitted from the request when using these. I could set these to `undefined` as well, but we shouldn't need to do any null checks with these values - as they should be sent as is.

Birkbjo:
> Btw: The `24 * 60 * 1000` is still present. Regarding the other comment with the suggestion I made, what was your decision there? Leave it as is?

Yes, because then we know that these values are stable and won't cause re-renders. It probably works as expected with the new object directly in `useQuery` but I just didn't really see the benefit of changing it, so I left it as is.

src/context-selection/use-context-selection/use-is-valid-selection.js
```diff
@@ -1,12 +1,31 @@
...
+            validAttributeComboSelection =
+                categoryCombo.categories.length === selectedOptions.length
+        }
+    }
+
+    if (orgUnitId && periodId && validAttributeComboSelection) {
         return true
     }
+
     return false
```
Mohammer5, line 30: I don't think we need to check whether `data` exists or not, the `<LoadApp />` component takes care of that, right?

```suggestion
    if (!dataSetId || !orgUnitId || !periodId) {
        return false
    }
    
    const dataSet = selectors.getDataSetById(data, dataSetId)
    const catComboId = dataSet.categoryCombo.id
    const categoryCombo = selectors.getCategoryComboById(data, catComboId)
    
    if (categoryCombo.isDefault) {
        return true
    }

    const selectedOptions = Object.values(attributeOptionComboSelection)
    return categoryCombo.categories.length === selectedOptions.length
```

src/context-selection/attribute-option-combo-selector-bar-item/index.js
```diff
@@ -1,2 +1,2 @@
 export { default as AttributeOptionComboSelectorBarItem } from './attribute-option-combo-selector-bar-item.js'
-export { useAttributeParams } from './use-attribute-params.js'
+export { useApiAttributeParams } from '../../data-workspace/use-api-attribute-params.js'
```
Mohammer5, line 2: This code artifact can be deleted, I suppose VS Code did this "magically" when moving the file

Mohammer5 (changes requested):

PR #59, opened 2022-05-06
refactor(context selection): use metadata from "useMetadata" hook

Mohammer5:
TECH-1074 ([link](https://dhis2.atlassian.net/browse/TECH-1074))


src/context-selection/attribute-option-combo-selector-bar-item/attribute-option-combo-selector-bar-item.js
```diff
@@ -26,62 +28,66 @@ const useSetSelectionHasNoFormMessage = (
...
+    return categoryWithNoOptionsExists
 }
 
 export default function AttributeOptionComboSelectorBarItem({
     setSelectionHasNoFormMessage,
 }) {
+    const { data: metadata } = useMetadata()
+    const [dataSetId] = useDataSetId()
+    const dataSets = selectors.getDataSets(metadata) || ''
+    const dataSet = dataSets[dataSetId]
```
Birkbjo, line 42: Why not use `selectors.getDataSetById`?

src/context-selection/attribute-option-combo-selector-bar-item/attribute-option-combo-selector-bar-item.js
```diff
@@ -26,62 +28,66 @@ const useSetSelectionHasNoFormMessage = (
...
+    const categoryComboId = dataSet?.categoryCombo.id
+    const categoryCombos = selectors.getCategoryCombos(metadata)
+    const categoryCombo = categoryCombos[categoryComboId]
```
Birkbjo, line 45: `selectors.getCategoryComboById` ?

src/context-selection/attribute-option-combo-selector-bar-item/use-categories-with-options-within-period.js
```diff
@@ -0,0 +1,86 @@
...
+    const { data: metadata } = useMetadata()
+    const [periodId] = usePeriodId()
+    const [dataSetId] = useDataSetId()
+
+    if (!dataSetId || !periodId) {
+        return []
+    }
+
+    const dataSets = selectors.getDataSets(metadata) || ''
+    const dataSet = dataSets[dataSetId]
```
Birkbjo, line 60: Same here; use `selectors.getDataSetById`

src/context-selection/data-set-selector-bar-item/data-set-selector-bar-item.js
```diff
@@ -1,82 +1,94 @@
...
+    // Select the first item if there's only one
+    useEffect(
+        () => {
+            if (dataSets.length === 1) {
+                setDataSetId(dataSets[0].id)
+            }
+        },
+        // We need to do this only once
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+        []
```
Birkbjo, line 70: What if metadata is reloaded? Shouldn't this run again then?

src/context-selection/data-set-selector-bar-item/data-set-selector-bar-item.js
```diff
@@ -1,82 +1,94 @@
...
+    const [dataSetOpen, setDataSetOpen] = useState(false)
+    const [dataSetId, setDataSetId] = useDataSetId()
+    const dataSets = Object.values(selectors.getDataSets(metadata))
+    const selectableDataSets = dataSets
+        .map(({ id, displayName }) => ({
+            label: displayName,
+            value: id,
+        }))
+        .sort((left, right) => left.label.localeCompare(right.label))
+    const dataSet = dataSets.find(({ id }) => id === dataSetId)
```
Birkbjo, line 59: Use `selectors.getDataSetById(metadata, dataSetId)`

src/context-selection/period-selector-bar-item/period-selector-bar-item.js
```diff
@@ -17,10 +17,12 @@ export const PeriodSelectorBarItem = () => {
     const [periodOpen, setPeriodOpen] = useState(false)
     const [periodId, setPeriodId] = usePeriodId()
     const [dataSetId] = useDataSetId()
-    const dataSetPeriodType = useDataSetPeriodType()
+    const { data: metadata } = useMetadata()
+    const dataSet = metadata.dataSets[dataSetId]
```
Birkbjo, line 21:
```suggestion
    const dataSet = selectors.getDataSetById(metadata, dataSetId)
```

src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
```diff
@@ -12,38 +15,34 @@ export default function SectionFilterSelectorBarItem() {
     const [open, setOpen] = useState(false)
     const [sectionFilter, setSectionFilter] = useSectionFilter()
     const deselect = () => setSectionFilter(undefined)
-    const dataSetSectionsInfo = useDataSetSectionsInfo()
-    const loading = !dataSetSectionsInfo.called || dataSetSectionsInfo.loading
+    const { data: metadata } = useMetadata()
+    const [dataSetId] = useDataSetId()
+    const dataSet = selectors.getDataSets(metadata)[dataSetId]
```
Birkbjo, line 20:
```suggestion
    const dataSet = selectors.getDataSetById(metadata, dataSetId)
```

src/context-selection/section-filter-selector-bar-item/use-should-component-render-null.js
```diff
@@ -1,20 +1,22 @@
...
- * should not render `null`, then it's most likely not going to change when
- * reloading
- */
 export default function useShouldComponentRenderNull() {
-    const dataSetSectionsInfo = useDataSetSectionsInfo()
+    const [dataSetId] = useDataSetId()
+    const [orgUnitId] = useOrgUnitId()
+    const [periodId] = usePeriodId()
+    const { data: metadata } = useMetadata()
+    const dataSet = selectors.getDataSets(metadata)[dataSetId]
```
Birkbjo, line 13:
```suggestion
    const dataSet = selectors.getDataSetById(metadata, dataSetId)
```

Birkbjo (changes requested):
Overall looks good, just some places where I think it's better to use selectors.

src/metadata/selectors.js
```diff
@@ -85,6 +85,29 @@ export const getCategoriesByCategoryComboId = createCachedSelector(
...
+ * @param {*} metadata
+ * @param {*} dataSetId
+ */
+export const getCategoryComboByDataSetId = createCachedSelector(
+    getDataSetById,
+    getCategoryCombos,
+    (dataSet, categoryCombos) => {
+        return categoryCombos?.[dataSet.categoryCombo.id]
+    }
+)((_, categoryComboId) => categoryComboId)
```
Birkbjo, line 98: Note that the `categoryComboId` variable is actually `dataSetId`. It has the same signature as the function - and generated a key based on that. This works though, but confusing name.

src/metadata/selectors.js
```diff
@@ -85,6 +85,29 @@ export const getCategoriesByCategoryComboId = createCachedSelector(
...
+
+/**
+ * @param {*} metadata
+ * @param {*} dataSetId
+ */
+export const getCategoriesByDataSetId = createCachedSelector(
+    getCategoryComboByDataSetId,
+    getCategories,
+    (categoryCombo, categories) =>
```
Birkbjo, line 107: There is a selector for `getCategoriesByCategoryComboId`, so you could probably do something like

```javascript
export const getCategoriesByDataSetId = createCachedSelector(
     metadata => metadata, 
    getCategoryComboByDataSetId,
    (metadata, categoryCombo) => getCategoriesByCategoryComboId(metadata, categoryCombo.id)
```

Not sure if this is any cleaner though, since `getCategoryComboByDataSetId` - returns the object and not just the ID.

Also it doesn't look like that function is checking for the existence of the `categoryCombo` - which I think it should. 
So if you decide to use that, I think we should add the defensive coding there as well.

src/metadata/selectors.js
```diff
@@ -85,6 +85,29 @@ export const getCategoriesByCategoryComboId = createCachedSelector(
...
+        categoryCombo?.categories.map((id) => categories[id]) || []
+)((_, categoryComboId) => categoryComboId)
```
Birkbjo, line 109:
```suggestion
)((_, dataSetId) => dataSetId)
```

src/metadata/use-metadata.js
```diff
@@ -13,9 +13,6 @@ const queryOpts = {
     refetchOnMount: false,
     select: selectorFunction,
     staleTime: 24 * 60 * 60 * 1000,
-    meta: {
```
Birkbjo, line 16: Isn't this needed to persist the metadata between reloads?

Birkbjo (changes requested):

src/metadata/use-metadata.js
Mohammer5: No, the SW caches it already. The philosophy in this app is:
* Everything that we mutate will be persisted by react-query
* Everything that we fetch only, but never mutate, can be handled by the SW

This way we have a clear separation for when to use what. We only really need the react-query cache when we want to do optimistic updates, which is impossible with the SW. And we don't need to do that for metadata. At least that's my current understanding.

Mohammer5:
@Birkbjo thanks for the suggestions, I've renamed the variables and add null-checks & warnings. I also added tests for the new selectors, I forgot about that..

src/metadata/use-metadata.js
Birkbjo: Right, it's just that this didn't seem to work as expected, and I got instant load-times when adding this to react-query. It also means that we have more (at least) easier control of the caching - eg. it would be very easy to refetch and clear the cache. I guess it's something we should investigate further, especially regarding e-tags, to have the caching work properly. Lars is aware of that though.

Mohammer5: Yeah, so ideally SW uses etags to determine whether it should use cached results or not. Right now the SW fetches everything everything while online, it's an offline-support tool. IMO caching the result should be handled through SW/browser caches in the network layer if etags work as expected.

If this is not the case, I could create in issue so we track it as I'd regard that as essential and strongly prefer if we don't add a workaround in the FE and fix it on the server side instead.

Birkbjo (approved):

PR #60, opened 2022-05-06
fix: use-query-param replaceIn instead of pushIn

Birkbjo:
While making selections in the context-selection, the history stack is "spammed" with new entries when making initial selections. This changes it so new entries are only pushed when it's a valid selection - reducing entries in the history that are not really useful.

### DIFF
--- a/src/context-selection/use-context-selection/use-context-selection.js
+++ b/src/context-selection/use-context-selection/use-context-selection.js
@@ -6,6 +6,7 @@ import {
     useQueryParam,
     withDefault,
 } from 'use-query-params'
+import { useIsValidSelection } from './use-is-valid-selection.js'
 
 export const PARAMS_SCHEMA = {
     dataSetId: StringParam,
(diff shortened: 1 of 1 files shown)


src/context-selection/use-context-selection/use-context-selection.js
```diff
@@ -15,16 +16,30 @@ export const PARAMS_SCHEMA = {
     sectionFilter: StringParam,
 }
 
+const useCustomQueryParam = (name, schema) => {
+    const [paramValue, setParamValue] = useQueryParam(name, schema)
+    const isValidSelection = useIsValidSelection()
```
Mohammer5, line 21: How does it know whether it's a valid selection before the change has been pushed to `use-query-param`.
So for example a data set has a default cat combo, when the data set id and the org unit id is already present and the user selects a period, when the logic reaches this point, the `useIsValidSelection` hook doesn't know about the period id yet, does it?

HendrikThePendric:
> While making selections in the context-selection, the history stack is "spammed" with new entries when making initial selections. This changes it so new entries are only pushed when it's a valid selection - reducing entries in the history that are not really useful.

This seems like it would be a very nice improvement to add @Birkbjo. Are we deferring this due to time constraints?

Mohammer5:
@Birkbjo could you just write a quick explanation on why this is neither continued nor merged? Couple of sentences should suffice

Mohammer5:
By @Birkbjo:
> also something we might investigate again, but pretty low priority, might not be something we need

PR #65, opened 2022-06-02
feat: render simple custom forms

HendrikThePendric:
This PR includes quite an extensive [markdown file](https://github.com/dhis2/data-entry-app/blob/af3719a6085df11c70f2f508a94c745600857ecc/docs/custom-forms.md). Please refer to that for some more context. A list of custom datasets using custom forms (on debug/dev) can be found [here](https://debug.dhis2.org/dev/api/39/dataSets?fields=id,displayName&filter=formType:eq:CUSTOM).


src/data-workspace/custom-form/replace-input-node.js
```diff
@@ -0,0 +1,43 @@
...
+    if (inputType === INPUT_TYPES.ENTRYFIELD) {
+        const [deId, cocId] = domNode.attribs.id.split('-')
+        const dataElement = getDataElementById(metadata, deId)
+        return (
+            <DataEntryField
+                dataElement={dataElement}
+                categoryOptionCombo={{ id: cocId }}
+            />
+        )
+    }
```
Mohammer5, line 42:
```suggestion
    if (inputType !== INPUT_TYPES.ENTRYFIELD) {
        return undefined
    }
    
    const [deId, cocId] = domNode.attribs.id.split('-')
    const dataElement = getDataElementById(metadata, deId)
    return (
        <DataEntryField
            dataElement={dataElement}
            categoryOptionCombo={{ id: cocId }}
        />
    )
```

Mohammer5: Less indented code & explicit `return undefined` statement

Mohammer5 (approved):
Just a couple of suggestions, I'll leave it up to you

PR #66, opened 2022-06-02
feat: greyed fields

Birkbjo:
Adds support for `greyedFields`/disabled-fields. 

This can be activated for `DataSet sections` in the `Maintenance app`. Navigate to list of `Data Sets`, find a section-dataSet -> Hamburger-menu -> Manage sections -> Hamburger Menu in list of Sections -> Manage Grey Fields -> Click field to be disabled. 

Note that there's a backend bug where setting a greyField for a dataElement with `categoryComb` `None/default` results in a 500-error. See [DHIS2-13371](https://jira.dhis2.org/browse/DHIS2-13371).

### DIFF
--- a/src/data-workspace/category-combo-table/category-combo-table.js
+++ b/src/data-workspace/category-combo-table/category-combo-table.js
@@ -16,6 +16,7 @@ import {
     DataEntryField,
     useActiveCell,
 } from '../data-entry-cell/index.js'
+import { getFieldId } from '../get-field-id.js'
 import styles from './category-combo-table.module.css'
 import { TotalHeader, ColumnTotals, RowTotal } from './total-cells.js'
 
@@ -24,6 +25,7 @@ export const CategoryComboTable = ({
     dataElements,
     filterText,
     globalFilterText,
+    greyedFields,
     maxColumnsInSection,
     renderRowTotals,
     renderColumnTotals,
@@ -119,6 +121,14 @@ export const CategoryComboTable = ({
         return isThisTableActive && idxDiff < headerColSpan && idxDiff >= 0
     }
 
+    const isFieldGreyed = (de, coc) => {
+        if (!greyedFields) {
+            return false
+        }
+        const key = getFieldId(de, coc)
+        return greyedFields[key] || false
+    }
+
     return (
         <TableBody>
             {rowToColumnsMap.map((colInfo, colInfoIndex) => {
@@ -174,6 +184,7 @@ export const CategoryComboTable = ({
                                 <DataEntryField
                                     dataElement={de}
                                     categoryOptionCombo={coc}
+                                    disabled={isFieldGreyed(de, coc)}
                                 />
                             </DataEntryCell>
                         ))}
@@ -233,6 +244,8 @@ CategoryComboTable.propTypes = {
     ),
     filterText: PropTypes.string,
     globalFilterText: PropTypes.string,
+    /** Greyed fields is an object-map where [deId.cocId]: true if that field is greyed/disabled */
+    greyedFields: PropTypes.object,
     maxColumnsInSection: PropTypes.number,
     renderColumnTotals: PropTypes.bool,
     renderRowTotals: PropTypes.bool,
--- a/src/data-workspace/category-combo-table/category-combo-table.js
+++ b/src/data-workspace/category-combo-table/category-combo-table.js
@@ -121,7 +121,7 @@ export const CategoryComboTable = ({
         return isThisTableActive && idxDiff < headerColSpan && idxDiff >= 0
     }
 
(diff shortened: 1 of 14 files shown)


src/data-workspace/get-field-id.js
```diff
@@ -0,0 +1,19 @@
+export const getFieldId = (de, coc) => {
```
Birkbjo, line 1: I decided to do some minor refactoring and made a central place for this. `field.split('.')` and generation of the field-id/name was done in different, and I think it make sense to have this in one place, since we introduced more usage of this in this PR.

Birkbjo: > What do you think about naming getFieldIdComponents as parseFieldId?

Sure

> getFieldId accepts parameters (deId, cocId) (maybe even as a destructured object e.g. ({ deId, cocId })) and returns a string

Yeah, I thought about this, and also thought about supporting both - eg. checking for `typeof de === 'string'`. However since all usages used/had access to de whole object I just kept it that way to keep it simple. But we can for sure just change it to strings.

> and parseFieldId accepts a string parameter and returns an object { deId, cocId } -- (those can be spelled out or not, depending on what's more consistent with other usage)

Yeah, lets keep it like that, and change the other since;

> and parseFieldId accepts a string and returns the whole metadata objects { dataElement, categoryOptionCombo } (same parenthetical comments as above)

is not straightforward, since you would need access to the metadata.

src/data-workspace/category-combo-table/category-combo-table.js
```diff
@@ -119,6 +121,14 @@ export const CategoryComboTable = ({
         return isThisTableActive && idxDiff < headerColSpan && idxDiff >= 0
     }
 
+    const isGreyedField = (de, coc) => {
+        if (!greyedFields) {
+            return false
+        }
+        const key = getFieldId(de, coc)
+        return greyedFields[key] || false
+    }
```
Mohammer5, line 130: This could be extracted so the component doesn't have to create a new function every time it renders.

Mohammer5 (approved):
This looks very good! Just one minor suggestion, you can take it or leave it

src/data-workspace/category-combo-table/category-combo-table.js
Mohammer5: You could pass `greyedFields` as a parameter to the function. It'b being called in the same component

src/data-workspace/section-form/section.js
```diff
@@ -40,8 +41,17 @@ export const SectionFormSection = ({
...
-    const filterInputId = `filter-input-${section.id}`
 
+    const greyedFieldsMap = section.greyedFields.reduce((acc, greyedField) => {
+        const key = getFieldId(
+            greyedField.dataElement,
+            greyedField.categoryOptionCombo
+        )
+        acc[key] = true
+        return acc
+    }, {})
```
KaiVandivier, line 52: I think this could be simplified by using a `Set` object:
```js
const greyedFields = new Set(section.greyedFields.map(greyedField => getFieldId(greyedField.dataElement, greyedField.categoryOptionCombo)))
// ...
const isGreyedField = greyedFields.has(fieldId)
```

src/data-workspace/category-combo-table/category-combo-table.js
```diff
@@ -174,6 +184,7 @@ export const CategoryComboTable = ({
                                 <DataEntryField
                                     dataElement={de}
                                     categoryOptionCombo={coc}
+                                    disabled={isGreyedField(de, coc)}
                                 />
                             </DataEntryCell>
```
KaiVandivier, line 189: What do you think about just rendering a padding cell here if this is a greyed cell? It would save rendering a disabled input:
```js
isGreyedField(de, coc) ? <PaddingCell key={coc.id} /> : <DataEntryCell /* ...etc... */ />
```

src/data-workspace/get-field-id.js
HendrikThePendric: > However since all usages used/had access to de whole object I just kept it that way to keep it simple. But we can for sure just change it to strings.

Let's not implement support for strings until we need to.

HendrikThePendric (approved):

src/data-workspace/section-form/section.js
Birkbjo: Yeah, I just tend to be cautious with Map/Sets when sharing them as props, since they can easily cause update/render bugs. But that is mostly when using `setState` I guess, since here we are creating a new object each render anyways.

src/data-workspace/category-combo-table/category-combo-table.js
KaiVandivier: Okay! I think that's a funny design haha, but I will defer to how the old app works 👍

PR #67, opened 2022-06-08
feat: add print support

kabaros:
Add ability to print the currently displayed data set

- [x] Add react-to-print library to allow printing only certain sections of the app
- [x] Make printing accessible from the options menu in the top bar
- [x] Support printing `section` forms
- [x] Support printing `tabbed` forms
- [x] Support printing `standard`, and `custom` forms
- [x] Allow printing empty forms

image

### DIFF
--- a/src/app/app-wrapper.js
+++ b/src/app/app-wrapper.js
@@ -3,6 +3,7 @@ import React from 'react'
 import { QueryClientProvider } from 'react-query'
 import { HashRouter as Router, Route } from 'react-router-dom'
 import { QueryParamProvider } from 'use-query-params'
+import PrintAreaProvider from '../data-workspace/print-area/print-area-provider.js'
 import { RightHandPanelProvider } from '../right-hand-panel/index.js'
 import { CurrentItemProvider } from '../shared/index.js'
 import App from './app.js'
@@ -20,7 +21,9 @@ const AppWrapper = () => {
                     <QueryParamProvider ReactRouterRoute={Route}>
                         <CurrentItemProvider>
                             <RightHandPanelProvider>
-                                <App />
+                                <PrintAreaProvider>
+                                    <App />
+                                </PrintAreaProvider>
                             </RightHandPanelProvider>
                         </CurrentItemProvider>
                     </QueryParamProvider>
--- a/src/data-workspace/print-area/print-area-context.js
+++ b/src/data-workspace/print-area/print-area-context.js
@@ -1,0 +1,15 @@
+import { useContext, createContext } from 'react'
+
+const PrintContext = createContext()
+
+export default PrintContext
+
+export function usePrintableArea() {
+    const context = useContext(PrintContext)
+    if (context === undefined) {
+        throw new Error(
+            'usePrintableArea must be used within a PrintContextProvider'
+        )
+    }
+    return context
+}
--- a/src/data-workspace/print-area/print-area-provider.js
+++ b/src/data-workspace/print-area/print-area-provider.js
@@ -1,0 +1,39 @@
+import PropTypes from 'prop-types'
+import React, { useState, useEffect } from 'react'
+import PrintAreaContext from './print-area-context.js'
+
+export default function PrintAreaProvider({ children }) {
+    const [formPrinting, setPrinting] = useState(null)
+    const [formCleared, setFormCleared] = useState(null)
+
+    useEffect(() => {
+        if (formPrinting) {
+            window.print()
+        }
+        // cleanup function, which is ran before running the effect next time, so it makes sure the state is reset
+        return () => {
+            setPrinting(false)
+            setFormCleared(false)
+        }
+    }, [formPrinting, formCleared])
+
+    const printForm = (isEmptyForm = false) => {
+        setFormCleared(isEmptyForm)
+        setPrinting(true)
+    }
+
+    const value = {
+        formPrinting,
+        printForm,
+    }
+
+    return (
+        <PrintAreaContext.Provider value={value}>
+            <div className={formCleared ? 'form-cleared' : ''}>{children}</div>
+        </PrintAreaContext.Provider>
+    )
+}
+
+PrintAreaProvider.propTypes = {
+    children: PropTypes.node.isRequired,
+}
--- a/src/data-workspace/section-form/section.module.css
+++ b/src/data-workspace/section-form/section.module.css
@@ -5,6 +5,14 @@
     height: 100% !important;
 }
 
+@media print {
+    .labelWrapper {
+        background-color: transparent !important;
+    }
+    .title, .description {
+        color: black !important;
+    }
+}
 .headerCell {
     padding: 0 !important;
     height: auto !important;
--- a/src/app/app.css
+++ b/src/app/app.css
@@ -1,3 +1,40 @@
 body {
     background: var(--colors-grey100);
 }
+
+@media print {
+    /* This style is needed otherwise Chrome is printing only the visible part of a scrollable div, so we end up with one page of the form printed
+    https://stackoverflow.com/questions/25187715/css-print-media-query-prints-only-first-page */
+    * { overflow: visible !important; } 
+    div {
+        display: block !important; /* This is needed to get around issues with FireFox printing CSS grids: https://tosbourn.com/firefox-printing-issue-for-grid-css/ */
+        max-width: 100% !important;
+    }
+    .app-shell-adapter > header  {
+        display: none;
+    }
+
+    .hide-for-print {
+        display: none !important;
+    }
+    
+    table {
+        max-width: 100% !important;
+        page-break-after: always;
+    }
+    
+    table:last-child {
+        page-break-after: avoid;
+    }
+    
+    tr, td {
+        break-inside: avoid;
+        page-break-inside: avoid;
+    }
+
+    .form-cleared input,
+    .form-cleared textarea,
+    .form-cleared input+div svg /* we are adding an svg to style the checkbox which we need to hide as well in empty forms*/ {
+        visibility: hidden !important;
+    }
+}
--- a/src/app/layout/layout.module.css
+++ b/src/app/layout/layout.module.css
@@ -9,6 +9,16 @@
     transition: grid-template-columns 200ms ease-out;
 }
 
+@media print {
+    .wrapper :global(.controls) {
+        display: flex !important;
+        flex-direction: row;
+    }
+    .wrapper :global(.controls) :global(.clear-selections) {
+        display: none !important;
+    }
+}
+
 .wrapper.showSidebar {
     grid-template-columns: 1fr 380px;
 }
--- a/src/context-selection/context-selection/right-hand-side-content.js
+++ b/src/context-selection/context-selection/right-hand-side-content.js
@@ -1,10 +1,12 @@
+import classNames from 'classnames'
 import React from 'react'
 import { OptionsButton } from '../options/index.js'
 import styles from './right-hand-side-content.module.css'
 
 export default function RightHandSideContent() {
+    const containerStyles = classNames(styles.container, 'hide-for-print')
     return (
-        <div className={styles.container}>
+        <div className={containerStyles}>
             <OptionsButton />
         </div>
     )
--- a/src/context-selection/options/options.js
+++ b/src/context-selection/options/options.js
@@ -1,16 +1,29 @@
 import i18n from '@dhis2/d2-i18n'
 import { FlyoutMenu, MenuItem, MenuDivider, DropdownButton } from '@dhis2/ui'
 import React, { useState } from 'react'
+import { usePrintableArea } from '../../data-workspace/print-area/print-area-context.js'
 import { useRightHandPanelContext } from '../../right-hand-panel/index.js'
 
 export default function Options() {
     const rightHandPanel = useRightHandPanelContext()
+    const { printForm } = usePrintableArea()
+
     const [showMenu, setShowMenu] = useState(false)
 
+    const print = (isEmptyForm) => () => {
+        setShowMenu(false)
+        printForm(isEmptyForm)
+    }
     const optionsMenu = (
         <FlyoutMenu>
-            <MenuItem label={i18n.t('Print form with values')} disabled />
-            <MenuItem label={i18n.t('Print empty form')} disabled />
+            <MenuItem
+                onClick={print(false)}
+                label={i18n.t('Print form with values')}
+            />
+            <MenuItem
+                onClick={print(true)}
+                label={i18n.t('Print empty form')}
+            />
             <MenuDivider />
             <MenuItem
                 label={i18n.t('Help')}
--- a/src/data-workspace/custom-form/custom-form.module.css
+++ b/src/data-workspace/custom-form/custom-form.module.css
@@ -56,3 +56,13 @@ prevent accidentally applying this style to cells in the table body
     border-color: var(--colors-grey800);
     color: var(--colors-white);
 }
+
+@media print {
+    .customForm :global(table) > :global(tr:first-child) > :global(th),  
+    .customForm :global(table) > :global(tbody:first-child) > :global(tr:first-child) > :global(th),
+    .customForm :global(table) > :global(thead) > :global(tr) > :global(th) {
+        background-color: transparent !important;
+        border: 1px solid var(--colors-grey400);
+        color: black !important;
+    }
+}
--- a/src/data-workspace/data-workspace.js
+++ b/src/data-workspace/data-workspace.js
@@ -1,5 +1,6 @@
 import i18n from '@dhis2/d2-i18n'
 import { CenteredContent, CircularLoader, NoticeBox } from '@dhis2/ui'
+import classNames from 'classnames'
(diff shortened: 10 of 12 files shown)


src/app/app-wrapper.js
```diff
@@ -19,7 +20,9 @@ const AppWrapper = () => {
                     <QueryParamProvider ReactRouterRoute={Route}>
                         <CurrentItemProvider>
                             <RightHandPanelProvider>
-                                <App />
+                                <PrintAreaProvider>
```
kabaros, line 23: I added a new context to be able to communicate between the `options` menu where the print trigger is, and the datasets to print. The library requires a `ref` to the area to print, and we now set it in the context that is consumed and triggered from the options menu. Please let me know if there is a different approach, or if I could add the functionality to one of the already existing providers.

src/data-workspace/print-area/printable-area.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types'
+import React, { useEffect } from 'react'
+import { usePrintableArea } from './print-area-context.js'
+
+const PrintableArea = ({ children }) => {
```
kabaros, line 5: this `PrintableArea` wraps any components we would like to print. It's needed for two reasons:

- to set the area to print in the Context, so that it's accessible from a different part of the app tree (the options menu)
- to wrap the component in a DOM element and set a `ref` to it. The `react-to-print` requires a `ref` to the DOM element to print and it's not possible to set a ref on functional components.

HendrikThePendric: Solving it this way makes sense to me 👍

HendrikThePendric: Solving it this way makes sense to me 👍

src/app/app-wrapper.js
HendrikThePendric: I think that using a context provider is the right approach here. We've also been talking about perhaps adding global state to this app. If we did that, I expect that this part would actually not be included in the global state, since a) this provider's value is a ref and b) I don't think this provider is going to trigger a lot of re-renders. As such, I also think that keeping this is a separate provider is better than integrating it into an existing one.

src/data-workspace/section-form/section.module.css
```diff
@@ -5,6 +5,14 @@
     height: 100% !important;
 }
 
+@media print {
+    .labelWrapper {
```
kabaros, line 9: I added these because otherwise the titles of sections are very hard to read when printing, because the gradient background is not printed by default so the text is grey on white.

image

image

Mohammer5:
* "Synced" is still visible at the bottom of the last page
* In Firefox all the borders are missing

---

localhost:3000/#/?dataSetId=V8MHeZHIrcP&orgUnitId=ImspTQPwCqd&periodId=2020&sectionFilter=d24k6UcxpEA
* The sections have a "filter" field, that one can be hidden
* When printing a tabbed form, only the visible tab is printed (I haven't checked if that's also the case in the old app, if yes, then that's totally fine)
* Empty date and time inputs can be hidden, otherwise it's hard to write down the data (This is a good example for adding individual print styles per input component that I haven't thought about before)

---

See localhost:3000/#/?attributeOptionComboSelection=LFsZ8v5v7rq-C6nZpLKjEJr_yY2bQYqNt0o-i4Nbp8S2G6A&dataSetId=lyLU2wR22tC&orgUnitId=ImspTQPwCqd&periodId=202205
* Some rows are being cut and displayed on two pages
* The form doesn't fit on a page in portrait mode, landscape does work. How does the old app display this?
* The first page is blank in landscape mode (at least in Firefox)

---

See localhost:3000/#/?dataSetId=BfMAe6Itzgt&orgUnitId=ImspTQPwCqd&periodId=202205
* Some input fields are quite large
* The file upload buttin is being display, I think we could replace it with some text that says sth along the lines of "File upload not supported on printed forms" (@cooper-joe might have a better idea exactly what we could say here). Could be done by using css' `:before { content: '.......'; }` rule
* some some is cut off (not due to page break)

kabaros:
> * The sections have a "filter" field, that one can be hidden
> * When printing a tabbed form, only the visible tab is printed (I haven't checked if that's also the case in the old app, if yes, then that's totally fine)

for these two points, the old does print a single tab and it leaves the section filter field in the printed version so I didn't want to change that behaviour as we want to achieve parity.

I will address the other points!

kabaros:
@Mohammer5 I addressed most of the issues in your comments (see my followup comments below), and also implemented the empty form printing. Please have a look and let me know. 

> "Synced" is still visible at the bottom of the last page

removed

> In Firefox all the borders are missing

we agreed this seems to work fine .. probably something with your settings (maybe worth having it tested by someone else though).

localhost:3000/#/?dataSetId=V8MHeZHIrcP&orgUnitId=ImspTQPwCqd&periodId=2020&sectionFilter=d24k6UcxpEA

> The sections have a "filter" field, that one can be hidden
> When printing a tabbed form, only the visible tab is printed (I haven't checked if that's also the case in the old app, if yes, then that's totally fine)

this is same behaviour as old app

> Empty date and time inputs can be hidden, otherwise it's hard to write down the data (This is a good example for adding individual print styles per input component that I haven't thought about before)

I cleared it when the user is printing an empty form (similar to the old app) but left it as it is when it's normal print - does that make sense?

> See localhost:3000/#/?attributeOptionComboSelection=LFsZ8v5v7rq-C6nZpLKjEJr_yY2bQYqNt0o-i4Nbp8S2G6A&dataSetId=lyLU2wR22tC&orgUnitId=ImspTQPwCqd&periodId=202205
> Some rows are being cut and displayed on two pages
> The form doesn't fit on a page in portrait mode, landscape does work. How does the old app display this?
> The first page is blank in landscape mode (at least in Firefox)
> 

These are all issues related to a quirk of printing a scrollable div. I had a fix for chrome but that broke things in Firefox. I think now I have a fix that works for both .. please check and let me know

> See localhost:3000/#/?dataSetId=BfMAe6Itzgt&orgUnitId=ImspTQPwCqd&periodId=202205
> Some input fields are quite large

I made the table be 100% width in print .. I feel that looks better and gives people more space when writing. I could revert it but it looks good imo .. maybe check it again now that Firefox issues were fixed.

> The file upload buttin is being display, I think we could replace it with some text that says sth along the lines of "File upload not supported on printed forms" (@cooper-joe might have a better idea exactly what we could say here). Could be done by using css' :before { content: '.......'; } rule

I haven't done this one .. I think this might be an overkill, we could just hide it but replacing it with text might require more than just CSS styles, or do you have a workaround in mind?

src/app/layout/layout.js
```diff
@@ -11,10 +11,14 @@ const Layout = ({ header, main, sidebar, footer, showSidebar, showFooter }) => {
 
     return (
         <div className={wrapperClass}>
-            <header className={css.header}>{header}</header>
+            <header className={`${css.header} hide-for-print`}>{header}</header>
```
Mohammer5, line 14: You can use the `classnames` library in this case.
[Here](https://github.com/dhis2/data-entry-app/blob/TECH-859/print-forms/src/app/layout/layout.js)'s a good example.

kabaros: > You can use the classnames library when applying multiple class names.

updated

src/data-workspace/print-area/print-area-context.js
```diff
@@ -0,0 +1,13 @@
...
+const PrintContext = createContext({
+    formPrinting: null,
+    printForm: () => {},
+})
+
+export default PrintContext
+
+export function usePrintableArea() {
+    const { formPrinting, printForm } = useContext(PrintContext)
+    return { formPrinting, printForm }
```
Mohammer5, line 12:
```suggestion
    return useContext(PrintContext)
```

Mohammer5 (commented):
Code looks good to me, left a few minor things / suggestions.
I'm still trying to access the app through chrome without any success so far.. I'll approve once I'm able to (I'll also ask the team if someone else could review this in the browser, I just can't get the proxy to work with chrome/chromium/brave)

src/data-workspace/print-area/print-area-provider.js
```diff
@@ -0,0 +1,40 @@
...
+import PrintAreaContext from './print-area-context.js'
+
+export default function PrintAreaProvider({ children }) {
+    const [isPrinting, setPrinting] = useState(null)
+    const [isFormCleared, setFormCleared] = useState(null)
+
+    useEffect(() => {
+        if (isPrinting) {
+            window.print()
+        }
```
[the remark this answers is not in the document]
kabaros: I added a comment - the component is not unmounted, but for effects, the cleanup function is ran on every render so that ensures it's "cleaned" before next run.

image

src/data-workspace/print-area/print-area-context.js
```diff
@@ -0,0 +1,13 @@
+import { useContext, createContext } from 'react'
+
+const PrintContext = createContext({
+    formPrinting: null,
+    printForm: () => {},
```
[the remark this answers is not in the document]
kabaros: updated the function to throw an error .. I thought I am using Kent Dodds technique 😅, as in using a custom consumer hook,  but maybe I am missing something?

src/data-workspace/print-area/print-area-provider.js
```diff
@@ -0,0 +1,40 @@
...
+    }, [isPrinting, isFormCleared])
+
+    const printForm = (isEmptyForm = false) => {
+        setFormCleared(isEmptyForm)
+        setPrinting(true)
+    }
+
+    const value = {
+        formPrinting: isPrinting,
```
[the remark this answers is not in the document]
kabaros: you're right .. updated

src/data-workspace/print-area/print-area-context.js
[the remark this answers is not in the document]
kabaros: I added a null check to the custom consumer hook as well now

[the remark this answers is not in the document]
Mohammer5: I was referring to throwing an error if the context is undefined.
The current approach of the app is similar but duplicates the shape of the state, which is less ideal I'd say

src/data-workspace/print-area/print-area-provider.js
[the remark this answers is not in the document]
Mohammer5: I'm not entirely sure why this works :sweat_smile: but it seems to work

src/data-workspace/print-area/print-area-context.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default PrintContext
+
+export function usePrintableArea() {
+    const context = useContext(PrintContext)
+    if (context === undefined) {
+        throw new Error(
+            'usePrintableArea must be used within a PrintContextProvider'
+        )
+    }
```
Mohammer5, line 18: The context will never be undefined as we pass a default value to `createContext`

Mohammer5 (approved):
I think it's good! Just checked it with chromium, too.
Some forms have quite small tables but there's a page break after each table, which could result in excess paper usage. But I don't think that we'll have to worry about that now.

Could you create a Jira issue and assign it to [TECH-1187](https://dhis2.atlassian.net/browse/TECH-1187)?

HendrikThePendric (commented):
The code is looking good and most of the print styles are good too. I found a few minor issues, which I will describe with screenshots below:

---

Below are two custom forms. I think we should let the font color and the border color match the regular forms.
Screenshot 2022-06-22 at 14 04 05
Screenshot 2022-06-22 at 14 07 25

---

Below is a default form, I think we need to hide the filter field.
_I think @Mohammer5 already mentioned this and perhaps you've fixed this already too. My availability is very patchy due to the conference, so I actually took these screenshots hours ago._
Screenshot 2022-06-22 at 14 06 12

---

Below we are missing the right border on the table. I'm not actually 100% sure this is a bug in the print-styles, or in the custom-forms implementation. This appears to be a custom form which is also using row total fields. This still needs to be added to the custom-forms implementation. However, I don't think that could explain the missing right border on the table header. So I think some more investigation is necessary here.

Screenshot 2022-06-22 at 15 40 18

---

Then lastly, there is this page wrapping issue which I'm not sure we can/should try to fix. It feels like one of those things that can be solved for a particular set of conditions, but that fix might actually break the behaviour for other conditions.... The screenshots below are produced under the following circumstances:
- Dataset *ART Monthly Summary*, which is a custom form, which consists of two fairly large tables. The first one doesn't fit on 1 page and this seems to be causing issues.
- Selected Printer: my actual physical printer, when selecting "Save as PDF" there still is a problem but this is slightly different: the first table now fits onto the page and is directly followed by the second table on page 2 and 3. However, we now have a 4th completely blank page.

I'll provide some additional clarification to the following images....

a) This is the page transition between page 1 and 2. You see the table extends just past the page end, and is then followed by a virtually empty page.
Screenshot 2022-06-22 at 15 45 59

b) Then this is the page transition between page 2 and 3. Page 2 is nearly completely empty and the second table is on page 3.
Screenshot 2022-06-22 at 15 46 51

kabaros:
> Below is a default form, I think we need to hide the filter field.
> I think @Mohammer5 already mentioned this and perhaps you've fixed this already too. My availability is very patchy due to the conference, so I actually took these screenshots hours ago.

@HendrikThePendric for this one .. I left the filter as this is the behaviour of the old app. I feel it should be removed too though, so happy to do so if we agree on it

> Then lastly, there is this page wrapping issue which I'm not sure we can/should try to fix. It feels like one of those things that can be solved for a particular set of conditions, but that fix might actually break the behaviour for other conditions.... 

I added a page break after each table for the print version but we can get rid of that. I am not sure if there is a way to add a page break based on how long is the table.

kabaros:
@Mohammer5  @HendrikThePendric I rebased `development` and I see the structure of the footer has changed significantly. I am still hiding the footer in the print view, is that still the behaviour we want? i.e. do we only want to hide "synced" message in print view but also the complete/incomplete bottom bar
 
image

the old app keeps the bottom bar, but it also shows a random hidden div at the bottom which seems like a bug .. so I am not sure I can use it as a reference
image

Mohammer5:
@kabaros one thing I just realized is: Maybe it makes sense to add the current content selection to the top of the form so people know what for in the app the values belong to when copying values from printed forms over to the app.

---

> > Below is a default form, I think we need to hide the filter field.
> > I think @Mohammer5 already mentioned this and perhaps you've fixed this already too. My availability is very patchy due to the conference, so I actually took these screenshots hours ago.
> 
> @HendrikThePendric for this one .. I left the filter as this is the behaviour of the old app. I feel it should be removed too though, so happy to do so if we agree on it

Yes, please remove it, doesn't make much sense to include it :rofl:

HendrikThePendric (approved):
> I left the filter as this is the behaviour of the old app. I feel it should be removed too though, so happy to do so if we agree on it

Yeah, I also discussed briefly with Jan-Gerke about this. Let's remove these filters.

> I added a page break after each table for the print version but we can get rid of that. I am not sure if there is a way to add a page break based on how long is the table.

Thanks for clarifying. I think a page break is probably nice to add after a table in most cases, so we should probably keep using it and accept that there will be some cases (i.e. the case I found, where one single table extends just across page), where it doesn't look good. I noticed you have also added `table:last-child { page-break-after: avoid; }` which is great. So let's just keep the current behaviour.

> > Below are two custom forms. I think we should let the font color and the border color match the regular forms.
> 
> implemented this [here](https://github.com/dhis2/data-entry-app/pull/67/files#diff-509adcfc4ac56dbc8383b87ebb8bb2fe2e5a218e817603463a720cb4b7c94f4aR60) - please have a look if it looks good.
I ran the app locally and opened a custom form. Looks great. Thanks.

> > Below we are missing the right border on the table. I'm not actually 100% sure this is a bug in the print-styles, or in the custom-forms implementation.
> 
> I am unable to reproduce it. The border prints fine for me (both Chrome and FF) but I think the underlying cause is that the table has a fixed inline `width: 1100px`, I put a `max-width: 100% !important` on the tables print style which should overwrite it, but please have a look as I can't reproduce it.

OK, let's forget about this point, at least for now. I could only reproduce this broken behaviour on a single custom forms which also has some other minor issues. On most custom forms this doesn't happen. If the cause is as you suspect, then I don't think we need to do anything to fix this (it would be the responsibility of who built the form in the first place). If it's to do with the lack of support for column totals in custom forms, as I suspect, then it is out-of-scope for the current PR.

> @kabaros I don't think we need to include the footer, it doesn't add any value to the printed form.

💯 agreed!

> @kabaros one thing I just realized is: Maybe it makes sense to add the current content selection to the top of the form so people know what for in the app the values belong to when copying values from printed forms over to the app. You can ask @cooper-joe what that should look like (maybe displaying the url is enough as it contains all the information already; then we don't need to add anything as that's a built-in feature when printing)

I agree this could be useful. If it's a relatively low-effort thing to add, please go ahead and do so. However, if it requires quite a lot of work, we should probably see what the old app does and make sure we offer similar functionality. I agree with Jan-Gerke that showing the address bar would be a nice seemingly low-effort solution that would provide the required functionality.

---
I'm approving this PR already, but please:
- remove/hide the filter fields when printing
- see what we can do to show the context selection

HendrikThePendric (approved):

PR #68, opened 2022-06-13
refactor(data values): use new endpoint

Mohammer5:
Closes TECH-1206 ([link](https://dhis2.atlassian.net/browse/TECH-1206))

PR includes to tasks, separated into two commits:
* [refactor(use data value mutation): split into separate hooks](https://github.com/dhis2/data-entry-app/pull/68/commits/40c7f1c70e1f191277008961cc1ae3a6a3ec959f)
* [refactor(data workspace): adapt to new data values endpoint](https://github.com/dhis2/data-entry-app/pull/68/commits/77520c1f5a4d5fe032dcd6aba2a43c6bc62acc6f)


HendrikThePendric (approved):
Looks good. Left one minor comment.

Birkbjo:
As said on slack, I think this refactors way more than is necessary. The `use-data-value-mutation.js` might've warranted some code-splitting but to me this is quite extreme. I also generally like that code like that is together, and everything related to `dataValue`-mutation was encapsulated in that file, of 170 lines. 
Now it's split into so many files, and I frankly find it much harder to get an overview over the mutation code now.

Again, some code-splitting could've been done here - like maybe adding `addDataValue`, `deleteDataValue` and `updateDataValue` to a `utility-file`, but having separate files for these simple function is so overkill and bloats the file-tree imo. I've said it before, and I will again - I really don't understand the "one-file-per-function"-thing. To me it's way more beneficial to have related code like this in one file, then spread them out to 3 files of 4 lines each. 

I've said this before as well - I really think we should be careful about refactors like this, especially when other people have written the code. It's very easy to introduce bugs to code your not 100% familiar with, and frankly the code worked before.  To me it's just a waste of resources that could've been done on something more productive. We cannot refactor code everytime we work on something, then we won't do anything else than refactor. Everyone gets the urge to refactor something all the time, but I think it's something we generally something we should resist, unless there's a good reason to do it. 

I actually had a branch to test the new API a few weeks back, and all that was really needed was to update the API-url and query-params.

Mohammer5:
Sorry for the wall of text in this comment. I'd like to talk about different perspectives how we can look at this issue right now, explain why I did things the way I did and also comment on your remarks.

Before we get into the weeds, I want to talk about the reason why I was working on this in the first place. The new endpoint's response contains the min-max limits, so this actually a sub-task of [TECH-1199](https://dhis2.atlassian.net/browse/TECH-1199) which adds functionality for adding/editing min-max values and comments.

In my opinion, there are two different topics which we need to talk about and the first one I want to comment on is the regression topic:

### Newly introduced regression

> It's very easy to introduce bugs to code your not 100% familiar with, and frankly the code worked before.

In this regard, I'm a bit more on the defensive side. The regression itself has nothing to do with the refactor, it's due to the fact that I simply copied the example api parameters without thinking about whether the endpoint itself is functionally complete or has bugs, I blindly trusted that it'd simply work and these things tend to happen. I'm certain the same regression would've happened had I not refactored anything at all.

This could've been a bit more obvious if I had split 40c7f1c into two commits (refactoring & changing the endpoint), but nonetheless I don't think the refactoring is what caused the regression and therefore would like to keep these to topic separate. (I'm not saying we shouldn't talk about it, but I just don't want to conflate it with the refactoring topic)

### Refactoring

When it comes to the refactoring, I think there are different perspective I would like to touch on.

#### The team play perspective

> I really think we should be careful about refactors like this, especially when other people have written the code.

We had a similar discussion in the past about the same topic. What I proposed back then was that we should communicate these things bit better before we actually commit to them, e.g. by raising awareness of what I wanted to do during our Tuesday meeting. In this particular case, I simply didn't think about doing so, and it's something I have to reflect on. Raising this during the Tuesday meeting could be the wrong approach. I suppose I should've tagged the person who wrote the code I refactored in the first place and get his opinion and feedback on the changes rather than rushing this.

#### The time perspective

> I appreciate the urge to iterate quickly on this, and I'm sorry I wasn't able to look at this earlier.

From a time perspective, one could definitely argue that we're approaching a deadline that we don't want to miss, so in order to not put stokes in our plan's wheels. Especially with having in mind that we should refine the code base post-release, I understand the concern that refactoring is costing us valuable time.

#### The code perspective - Part 1

> The use-data-value-mutation.js might've warranted some code-splitting but to me this is quite extreme. I also generally like that code like that is together, and everything related to dataValue-mutation was encapsulated in that file, of 170 lines. Now it's split into so many files, and I frankly find it much harder to get an overview over the mutation code now. Again, some code-splitting could've been done here - like maybe adding addDataValue, deleteDataValue and updateDataValue to a shared "utility-file", but having separate files for these simple function is so overkill and bloats the file-tree imo.

Of course, the way I split the code and the amount of files I created could be done differently, I won't argue against that and I don't mind changing things here at all. These are details where we can find a middle ground. To be honest, I quite like the idea of putting the three utility functions in one file (we could even use lodash's `set` & `pullAt` utilities to complete get rid of these helpers).

#### The code perspective - Part 2

The way I interpret your feedback, you're mostly commenting on the code perspective (disregarding the regression part for reasons I highlighted above). So I'll talk about conceptual changes I introduced. I'd like to highlight again that this is not a justification for doing this now or merging the PR right away; since your comments regarding the refactoring are mostly about this code perspective, I'd like to respond and explain my train of thought while working on this PR.

*Disclaimer: I think in our team we sometimes have a completely different conceptual approach to writing code, and neither of those are better than the others, they're just different. I'll spend a bit more time highlighting my approach simply because I think it's quite difficult to truly understand the approach of another person (which applies to me as well of course). So what I can do at least is to try to explain my approach as good as possible, so you can retrace my train of thought. I'll try to explain what I was thinking at the time, which is not meant as a justification for doing that now*

The way I look at the code changes is: The entire mutating functionality was located in

* [`src/data-workspace/data-entry-cell/use-data-value-mutation.js`](https://github.com/dhis2/data-entry-app/blob/dc62d661f485bc0059e88df3f334e47d72f5f86c/src/data-workspace/data-entry-cell/use-data-value-mutation.js)

In terms of refactoring, this is the only place I really touched (of course I had to adjust the important hook names in the input components after having worked on this). And when trying to understand the codebase in its entirety, at some point, you'll end up getting all the way down to the inputs. At this point the original `useDataValueMutation` was comparatively difficult to understand whereas now, it's quite obvious because you don't really have to deal with implementation details of logic branching.

When trying to understand the original `use-data-value-mutation.js` module (it was the first time I had to look at this particular part of the code base), I had to wrap my mind around how this the hook is supposed to be used. I figured out that there are two config objects (`MUTATION_TYPES` and `mutationsByType`) of which the former had to be exported as well and used together with `useDataValueMutation` (unless when using the default value of the hook's argument). When trying to understand what's happening in the `onMutate` callback, I had to understand what's happening with an imperative perspective, understand the different logical branches the code can get into, as it's covering 3 cases now (and soon 4 cases as comment editing has to be supported as well (see [here](https://github.com/dhis2/data-entry-app/blob/TECH-1199/src/data-workspace/use-data-value-mutation/optimistically-set-data-value-comment.js))).

In my opinion, the pattern that was chosen is odd, as it feels like a rather ad hoc solution to get things working. The original implementation tries to implement the functionality itself, taking various differences into account and then the inputs are built around it (in the sense that they now have to use that functionality, the way it was designed). 
What I tried to do is to be more declarative. There are essentially three (soon
four) different things we want to do:

1. Set a data value's `value` (either add a new one or update an existing one)
2. Upload a file
3. Delete a file

So instead of having compact functionality with a minimum of lines of code, I tried to improve being able to reason about the code. For every type of mutation there's now a hook that you can take a look at specifically when trying to understand the code.

When I write code, the mindset I try to embrace is to try to make the code as simple as possible as well as to make it maintainable and adaptable without having to introduce exceptions. Of course, the term "simplicity" can be interpreted in different way. One way is to say you want to reduce the lines of code so you can minimize the work when changing things. Another way to look at simplicity is to ease the path of understanding the code when looking at the code, this is especially true for developers new to the code base or us in the future when we don't have the details in memory anymore.

For a rather long time I've been a very strong "DRY code" proponent and in the last couple of years I have changed a bit on that because I started seeing issues with that:

##### Ad hoc code creep

One issue that I've seen time and time again is that developers (including myself) are adding code and consciously defer the decision to refactor things while still working on the implementation. While this is not an issue on a pull request / feature basis, it tends to introduce a lot of inconsistencies into the codebase, which then in turn subconsciously promotes writing less clean, well written code, as the existing code already is not that consistent itself.

Of course, we have to strike a balance between trying to write perfect code (which is impossible) and getting things done, especially when there are deadlines. But it shouldn't be overlooked at all and always be kept in the back of our heads in my opinion.

##### Comprehensibility of the code

As I mentioned earlier, I think that the conceptual changes I introduced benefit the comprehensibility - regardless of whether everything is in one file or in many. The path from the root (src/app/app.js) all the way down to how an individual input handles data value changes is more straight forward.

##### DRY code

@birk you've mentioned on slack that some file share a lot of code and the diff is actually quite small (I think you specifically mentioned the `optimistically-update-*.js` files).

I agree that these files violate the DRY code approach. As mentioned above, I was a very strong proponent of DRY code myself for a long time. What I have observered over time though is that it happens - at some point or another - that we'll have to add exceptions for one particular case but not for others. This tends to complicate DRY code and impairs the comprehensibility of the code base. Over time this causes the codebase to be quite complicated, despite it being relatively compact.

When I started to write code that has distinct function for different functionality (e.g. setting the value, uploading a file and deleting a file), whenever only a particular functionality has to be changed, the other functions remain untouched, which increases the maintainability of the code base.

You could argue on the other hand that we want the code to be as adaptive as possible and that this more important than the comprehensibility, but in my experience, we these parts of the code base have to be changed rather rarely. But it's quite normal for a new developer or future-you/future-me to look at this code.

My conclusion is that comprehensible code with some repetition saves more time in the end, than DRY code, which doesn't mean that you can't compose the functionality with shared utilities.

### Conclusion

That's a lot of text that I just wrote, I'm sorry in advance, I just feel the urge to explain my motivation and train of thought as detailed as properly so any team member can follow what I was thinking.

I don't mind using fewer files, if that's the issue, I'm more than just willing to change that! I also agree that the timing of this refactoring was just quite unfortunate, and I should've picked both @kai and @ismay as reviewers to get some input on this. I think I learned my lesson here and will try to do this better in the future. You can make a case for why we could've deferred this to post-release and then also would've had more time to talk about this in detail.

When it comes to the refactoring, disregarding the "how many files" part, what do you think about what I was trying to explain? Did I manage to express my motivation well enough? And if yes, what are your thoughts on this?

PR #69, opened 2022-06-13
fix: try to prefetch optionSets

Birkbjo:


src/context-selection/org-unit-selector-bar-item/use-org-unit-paths-by-name.js
```diff
@@ -11,19 +11,14 @@ export default function useOrgUnitPathsByName(searchTerm) {
...
-    } = useQuery(queryKey, {
+    const { isIdle, isFetching, error, data } = useQuery(queryKey, {
         enabled: !!searchTerm,
         select: (data) => data.organisationUnits.map(({ path }) => path),
     })
 
     return {
         called: !isIdle,
-        loading,
+        loading: isFetching,
```
Birkbjo, line 21: "isLoading" behaviour changed in `react-query@4`. Now `isLoading` is `true` if the query is `disabled`.

src/context-selection/org-unit-selector-bar-item/get-offline-levels-to-load.js
```diff
@@ -15,7 +15,7 @@ export default function getOfflineLevelsToLoad({
 }) {
     return userOrganisationUnits
         .map((userOrgUnit) => {
-            const foundLevel = organisationUnitLevels.find(
+            const foundLevel = organisationUnitLevels?.find(
```
Birkbjo, line 18: This crashed for some reason for me, so I just added this as a saveguard. Not sure if necessary though.

src/app/app-wrapper.js
```diff
@@ -1,21 +1,43 @@
...
 import { QueryParamProvider } from 'use-query-params'
 import { RightHandPanelProvider } from '../right-hand-panel/index.js'
 import { CurrentItemProvider } from '../shared/index.js'
 import App from './app.js'
+import createIDBPersister from './query-client/persister.js'
 import useQueryClient from './query-client/use-query-client.js'
 
+const persister = createIDBPersister()
+
+const persistOptions = {
```
Birkbjo, line 15: Should we export this from `persistor.js`, and import here instead of specifying the options here?

package.json
```diff
@@ -57,14 +57,13 @@
         "re-reselect": "^4.0.0",
         "react-chartjs-2": "^4.1.0",
         "react-final-form": "^6.5.7",
-        "react-query": "4.0.0-alpha.10",
+        "react-query": "^4.0.0-beta.23",
```
Birkbjo, line 60: Note that this means that we now have both `react-query@3` and `react-query@4` in the app. V3 is used by `useDataQuery`, which should only be used by the `OrganisationUnitTree`.

src/app/app-wrapper.js
Mohammer5: There's an issue in regards to the hook you've touched: [TECH-1124](https://dhis2.atlassian.net/browse/TECH-1124). It's scheduled to be worked on post-release, but now that you've touched this area, I think this could be a good time. I really don't like the name `useQueryClient` because it's conflicting with the hook with the same name exported by `react-query`, which is why I was suggesting a different name.

I can see why you've moved the persister stuff from the hook to the `app-wrapper.js`. To me it'd make sense to keep the react-query related code in a separate file entirely, e.g. in a `<ConfiguedQueryClientProvider />` component file that internally creates the client instance and configures it, regardless of the implementation detail (with a config passed to the provider or through a hook).

src/data-workspace/inputs/option-set.js
```diff
@@ -42,10 +41,19 @@ export const OptionSet = ({
         }
     }
 
-    const optionSet = selectors.getOptionSetById(metadata, optionSetId)
+    const optionSet = useOptionSet(optionSetId)
+
+    if (!optionSet) {
+        console.warn('Could not find optionSet with id:', optionSetId)
```
Mohammer5, line 47: I'm wondering: Do we have an official policy whether we want to show console output in production code?

src/metadata/option-set-query-key-factory.js
```diff
@@ -0,0 +1,44 @@
+const optionSetFields = 'id,version,options[id,displayFormName]'
+
+const queryKeyFactory = {
```
Mohammer5, line 3: The way I understand the current architecture for query key factories is: There should be a `query-key-factory.js` in the root of the module (this file is already at the root of metadata, so the location is fine, but the name is not). And that file can export multiple things, like the [`data-workspace/query-key-factory.js`](https://github.com/dhis2/data-entry-app/blob/development/src/data-workspace/query-key-factory.js).

Not sure whether that's the best approach, but it's the current "standard" which @ismay landed at after reading through the `react-query`'s author's blog posts.

src/metadata/use-option-sets.js
```diff
@@ -0,0 +1,103 @@
+import { useQueryClient, useQuery } from 'react-query'
+import keys from './option-set-query-key-factory.js'
```
Mohammer5, line 2: Just a suggestion: I think the identifier `keys` is a bit ambiguous and when reading this module without having a complete understanding of the code already, the term `keys` take a small amount of time to be interpreted (e.g. "key for what?"; after some reading: "ah.. key for react-query"). Maybe `queryKeys` is better suited, or after renaming the file above this one to `query-key-factory`, this would turn out as: `optionSetQueryKeyFactory.keys` (or so..) which make this very explicit.

Mohammer5 (changes requested):

src/metadata/use-option-sets.js
Birkbjo: Yeah I actually saw this when going over the PR as well, and agree. Will rename to `queryKeys`

src/metadata/use-option-sets.js
```diff
@@ -0,0 +1,103 @@
...
+// refetching is done by comparing versions
+export const useAllOptionSets = () => {
+    return useQuery(keys.all, {
+        meta: {
+            persist: true,
+        },
+        staleTime: Infinity,
+        cacheTime: Infinity,
+        notifyOnChangeProps: ['data', 'error'],
+        select: (data) => {
```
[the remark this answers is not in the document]
Birkbjo: True, thanks for the reminder

src/metadata/option-set-query-key-factory.js
Birkbjo: Are you referring to the prefix `option-set`? If so I can rename it, It was just more specific... 
Not sure what else you are referring to? This does not need to return "multiple" things. It's all just the same API-endpoint, and thus the only object needed. I feel like the structure of the actual factory complies to the structure of the other factories? Just that this only operates on the "optionSet"-endpoint, and thus all methods are related?

I guess you could argue that the `allVersion`-one is a "separate"-entity, and thus could be it's own object. But I decided to keep it simple, and "group" them by endpoint. If that is what you are referring to, we can change it. But if the structure is somehow otherwise different than the other factories or the link provided by @ismay please explain.

src/app/app-wrapper.js
Birkbjo: Agreed,  I will move this setup to a `ConfiguedQueryClientProvider`-component.

I'm not sure I necessarily agree that the naming is that bad, though. And I wouldn't say that it's a "completely different thing". The hook sets up a `react-query` `queryClient`, and thus can be used as an "override" for `useQueryClient`, so I actually think the naming is fair. But yeah, it's conflicting and you could probably be more specific like `useConfiguredQueryClient`.

HendrikThePendric: I agree it would be good to rename "our" `useQueryClient` to `useConfiguredQueryClient` in the long run. We do import `useQueryClient` from 'react-query' in a few places as well, so it can be slightly confusing to use the same variable name for two distinct entities. But this can remain a task for after the soft freeze IMO.

Birkbjo: Let's handle the `useQueryClient`-thing in a separate PR.

src/metadata/option-set-query-key-factory.js
Birkbjo: btw. I think I followed the style in `query-key-factory` in custom-forms. Which exports the `queryKeyFactory` as a default export.

src/metadata/use-option-sets.js
```diff
@@ -0,0 +1,103 @@
...
+        },
+    })
+}
+
+export const useOptionSet = (optionSetId) => {
+    const { data: optionSets } = useAllOptionSets()
+
+    const optionSet = optionSets?.[optionSetId]
+    return optionSet
+}
```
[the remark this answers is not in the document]
Birkbjo: Maybe we should handle the loading in `LoadApp`, and catch the loading from the `usePrefetch`-hook? That would simplify the use of this hook deeper down the tree.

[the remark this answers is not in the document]
Mohammer5: We discussed this this morning and decided against doing it in the load app:
* Increases the time before app is usable, especially on slow connections
* Data in not necessarily needed by the forms the user wants to edit

Birkbjo:
Closed since this is now part of metadata-response.

PR #73, opened 2022-06-17
feat(right hand panel): add min-max limits and comment functionality

Mohammer5:
Subtask of [TECH-856](https://jira.dhis2.org/browse/TECH-856)
Closes [TECH-1199](https://jira.dhis2.org/browse/TECH-1199)
Closes [TECH-1221](https://jira.dhis2.org/browse/TECH-1221)
Closes [TECH-1225](https://jira.dhis2.org/browse/TECH-1225)
Closes [TECH-1227](https://jira.dhis2.org/browse/TECH-1227)
Relates to [TECH-1206](https://jira.dhis2.org/browse/TECH-1206) / #68 

This PR finished the right hand panel topic and completes the functionality:

* [x] Adding/changing/deleting min-max values
* [x] Adding/changing/removing comments
* [x] Button in the footer that opens the data details sidebar
  * Handled by #74

Bugs still present in the PR 
* [x] Issue with setting the limit, both api- and UI-wise
   * [x] Issue with api: https://dhis2.slack.com/archives/C02NHB92S2H/p1656662944758069 
     * [x] Make both values required when the other one is present
   * [x] Issue with code: For some reason the update request remains "pending", the `onMutate` callback won't be called at all (do you have an idea @ismay? Maybe we can take a look at this together on Monday)

Tasks
* [x] Show data details for active field at all times
* [x] When focussing new field, keep details + comment + min-max field open because they are in the payload already.
* [x] Audit log and history need to be loaded when expanding those sections, not when opening the sidebar / changing the item.
* [x] Audit log and history sections need to collapse when changing the detail item to reduce network bandwidth
* [x] Shift + Enter => Ctrl / Cmd + Enter
* [x] [TECH-1221](https://dhis2.atlassian.net/browse/TECH-1221)
* [x] [TECH-1225](https://dhis2.atlassian.net/browse/TECH-1225)
* [x] [TECH-1227](https://dhis2.atlassian.net/browse/TECH-1227)

Tasks that will be done in a different PR
* [ ] Audit log, consider tabular view, consider changing name to change log ([TECH-1281](https://dhis2.atlassian.net/browse/TECH-1281))
* [ ] Implement marking for follow-up ([TECH-1280](https://dhis2.atlassian.net/browse/TECH-1280))

---

#18 mentions a few more tasks which I'd like to defer and address after merging this issue:
* [ ] Cypress tests
* [x] Toggle sidebar
   * [ ] when user double clicks on cell -> Skip for now.  
     ask @cooper-joe: Double click normally selects the text in an input, do we really want to disable that behavior?


src/data-workspace/min-max-limits-mutations/use-delete-limits.js
```diff
@@ -0,0 +1,108 @@
...
+
+        // Used so the limits UI can switch from form to value display
+        onSuccess: onDone,
+
+        onMutate: async (variables) => {
+            // Cancel any outgoing refetches (so they don't overwrite our optimistic delete)
+            await queryClient.cancelQueries(dataValueSetQueryKey)
+
+            // Snapshot the previous value
+            const previousDataValueSet = queryClient.getQueryData(dataValueSetQueryKey)
```
Birkbjo, line 79: I don't think this is needed. You will get the `oldValue` in the callback from `setQueryData`. See next comment.

src/data-workspace/use-data-value-mutation/optimistically-set-data-value-comment.js
```diff
@@ -0,0 +1,54 @@
...
+                categoryOptionCombo: newDataValue.co,
+                dataElement: newDataValue.de,
+                orgUnit: newDataValue.ou,
+                period: newDataValue.pe,
+            }
+
+            return addDataValue(previousDataValueSet, formattedNewDataValue)
+        } else {
+            const { comment } = newDataValue
+            const formattedNewDataValue = {
```
Birkbjo, line 40: I don't really understand why we can't consolidate all these `optimisticallyUpdate-functions`. Just transform the "new-value" before you send it to a shared `optimisticalyUpdateDataValue`-function.

src/data-workspace/use-data-value-mutation/use-set-data-value-comment-mutation.js
```diff
@@ -0,0 +1,38 @@
...
+    const { attributeCombo, attributeOptions } = useApiAttributeParams()
+    const mutationFn = useMutationFn(SET_DATA_VALUE_COMMENT_MUTATION)
+
+    return useDataValueMutation({
+        mutationFn,
+        onSuccess: onDone,
+        onMutate: ({ queryClient, newDataValue, dataValueSetQueryKey }) =>
+            optimisticallySetDataValueComment({
+                queryClient,
+                newDataValue,
```
Birkbjo, line 32: Regarding comment above. You could format the value of `newDataValue` here. Could even create different functions to handle the transformation, and then use a shared `optimisticallySetDataValue`-function.

src/data-workspace/min-max-limits-mutations/use-delete-limits.js
```diff
@@ -0,0 +1,108 @@
...
+
+            // Optimistically delete to the new value
+            queryClient.setQueryData(dataValueSetQueryKey, () => {
```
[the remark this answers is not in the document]
Mohammer5: We need that value outside of that function though, see [here](https://github.com/dhis2/data-entry-app/blob/72dfc32dd30eb7e6b4e034f3eb9381a6d4c73de8/src/data-workspace/min-max-limits-mutations/use-delete-limits.js#L94):

```js
const context = { previousDataValueSet, dataValueSetQueryKey }
return context
```

src/data-workspace/min-max-limits-mutations/use-delete-limits.js
Mohammer5: I'll resolve this one, we can talk about it in the next comment

src/data-workspace/min-max-limits-mutations/use-delete-limits.js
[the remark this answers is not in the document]
Birkbjo: Ah, right. Nvm then :)

src/data-workspace/use-data-value-mutation/optimistically-set-data-value-comment.js
Birkbjo: I mean, as far as I can tell the only difference between the functions was the actual updated value. I might have missed some other differences, but as far as I can see, in the end these functions are actually the same mutation - the value is just different.

Birkbjo: I will come back to this tomorrow, currently only have access to phone.

src/data-workspace/min-max-limits-mutations/use-delete-limits.js
```diff
@@ -0,0 +1,108 @@
...
+        categoryOptionCombo,
+    }),
+}
+
+export default function useDeleteLimits(onDone) {
+    // These are needed for the optimistic delete
+    const queryClient = useQueryClient()
+    const [{ dataSetId: ds, orgUnitId: ou, periodId: pe }] =
+        useContextSelection()
+    const dataValueSetQueryKey = dataValueSets.byIds({ ds, pe, ou })
```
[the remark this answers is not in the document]
Mohammer5: #76

src/data-workspace/use-data-value-mutation/use-set-data-value-comment-mutation.js
Mohammer5: I don't understand this exactly. So in my head this would look like this:

1. The individual hooks format the next value. In order to do so, they need to know whether it's a new value or an update to an existing value
2. The hook then passes the updated value and a boolean whether this was an update or not to the optimistically update function

We'll end up with
* two different optimistic update functions (essentially "optimistically update" and "optimistically delete")
* either four new functions for formatting the slightly different values OR we'll add this to the hooks without exacting into a function

We'd either end up with more functions or we'd have optimistic update related code in two places (the hook and the helper), which seems worse that the current implementation (as there are four optimistic update related places for four slightly different update scenarios).

Maybe you meant something different though?

src/data-workspace/use-data-value-mutation/use-set-data-value-comment-mutation.js
```diff
@@ -0,0 +1,40 @@
...
+        return {
+            co,
+            de,
+            ds,
+            ou,
+            pe,
+
+            // when sending `undefined` something causes that
+            // to get stored as a string `"undefined"`
+            comment: comment ? comment : '',
```
kabaros, line 19:
```suggestion
            comment: comment ??  '',
```
would that work?

src/data-workspace/use-data-value-mutation/use-upload-data-value-file-mutation.js
```diff
@@ -6,7 +6,15 @@ import useMutationFn from './use-mutation-fn.js'
 export const UPLOAD_FILE_MUTATION = {
     resource: 'dataValues/file',
     type: 'create',
-    data: (data) => data,
+    data: ({ co, de, ds, ou, pe, value, file }) => ({
```
kabaros, line 9: again would be nice not to have these short forms as the input for the mutations, but probably should be tackled separately

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -1,65 +1,49 @@
...
-    //
-    // eslint-disable-next-line no-unused-vars
-    const [updateComment, { loading, error }] = useDataMutation(mutation, {
-        onError: () => errorAlert.show(),
-    })
-
-    if (loading) {
-        return <CircularLoader small />
-    }
+function CommentEditForm({ item, open, setOpen, syncComment, onCancel }) {
```
kabaros, line 17: General performance comment while testing adding/editing comments.

1. The Comment component keeps re-rendering while typing (minor)
2. The whole form re-renders twice when the comment is saved (maybe this is tackled in the other performance fixes we're doing)

chrome-capture-2022-6-27 (1)

kabaros (changes requested):

src/data-workspace/data-details-sidebar/audit-log.test.js
```diff
@@ -47,13 +59,16 @@ describe('<AuditLog />', () => {
         ).toBeInTheDocument()
     })
 
-    it('renders the item audit log once loaded', async () => {
+    // @TODO: Enable and fix when working on:
+    // https://dhis2.atlassian.net/browse/TECH-1281
+    it.skip('renders the item audit log once loaded', async () => {
```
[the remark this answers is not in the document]
Mohammer5: That test was covering the UI before I was adding api functionality.
But the UI has to be changed again, which is blocked by pending api changes.
Testing the current UI would be a waste of time because the UI will change anyway

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -1,65 +1,49 @@
...
+    const [{ dataSetId: ds, periodId: pe, orgUnitId: ou }] =
+        useContextSelection()
+    const onSubmit = (values) => {
+        const variables = {
+            ds,
```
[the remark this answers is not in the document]
Mohammer5: >  we should update SET_DATA_VALUE_COMMENT_MUTATION to expect the long form as input instead then map to the short form for the api. I think the least time we spend passing around the abbreviated variables in the code, the clearer it becomes. It's not very intuitive to know what ds, ou, pe is when you're new to dhis or the codebase.

Yes, 100% agreed. This is something I wanted to do for quite a while already. Ideally the only places where we use `ds`, `ou`, etc. is in the endpoint params object and nowhere else. There are many places where we use `dataElement` where it's actually the `dataElementId`. That's something I'd like to get consistens as well. These two things belong together to me, so I'd prefer to do that in a separate PR. I'll create a jira issue and add it to the post-release epic as it's not required for the release

src/data-workspace/data-details-sidebar/history.js
```diff
@@ -0,0 +1,73 @@
...
+                    title={i18n.t('Something went wrong loading the history')}
+                >
+                    <p>{dataValueContext.error.toString()}</p>
+                </NoticeBox>
+            </ExpandableUnit>
+        )
+    }
+
+    if (offline && !dataValueContext.data) {
+        return (
```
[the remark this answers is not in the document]
Mohammer5: To me that'd be over-abstraction as we're talking about 18 LoCs in this case.

* It wouldn't save LoCs as the new components would have prop types and boilerplate code, too
* As this is just a couple of lines, I think the duplication makes it easier to reason about the code as there's one less abstraction

The amount of work if we had to change these in some way is more or less nothing, despite having to edit two components. If this ever gets more complicated I wouldn't mind abstracting this away

src/data-workspace/data-details-sidebar/limits-delete-button.js
```diff
@@ -0,0 +1,55 @@
...
+            <Tooltip content={i18n.t('Not available offline')}>
+                <Button small secondary disabled>
+                    {label}
+                </Button>
+            </Tooltip>
+        )
+    }
+
+    const onClick = async () => {
+        await deleteLimit.mutate({
```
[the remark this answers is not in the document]
Mohammer5: You found a bug in the code, which I fixed. Deleting limits isn't working though, see here: https://dhis2.atlassian.net/browse/TECH-1290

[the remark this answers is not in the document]
Mohammer5: Once the endpoint has been fixed, this should work though

src/data-workspace/data-details-sidebar/comment.js
Mohammer5: Yes, this has been addressed by the performance fixes which have been merged yesterday

src/data-workspace/data-details-sidebar/audit-log.js
```diff
@@ -1,115 +1,161 @@
...
+                            value,
+                            created,
+                            auditType,
+                        } = audit
 
                         return (
-                            <li key={index} className={styles.entry}>
-                                <Entry entry={payload} />
-                            </li>
+                            <DataTableRow key={index}>
```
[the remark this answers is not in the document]
Mohammer5: They do not have an id, so I used a combination of the ids that are used to identify the current audit with the particular field

src/data-workspace/data-details-sidebar/comment.js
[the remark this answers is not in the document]
Mohammer5: https://dhis2.atlassian.net/browse/TECH-1293

src/data-workspace/use-data-value-mutation/use-upload-data-value-file-mutation.js
Mohammer5: https://dhis2.atlassian.net/browse/TECH-1293

src/data-workspace/data-details-sidebar/limits.test.js
```diff
@@ -1,87 +1,57 @@
-import { CustomDataProvider } from '@dhis2/app-runtime'
-import { waitFor } from '@testing-library/react'
-import userEvent from '@testing-library/user-event'
 import React from 'react'
 import { render } from '../../test-utils/index.js'
 import Limits from './limits.js'
 
 describe('<Limits />', () => {
-    it('is collapsed by default', () => {
+    it('is expanded by default', () => {
```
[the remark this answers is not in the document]
Mohammer5: No, not yet. These will follow during the soft-/hard-freeze as we're skipping adding tests unless it's really crucial in favor of getting things into the next release before the soft-freeze.

src/bottom-bar/data-item-bar.module.css
```diff
@@ -7,5 +7,15 @@
 
 .name {
     display: block;
-    margin-right: 4px;
+    margin-right: 8px;
```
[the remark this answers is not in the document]
Mohammer5: I agree that that would be good to do, it's a separate issue though as we've used this way of spacing things in the entire app: https://dhis2.atlassian.net/browse/TECH-1294

src/data-workspace/data-details-sidebar/comment.test.js
```diff
@@ -28,19 +43,149 @@ describe('<Comment />', () => {
...
+        fireEvent.change(input, {
+            target: { value: 'Changed' },
+        })
+        expect(input.value).toBe('Changed')
+
+        expect(queryByRole('progressbar')).not.toBeInTheDocument()
+        userEvent.click(getByRole('button', { name: 'Save comment' }))
+        await waitFor(() => {
+            expect(getByRole('progressbar')).toBeInTheDocument()
+        })
```
kabaros, line 99: nit (feel free to ignore): this could be `expect(findByRole('progressbar')).not.toBeNull()` without the need for await waitFor

src/data-workspace/data-details-sidebar/comment.test.js
```diff
@@ -28,19 +43,149 @@ describe('<Comment />', () => {
...
+        expect(
+            queryByRole('heading', {
+                name: 'There was a problem loading the comment for this data item',
+            })
+        ).not.toBeInTheDocument()
+
+        userEvent.click(getByRole('button', { name: 'Save comment' }))
+        await waitFor(() => {
+            expect(
+                getByRole('heading', {
```
kabaros, line 140: same can be `findByRole` instead of waitFor + get

kabaros (approved):

PR #74, opened 2022-06-21
feat(footer): add footer UI

Mohammer5:
Closes [TECH-1210](https://dhis2.atlassian.net/browse/TECH-1210)

* Refactors the layout a little: Puts the footer into the data workspace
  * I did this because the app is supposed to display the footer only when rendering a form. The data workspace renders the form conditionally and I didn't want to duplicate the conditions in the footer / app
  * Marking @ismay as a reviewer as he's the one who created this part of the app
* Adds the UI + functionality for the "data item bar" (the upper one; only displayed when focusing an item)
* Adds the UI for the "main tool bar", but not the functionality (the lower one; always displayed)
  * Marking @KaiVandivier as a reviewer as he's already working on the server side validation which includes the missing functionality in the footer's main tool bar)

EDIT: I talked to @ismay about the refactoring of the layout and we concluded that this topic can be seen from two different angles:

**The footer is a global part of the layout**
When looking at the UI specs, the footer could be regarded as app-wide. On the `development` branch, the layout is in one location (the layout component and its styles).

**The footer is a part of the data workspace**
When looking at the functionality and the position in the app, the footer seems to be a part of the data workspace itself. The condition when to render to footer is tightly coupled to what the data workspace component renders.

My motivation for moving the footer to the data workspace was that otherwise we'd have to either introduce state in the app component and set/update that state from within the data workspace component or duplicate the conditions and copy them from the data workspace component into the app component.

I'm wondering about how others see this. Do you think the footer is supposed to be an app-wide component and we should keep it in the layout component or does the footer's functionality and position in the app justify moving it to the data workspace?

_The `<MutationIndicator />`, that's been rendered by the footer is actually a global piece, but its current location is just a workaround for not being able to display that information in the headerbar right now_

### DIFF
--- a/src/data-workspace/data-entry-cell/entry-field-input.js
+++ b/src/data-workspace/data-entry-cell/entry-field-input.js
@@ -100,7 +100,6 @@ export function EntryFieldInput({
         const { key, shiftKey } = event
 
         if (shiftKey && key === 'Enter') {
-            currentItemContext.setItem(currentItem)
             rightHandPanel.show('data-details')
         } else if (key === 'ArrowDown' || key === 'Enter') {
             event.preventDefault()
@@ -112,6 +111,7 @@ export function EntryFieldInput({
     }
 
     const onFocus = () => {
+        currentItemContext.setItem(currentItem)
         rightHandPanel.hide()
     }
 
--- a/src/metadata/selectors.js
+++ b/src/metadata/selectors.js
@@ -72,13 +72,14 @@ export const getCategoryOptionsByCategoryOptionComboId = createCachedSelector(
     (metadata, categoryOptionComboId) => {
         const categoryCombos = Object.values(metadata.categoryCombos)
 
-        // "outer:" is a positional label that can be used to break multiple
-        // loops: https://stackoverflow.com/a/1564838
         for (const categoryCombo of categoryCombos) {
             for (const curCategoryOptionCombo of categoryCombo.categoryOptionCombos) {
                 if (curCategoryOptionCombo.id === categoryOptionComboId) {
-                    const categoryOptions = Object.values(metadata.categoryOptions)
-                    const { categoryOptions: curCategoryOptions } = curCategoryOptionCombo
+                    const categoryOptions = Object.values(
+                        metadata.categoryOptions
+                    )
+                    const { categoryOptions: curCategoryOptions } =
+                        curCategoryOptionCombo
 
                     return categoryOptions.filter(({ id }) =>
                         curCategoryOptions.includes(id)
--- a/src/metadata/selectors.js
+++ b/src/metadata/selectors.js
@@ -70,28 +70,24 @@ export const getCategoryOptionsByCategoryOptionComboId = createCachedSelector(
     (metadata) => metadata,
     (_, categoryOptionComboId) => categoryOptionComboId,
     (metadata, categoryOptionComboId) => {
-        let categoryOptionCombo
-
         const categoryCombos = Object.values(metadata.categoryCombos)
 
         // "outer:" is a positional label that can be used to break multiple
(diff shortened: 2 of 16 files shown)


src/metadata/selectors.js
```diff
@@ -66,6 +66,35 @@ export const getCategoryOptionCombosByCategoryComboId = (
...
+
+        if (!categoryOptionCombo) {
+            return []
+        }
+
+        return Object.values(metadata.categoryOptions).filter(({ id }) =>
+            categoryOptionCombo.categoryOptions.includes(id)
+        )
+    }
+)((_, categoryOptionComboId) => categoryOptionComboId)
```
HendrikThePendric, line 96: Wow, this is the first time I've seen the legendary position label used for real. I do think this can be simplied though to avoid it's usage:
```suggestion
export const getCategoryOptionsByCategoryOptionComboId = createCachedSelector(
    (metadata) => metadata,
    (_, categoryOptionComboId) => categoryOptionComboId,
    (metadata, categoryOptionComboId) => {
        const categoryCombos = Object.values(metadata.categoryCombos)

        for (const categoryCombo of categoryCombos) {
            for (const curCategoryOptionCombo of categoryCombo.categoryOptionCombos) {
                if (curCategoryOptionCombo.id === categoryOptionComboId) {
                    return Object.values(metadata.categoryOptions).filter(
                        ({ id }) =>
                            curCategoryOptionCombo.categoryOptions.includes(id)
                    )
                }
            }
        }

        return []
    }
)((_, categoryOptionComboId) => categoryOptionComboId)
```
I ran this locally and the tests still pass. But then I realised there wasn't a test for this one. So this means that:
- I'm not sure this will work as expected (I'm in a lecture and finding it hard to focus)
- We should add a test to verify this complex logic is working as expected

HendrikThePendric (commented):
I've left a few suggestions, but I will leave it to @KaiVandivier / @ismay / @Birkbjo to approve or request changes.

HendrikThePendric (changes requested):
Sorry, I changed my mind. Requesting changes because I think we need to add a test for the complex `getCategoryOptionsByCategoryOptionComboId` selector before this gets merged.

src/metadata/selectors.js
Mohammer5: I've added a test and implemented your suggestion

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -112,6 +111,7 @@ export function EntryFieldInput({
     }
 
     const onFocus = () => {
+        currentItemContext.setItem(currentItem)
         rightHandPanel.hide()
```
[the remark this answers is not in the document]
Mohammer5: Out of scope for this PR. We'll discuss this separately: https://dhis2.atlassian.net/browse/TECH-1211

src/metadata/selectors.js
```diff
@@ -66,6 +66,31 @@ export const getCategoryOptionCombosByCategoryComboId = (
...
  */
 
+export const getCategoryOptionsByCategoryOptionComboId = createCachedSelector(
+    (metadata) => metadata,
+    (_, categoryOptionComboId) => categoryOptionComboId,
+    (metadata, categoryOptionComboId) => {
+        const categoryCombos = Object.values(metadata.categoryCombos)
+
+        // "outer:" is a positional label that can be used to break multiple
+        // loops: https://stackoverflow.com/a/1564838
```
HendrikThePendric, line 76: Oops, in my suggested commit forgot I forgot to remove the comment which is now redundant.

HendrikThePendric (changes requested):
Please remove the comment, and run yarn format as well.

HendrikThePendric (approved):
🎉 awesome!

PR #77, opened 2022-06-27
feat(custom-forms): add support for total fields

HendrikThePendric:
Code-wise it would have made a lot more sense to replace the input tag, but that caused issues with the table-cell's background color. I couldn't get the content of the `td` to be filled completely, so I ended up with some whitespace inside the `td`. So I did it this way instead, replacing the `td`. Another advantage of this was that I could re-use the `TotalCell` which already came with the right styles.

To test this functionality, I recommend using the EPI Stock dataset.

### DIFF
--- a/src/data-workspace/custom-form/replace-td-node.js
+++ b/src/data-workspace/custom-form/replace-td-node.js
@@ -1,16 +1,31 @@
 import { attributesToProps } from 'html-react-parser'
 import React from 'react'
+import { TotalCell } from '../category-combo-table/total-cells.js'
 
 const TD_LABEL_CLASS = 'dhis2-data-entry-app-custom-form-label-cell'
 
-export const replaceTdNode = (domNode) => {
-    const hasSingleTextNodeChild =
-        domNode.children.length === 1 && domNode.children[0].type === 'text'
-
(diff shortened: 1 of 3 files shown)


src/data-workspace/custom-form/replace-td-node.js
```diff
@@ -1,16 +1,31 @@
...
+    return Object.values(values).reduce((sum, value) => {
+        if (!isNaN(value)) {
+            sum = isNaN(sum) ? value : sum + Number(value)
+        }
+        return sum
+    }, null)
+}
+
+const replaceTotalCell = (dataElementId, formState) => {
+    const total = computeTotal(formState.values[dataElementId])
```
KaiVandivier, line 23:
```suggestion
const replaceTotalCell = (dataElementId, formState) => {
    // Get values from all cells associated with this data element from form state and sum them.
    // Object passed to computeTotal should look like `{ [cocId1]: val1, [cocId2]: val2, ... }`
    const total = computeTotal(formState.values[dataElementId])
```

KaiVandivier (approved):
Looks good! Tested on the EPI stock dataset. Nice job handling non-numerical values smartly :)

Just added a comment suggestion that might help future code readers

PR #80, opened 2022-06-29
fix: display last updated date in basic info correctly

kabaros:
Closes [TECH-1220](https://dhis2.atlassian.net/browse/TECH-1220)

**Changes in this PR**
- Fixes the issue where date and user are displayed as null
- Also fixes an issue with the time not being displayed correctly  when the user is in a zone other than UTC
- Updates to tests

| Before | After |
| --- | -- |
| image | image |

### DIFF
--- a/src/data-workspace/data-details-sidebar/basic-information.js
+++ b/src/data-workspace/data-details-sidebar/basic-information.js
@@ -27,7 +27,7 @@ const BasicInformation = ({ item, onMarkForFollowup, onUnmarkForFollowup }) => (
             <li>
                 <Tooltip content={item.lastUpdated.toString()}>
                     {i18n.t('Last updated {{- timeAgo}} by {{- name}}', {
-                        timeAgo: moment(item.lastUpdated).fromNow(),
+                        timeAgo: moment.utc(item.lastUpdated).fromNow(),
                         name: item.storedBy,
                     })}
                 </Tooltip>
--- a/src/data-workspace/data-details-sidebar/basic-information.test.js
+++ b/src/data-workspace/data-details-sidebar/basic-information.test.js
@@ -1,19 +1,26 @@
 import { render } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import moment from 'moment'
 import React from 'react'
 import BasicInformation from './basic-information.js'
 
 const noop = () => {}
 
 describe('<BasicInformation />', () => {
+    beforeEach(() => {
+        jest.useFakeTimers('modern')
+        // set a fake time (note time zone is UTC+1 while API returns UTC)
+        jest.setSystemTime(new Date('2022-06-28T15:52+01:00').getTime())
+    })
+    afterEach(() => {
+        jest.useRealTimers()
+    })
     it('renders the item name in a heading', () => {
         const item = {
             categoryOptionCombo: 'coc-1',
             comment: null,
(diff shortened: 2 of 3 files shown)


src/data-workspace/data-details-sidebar/basic-information.js
```diff
@@ -27,7 +27,7 @@ const BasicInformation = ({ item, onMarkForFollowup, onUnmarkForFollowup }) => (
             <li>
                 <Tooltip content={item.lastUpdated.toString()}>
                     {i18n.t('Last updated {{- timeAgo}} by {{- name}}', {
-                        timeAgo: moment(item.lastUpdated).fromNow(),
+                        timeAgo: moment.utc(item.lastUpdated).fromNow(),
```
kabaros, line 30: I noticed that in my case, changes that I just did appeared as `an hour ago` because I am in UTC+1 while the API is returning UTC. This change makes it take into consideration the local timezone vs api timezone.

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -16,9 +16,10 @@ import { useDataValueParams } from './use-data-value-params.js'
 import { VALUE_TYPES } from './value-types.js'
 
 function createCurrentItem({ de, coc, dataValueSet }) {
-    if (dataValueSet?.data[de.id]?.[coc.id]) {
+    const dataValue = dataValueSet?.data.dataValues[de.id]?.[coc.id]
```
kabaros, line 19: this is the actual fix for date/name not being displayed correctly .. the data exists in `.dataValues`

kabaros, line 19: this is the actual fix for date/name not being displayed correctly .. the data exists in `.dataValues` .. @Mohammer5 @HendrikThePendric please let me know if this change might have some unintended consequences

src/data-workspace/data-details-sidebar/basic-information.test.js
```diff
@@ -1,19 +1,27 @@
...
+        jest.useRealTimers()
+    })
     it('renders the item name in a heading', () => {
         const item = {
             categoryOptionCombo: 'coc-1',
             comment: null,
             dataElement: 'item-1',
             name: 'Item name',
-            lastUpdated: new Date().toISOString(),
+            lastUpdated,
```
Mohammer5, line 24: This is just a suggestion and will leave it up to you and your preference. I'd prefer to duplicate the string in the mock data instead of extracting it into a variable.

---

Of course it's not bad/wrong whatsoever and subject to opinion. IMO this decreases readability of tests as you'll have to scroll to / remember the value when using tests as documentation. And I think the documentation-like role of tests is more important than reducing the amount of code / repetition.

src/data-workspace/data-entry-cell/entry-field-input.js
Mohammer5: This has been addressed already, but hasn't been merged: https://github.com/dhis2/data-entry-app/pull/73/files#diff-9ee92ce88ffa5ba79b5e4ef858a93314f19b500c331541899c6d368437ed3c74

I don't mind fixing it here either as it will make it into `development` much quicker than #73. The merge conflicts will be quite trivial to fix.

Mohammer5 (approved):

PR #85, opened 2022-07-04
TECH-1249-fix-performance-issues

HendrikThePendric:
**DO NOT MERGE**

This PR is currently at a stage where the data entry form is performant, but quite a lot of functionality is broken:
- Initial values are not fetched from the server
- Data is no longer refetched after a successful POST
- The active cell state is no longer updated

We need to first assess which of the broken functionality is actually crucial, and then how to implement that in a more performant way.


HendrikThePendric:
This branch was mostly to identify performance bottleneck rather than fixing them. Performance issues have been tackled already, so closing this.

PR #86, opened 2022-07-05
fix: make sure integer validation does not allow decimal

kabaros:
Closes [TECH-1230](https://dhis2.atlassian.net/browse/TECH-1230)

### Description
This fix ensures that a value such as `6.0` is not allowed by client side validation. Currently the client-side allows it, but then it's rejected by the API.

| Before | After |
| --- | --- |
| The value is accepted by client then rejected by API | client-side validation catches the decimal point |
| image | image |

### DIFF
--- a/src/data-workspace/inputs/generic-input.js
+++ b/src/data-workspace/inputs/generic-input.js
@@ -1,10 +1,4 @@
-import {
-    email,
-    integer,
-    internationalPhoneNumber,
-    number,
-    url,
-} from '@dhis2/ui-forms'
+import { email, internationalPhoneNumber, number, url } from '@dhis2/ui-forms'
 import PropTypes from 'prop-types'
 import React, { useState } from 'react'
 import { useField } from 'react-final-form'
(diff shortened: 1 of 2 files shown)


kabaros:
this is a potential fix in data-entry, but I think it would be cleaner to implement this logic in dhis/ui as part of the [integer validator](https://github.com/dhis2/ui/blob/master/collections/forms/src/validators/integer.js#L4)? @KaiVandivier @Mohammer5 @HendrikThePendric

Mohammer5:
> this is a potential fix in data-entry, but I think it would be cleaner to implement this logic in dhis/ui as part of the [integer validator](https://github.com/dhis2/ui/blob/master/collections/forms/src/validators/integer.js#L4)? @KaiVandivier @Mohammer5 @HendrikThePendric

I agree, this should be fixed in the [`integer` validator](https://github.com/dhis2/ui/blob/master/collections/forms/src/validators/integer.js) instead of here. I think your code is fine, so we could just add that fix to the validator, I don't think we have to treat "no decimals" and "integer" separately.

PR #87, opened 2022-07-05
refactor: optimize re-renders

Birkbjo:
Refactors the header in to a separate component. This PR is meant to make the `CategoryComboTable`-component a bit easier to reason about, as well as solve some re-renders.

- useActiveCell caused entire table to re-render
   - This is now only re-renders header-component and DataElementCell (header of the data-element)
- CurrentItemContext caused all cells to re-render because they subscribed to this context
   - Splitting this to a separate "setter"-context, solves this, as the cells only need to set the item

### DIFF
--- a/src/data-workspace/category-combo-table/category-combo-table-header.js
+++ b/src/data-workspace/category-combo-table/category-combo-table-header.js
@@ -2,62 +2,66 @@ import i18n from '@dhis2/d2-i18n'
 import { TableRowHead, TableCellHead } from '@dhis2/ui'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
-import React from 'react'
+import React, { useMemo } from 'react'
 import { useMetadata, selectors } from '../../metadata/index.js'
 import { useActiveCell } from '../data-entry-cell/index.js'
 import styles from './category-combo-table.module.css'
 import { PaddingCell } from './padding-cell.js'
 import { TotalHeader } from './total-cells.js'
 
+// Computes the span and columns to render in each category-row
+// columns are category-options, and needs to be repeated in case of multiple categories
+const useCategoryColumns = (categories, numberOfCoCs) => {
+    const { data: metadata } = useMetadata()
+    return useMemo(() => {
+        let catColSpan = numberOfCoCs
+        return categories.map((c) => {
+            const categoryOptions = selectors.getCategoryOptionsByCategoryId(
+                metadata,
+                c.id
+            )
+            const nrOfOptions = c.categoryOptions.length
+            // catColSpan should always be equal to nrOfOptions in last iteration
+            // unless anomaly with categoryOptionCombo-generation server-side
+            if (nrOfOptions > 0 && catColSpan >= nrOfOptions) {
+                // calculate colSpan for current options
+                // this is the span for each option, not the "total" span of the row
+                catColSpan = catColSpan / nrOfOptions
+                // when table have multiple categories, options need to be repeated for each disaggregation "above" current-category
+                const repeat = numberOfCoCs / (catColSpan * nrOfOptions)
+
+                const columnsToRender = new Array(repeat)
+                    .fill(0)
+                    .flatMap(() => categoryOptions)
+
+                return {
+                    span: catColSpan,
+                    columns: columnsToRender,
+                    category: c,
+                }
+            } else {
+                console.warn(
+                    `Category ${c.displayFormName} malformed. Number of options: ${nrOfOptions}, span: ${catColSpan}`
+                )
+            }
+            return c
+        })
+    }, [metadata, categories, numberOfCoCs])
+}
+
 export const CategoryComboTableHeader = ({
-    dataElements,
     renderRowTotals,
     paddingCells,
     categoryOptionCombos,
     categories,
+    checkTableActive,
 }) => {
-    const { data: metadata } = useMetadata()
     const { deId: activeDeId, cocId: activeCocId } = useActiveCell()
 
-    // Computes the span and repeats for each columns in a category-row.
-    // Repeats are the number of times the options in a category needs to be rendered per category-row
-    let catColSpan = categoryOptionCombos.length
-    const rowToColumnsMap = categories.map((c) => {
-        const categoryOptions = selectors.getCategoryOptionsByCategoryId(
-            metadata,
-            c.id
-        )
-        const nrOfOptions = c.categoryOptions.length
-        // catColSpan should always be equal to nrOfOptions in last iteration
-        // unless anomaly with categoryOptionCombo-generation server-side
-        if (nrOfOptions > 0 && catColSpan >= nrOfOptions) {
-            // calculate colSpan for current options
-            // this is the span for each option, not the "total" span of the row
-            catColSpan = catColSpan / nrOfOptions
-            // when table have multiple categories, options need to be repeated for each disaggregation "above" current-category
-            const repeat =
-                categoryOptionCombos.length / (catColSpan * nrOfOptions)
-
-            const columnsToRender = new Array(repeat)
-                .fill(0)
-                .flatMap(() => categoryOptions)
-
-            return {
-                span: catColSpan,
-                columns: columnsToRender,
-                category: c,
-            }
-        } else {
-            console.warn(
-                `Category ${c.displayFormName} malformed. Number of options: ${nrOfOptions}, span: ${catColSpan}`
-            )
-        }
-        return c
-    })
-
-    // Is the active cell in this cat-combo table? Check to see if active
-    // data element is in this CCT
-    const isThisTableActive = dataElements.some(({ id }) => id === activeDeId)
+    const rowToColumnsMap = useCategoryColumns(
+        categories,
+        categoryOptionCombos.length
+    )
 
     // Find if this column header includes the active cell's column
     const isHeaderActive = (headerIdx, headerColSpan) => {
@@ -65,7 +69,11 @@ export const CategoryComboTableHeader = ({
             (coc) => activeCocId === coc.id
         )
         const idxDiff = activeCellColIdx - headerIdx * headerColSpan
-        return isThisTableActive && idxDiff < headerColSpan && idxDiff >= 0
+        return (
+            checkTableActive(activeDeId) &&
+            idxDiff < headerColSpan &&
+            idxDiff >= 0
+        )
     }
 
     return rowToColumnsMap.map((colInfo, colInfoIndex) => {
@@ -109,22 +117,14 @@ export const CategoryComboTableHeader = ({
 }
 
 CategoryComboTableHeader.propTypes = {
-    categoryCombo: PropTypes.shape({
-        id: PropTypes.string.isRequired,
-        categoryOptionCombos: PropTypes.array,
-    }).isRequired,
     categories: PropTypes.array,
-    categoryOptionCombos: PropTypes.array,
-    dataElements: PropTypes.arrayOf(
+    // Note that this must be the sorted categoryoOptionCombos, eg. in the same order as they are rendered
+    categoryOptionCombos: PropTypes.arrayOf(
         PropTypes.shape({
-            id: PropTypes.string.isRequired,
-            categoryCombo: PropTypes.shape({
-                id: PropTypes.string,
-            }),
-            displayFormName: PropTypes.string,
-            valueType: PropTypes.string,
+            id: PropTypes.string,
         })
     ),
+    checkTableActive: PropTypes.func,
     paddingCells: PropTypes.array,
     renderRowTotals: PropTypes.bool,
 }
--- a/src/data-workspace/category-combo-table/category-combo-table-header.js
+++ b/src/data-workspace/category-combo-table/category-combo-table-header.js
@@ -1,0 +1,130 @@
+import i18n from '@dhis2/d2-i18n'
+import { TableRowHead, TableCellHead } from '@dhis2/ui'
+import cx from 'classnames'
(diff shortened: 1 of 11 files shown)


src/data-workspace/category-combo-table/category-combo-table-header.js
```diff
@@ -0,0 +1,130 @@
...
+
+export const CategoryComboTableHeader = ({
+    dataElements,
+    renderRowTotals,
+    paddingCells,
+    categoryOptionCombos,
+    categories,
+}) => {
+    const { data: metadata } = useMetadata()
+    const { deId: activeDeId, cocId: activeCocId } = useActiveCell()
```
HendrikThePendric, line 20: I'm thinking that it would probably benefit the app's architecture and performance if we were to push this form state subscription down to the level of the actual header cell.

src/data-workspace/category-combo-table/category-combo-table-header.js
```diff
@@ -0,0 +1,130 @@
...
+    const isThisTableActive = dataElements.some(({ id }) => id === activeDeId)
+
+    // Find if this column header includes the active cell's column
+    const isHeaderActive = (headerIdx, headerColSpan) => {
+        const activeCellColIdx = categoryOptionCombos.findIndex(
+            (coc) => activeCocId === coc.id
+        )
+        const idxDiff = activeCellColIdx - headerIdx * headerColSpan
+        return isThisTableActive && idxDiff < headerColSpan && idxDiff >= 0
+    }
```
HendrikThePendric, line 69: I was just wondering why we also need to `isThisTableActive` const. Is it possible that a form contains several tables with the same dataElements?

Look at this very naively, I would expect that every cell will be unique so we don't need this variable per se. If you have added that `isThisTableActive` variable to make the code more performant, then you should add it as an early return in the `isHeaderActive` function.

Having said all of that, if you agree with my earlier suggestion, about moving the active/inactive logic down to the actual header cell, then this code is going to change anyway.

HendrikThePendric (commented):
### Review 

Great work @Birkbjo 🦾 

I ran the app locally with my CPU throttled 4x. It is already much better than it was, and we still can make more performance optimisations in other areas (i.e. not resetting the initial values). So I'm feeling quite confident we can get this app optimised without having to change too much. 

I left a few minor comments. Looking forward to finding out what you think about them.

---

### Side note

When I tried the [Reproductive Health](https://pr-87--dhis2-data-entry.netlify.app/#/?dataSetId=QX4ZTUbOt3a&orgUnitId=ImspTQPwCqd&periodId=202201) form (which is quite a large form, so I tend to use it for performance checks), I did notice something strange though....

There are some empty cells at the end of the table, which don't seem to be needed....
Screenshot 2022-07-06 at 15 27 15

The old app doesn't have this issue so I guess we have a bug in our category-combo-table. Shall we try to fix this in the current PR, or shall I open a JIRA ticket for it?

src/data-workspace/category-combo-table/category-combo-table-header.js
Birkbjo: I don't really think it's necessary to push it further. You could in theory have a `CategoryTableHeaderColumns` and move it there, and basically move the rendered component from `columns.map...`. But that would only save re-rendering of the static components "Category name" and "PaddingCell". Again these are static, with no dynamic props and no DOM-updated are needed.

HendrikThePendric: I was mainly concerned about the nested iterations needed to compute the columns, and that fact that they would need to run whenever a the active cell changes. I suggested solving this by pushing the subscription to the active cell down, but I noticed you are now memoizing these columns in the `useCategoryColumns`. I think that's an equally valid solution.

src/data-workspace/category-combo-table/category-combo-table-header.js
HendrikThePendric: > I did some refactoring to not need to pass all `dataElements` to this component. So I pass a function `checkTableActive` instead. But yes, it's needed because section-forms render multiple tables, and we only want to highlight the active table.
> 
> > Is it possible that a form contains several tables with the same dataElements?
> 
> No, and that's why this fact is used to check if the current table is "active" or not.

Partly from your explanation above, and by looking at the code a bit more closely, I get this now 👍 .

---

> > then you should add it as an early return in the isHeaderActive function.
> 
> I don't really understand what you mean here?

I was mainly just figuring out what was happening, and my suggestion about returning early was basically just this (see the comment):

```js
    const isHeaderActive = (headerIdx, headerColSpan) => {
        if (!checkTableActive(activeDeId)) {
            return false
        }
        // by returning early you can avoid this iteration
        const activeCellColIdx = categoryOptionCombos.findIndex(
            (coc) => activeCocId === coc.id
        )
        const idxDiff = activeCellColIdx - headerIdx * headerColSpan
        return (idxDiff < headerColSpan && idxDiff >= 0)
    }
```

But this is very much a micro-optimisation, so I'm very happy to skip this.

Birkbjo:
> There are some empty cells at the end of the table, which don't seem to be needed.... Screenshot 2022-07-06 at 15 27 15
> 
> The old app doesn't have this issue so I guess we have a bug in our category-combo-table. Shall we try to fix this in the current PR, or shall I open a JIRA ticket for it?

This is a backend issue. Seems like someone has changed the categoryCombo, and as you probably know there's plenty of issues when doing this after it's created. 

This is also logged in the console:

```
Computed combination of categoryOptions for catCombo(m2jTvAj5kkm) is different from server.
            Please regenerate categoryOptionCombos.
            Computed: 12
            Server: 24) 
```

HendrikThePendric (approved):
Thanks for clarifying some things all good now 👍 

Reg. this:
> The old app doesn't have this issue so I guess we have a bug in our category-combo-table. Shall we try to fix this in the current PR, or shall I open a JIRA ticket for it?

Let's just defer this, since it's unrelated to your current work. I created [TECH-1255](https://dhis2.atlassian.net/browse/TECH-1255) for this.

Birkbjo:
> I was mainly just figuring out what was happening, and my suggestion about returning early was basically just this (see the comment):

    const isHeaderActive = (headerIdx, headerColSpan) => {
        if (!checkTableActive(activeDeId)) {
            return false
        }
        // by returning early you can avoid this iteration
        const activeCellColIdx = categoryOptionCombos.findIndex(
            (coc) => activeCocId === coc.id
        )
        const idxDiff = activeCellColIdx - headerIdx * headerColSpan
        return (idxDiff < headerColSpan && idxDiff >= 0)
    }
But this is very much a micro-optimisation, so I'm very happy to skip this.

Ah, I understand now! Yeah we could indeed do that, but yeah probably very micro-optimisation. `categoryOptionCombos.length` would probably never be a big number. They are basically the number of columns in the table. 
But I appreciate the thought, since this is a optimization PR afterall!

src/data-workspace/category-combo-table/category-combo-table-header.js
```diff
@@ -0,0 +1,130 @@
...
+        const activeCellColIdx = categoryOptionCombos.findIndex(
+            (coc) => activeCocId === coc.id
+        )
+        const idxDiff = activeCellColIdx - headerIdx * headerColSpan
+        return (
+            checkTableActive(activeDeId) &&
+            idxDiff < headerColSpan &&
+            idxDiff >= 0
+        )
+    }
```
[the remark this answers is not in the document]
Birkbjo: > I have one question for my own understanding: Why is using two providers (one "getter provider" and one "setter provider") functionally different from using a single provider? As I understanding it, it does seem to make a different, I just don't understand why yet. Could you explain that?


It's because when you are using context, you subscribe to the entire value of the context, and when it changes all subscribers re-render. So when the "item" part of the context-changes, all `EntryFieldInput`-components will re-render, even though they just need the "setItem"-part of the context. There's no way to choose to just subscribe to the setter-part. That's why a separate provider with just the setter-function makes it possible for this component to not re-render, they don't care about the actual value.

> I'm generally not a friend of abbreviations. It makes code more difficult to understand, in some cases it can even be ambiguous. Here it saves 2 characters (idx -> index) and at first I thought this is a weird plural of id). It also doesn't add any value other than saving typing time.


This is code that has been there before, I just moved it, and generally I'm not  a fan of refactoring others people's code, especially variable names when they're already in the codebase. And `idx` is a pretty usual abbreviation for index in programming, no? Is `diff` also too much abbreviation? Should it be called `indexDifference`? I think abbreviations have their place, and sometimes expanding everything can make it harder to keep track of variables etc. 


> Same applied to the map over rowToColumnsMap below, where sometimes it's colInfoIndex (column abbreviated, but Index is not) and columnIndex (where neither is abbreviated)


I mean, naming can be argued back and forth for hours. I agree these are not consistent but at some point the names would get too long. I think most understand what `colInfoIndex` means in this context.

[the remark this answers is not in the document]
Mohammer5: ## Question about the two providers

Thanks! It makes complete sense to me now :)

## Abbreviations

> This is code that has been there before, I just moved it

Did mean to say that this is your fault or something like that! Just commented on the code I saw :)

---

> I'm not a fan of refactoring others people's code, especially variable names when they're already in the codebase.

I wouldn't count un-abbreviating as refactoring, I don't think we have to be that sensitive when refactoring. Of course if you're worried that this is important to original author, you can request a review from that person and I'd even encourage you to do that!

---

> And idx is a pretty usual abbreviation for index in programming, no?

Despite "refactoring", what's the advantage of using the abbreviation other than saving 2 characters? There's a disadvantage which is worse in relation to saving two character: _Understandability_. Yes `idx` is probably used in many places, but the fact that it confused me at first already means there are "developer bubbles" which haven't seen it, so I'm still wondering: What's the advantage here?

---

> I think abbreviations have their place, and sometimes expanding everything can make it harder to keep track of variables etc.

Yes, you're right, I don't disagree. `id` is an abbreviation in itself, but it's so common that it's sort of its own word already. I'd argue you can say the same about `diff`. To me these are different from `idx` as that one is less common (I actually have seen `i` more often, which I'm actually OK with as that's as common as `id`), it also looks very similar to `ids`, which makes it mentally more taxing.

> expanding everything can make it harder to keep track of variables etc.

Coming back to this and `idx` specifically. I don't see why `index` makes it harder to keep track of variables? As mentioned, I wouldn't un-abbreviate everything (e.g. `de`, as it has an established meaning in the context of dhis2)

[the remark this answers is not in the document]
Birkbjo: I see that you marked this as old, but will just follow up with some minor comments. 

> Did mean to say that this is your fault or something like that! Just commented on the code I saw :)

I totally understand! Just wanted to make it clear and was mostly and argument for why I didn't want to touch too much of that code.


> I wouldn't count un-abbreviating as refactoring, I don't think we have to be that sensitive when refactoring. Of course if you're worried that this is important to original author, you can request a review from that person and I'd even encourage you to do that!

Hm... I think it's definitely refactoring. I might be too strict about my thoughts on refactoring and stuff, and probably need to learn to be a bit more lenient. Not sure what it is, but it's probably a bit about the respect of the other's code. It's like, who am I to say that "I don't like this code, let's refactor it to my liking". Of course, sometimes it warranted, and if the variables was to be like `i, x, c` - then yes, but here it's so minor I don't really think it's necessary. I do appreciate the thought to have more consistent code though, and maybe it's something we (I?) should weigh more, but at the end of the day we are multiple people working on the codebase, and a lot of time even the code I write is very inconsistent in regards to naming and other code-style (eg. `function` vs `const`). That's not to say I disregard it at all - but it's a balance of what is worth changing. But it's definitely something to keep in mind, and try to keep in check. 


> Coming back to this and idx specifically. I don't see why index makes it harder to keep track of variables? As mentioned, I wouldn't un-abbreviate everything (e.g. de, as it has an established meaning in the context of dhis2)

I didn't mean to comment on this example specifically. It was more in general. In this case I do agree with you; I do prefer the `index`-version. It was more that it feels so petty to me to change variable names in other people's code. Again, I might be weird that way. I probably have more against it if a change was to only rename variables - but here I did actually do some other changes, so it wouldn't be that unnatural to also rename them -  so I do understand your comment.

Birkbjo:
> 



> Actually, in regards to the provider thing, Dan Abramov said something about this: [facebook/react#15156 (comment)](https://github.com/facebook/react/issues/15156#issuecomment-474590693)
> 
> Option 2 seems to be similar to something I've suggested before: The only component inside the form that needs the context is `<EntryFieldInput />`. Instead of using two providers, inside the entry field input module, have a wrapper component that extracts the setter while the actual entry field input expects the setter as prop and is wrapped with `memo`. That seems more canonical to me.
> 
> I guess as the provider stuff already works, I don't see a need to change that now though! Just wanted to mention that there seems to be a more "official" solution as it's coming from Dan, might be worth considering that solution if we end up in this situation again

Thanks for the link, that's helpful! But our solution is his "option 1" and preferred solution. The other solutions will still cause re-renders. Yes, I guess they would be cheap - but they are still re-renders. So I have to say I prefer the current solution - even for the future. Of course, this is only needed if a value is subscribed to in a lot of components - this kind of optimisation shouldn't be that common. 

We are rendering the extra context-provider in the same `Provider`-component, so having an extra context is not really something other components really need to worry about. We just expose a separate hook for the setter.

Mohammer5:
> But our solution is his "option 1" and preferred solution

You're right! I thought about this differently, I read this like "Split providers if they handle different things on a conceptual level", but of course, you're right, in our case the value and setter are not used together, so this makes perfect sense! Never mind me, then!

PR #89, opened 2022-07-07
feat: add indicator support (TECH-1155)

HendrikThePendric:
### Checklist
This PR adds indicator support and also touched a few loosely related parts of the code:
- [x] Add indicator support to **default** forms
- [x] Add indicator support to **section** forms
- [x] Add indicator support to **custom** forms
- [x] Include `factor` in indicator calculations
- [x] Optimise re-rendering of custom forms
- [x] Optimise re-rendering of total fields
- [x] Add tests for indicator parsing and evaluation
- [x] Add comments to indicator parsing and evaluation code because it is fairly complex

---

### Known issues and blockers

- ~Currently we can't read indicators on default forms, so that part can't be implemented, see [TECH-1265](https://dhis2.atlassian.net/browse/TECH-1265)~
- ~Currently we can't display indicators with a factor correctly because the factor is not returned in the response data, see [TECH-1266](https://dhis2.atlassian.net/browse/TECH-1266)~
- Currently the `useBlurredField` hook causes the totals and indicators to be recomputed fairly often. This is due to bug [TECH-1252](https://dhis2.atlassian.net/browse/TECH-1252), which is causing the form to be re-initialised on each field blur. Due to this re-initialising the `active` property temporarily has an `undefined` when navigating from one field to the next. I have verified that once the form stops re-initialising on every re-render, things start behaving as expected.

--- 

### Notes
- Probably we can defer this, since the old app doesn't handle this very gracefully at all either, but it seems to be a good idea to show a human-readable version of the value type somewhere. Especially for these values that have a factor it seems we are omitting vital information. For example, suppose an indicator is showing a percentage and the displayed value is `0.5`, then actually this means "half a percent" so the actual numerical value is `0.005`. We probably need to show this somewhere.
- This PR includes some some level of optimisation in the indicator calculations. The expensive calculation is only executed when a related field is blurred. But the `useIndicatorValue` hook will still cause the component to re-render on each blur. Since these are all re-renders with the same value, I don't think this should be an issue. But if it is, we can optimise some more....


src/data-workspace/category-combo-table-body/use-value-matrix.js
```diff
@@ -0,0 +1,46 @@
...
+    const affectedFieldsLookup = useMemo(
+        () =>
+            new Set(
+                dataElements.flatMap((de) =>
+                    sortedCOCs.map((coc) => `${de.id}.${coc.id}`)
+                )
+            ),
+        [dataElements, sortedCOCs]
+    )
+    const containsAffectedField = useMemo(
```
Birkbjo, line 32: I don't think `useMemo` is needed here. `set.has` is O(1), and this is a primitive value, so using it directly should be fine and is probably more performant than `useMemo` would be.

src/shared/use-blurred-field.js
```diff
@@ -0,0 +1,18 @@
+import { useState, useEffect, useRef } from 'react'
+import { useFormState } from 'react-final-form'
+
+export const useBlurredField = () => {
+    const previouslyActiveFieldRef = useRef(null)
+    const [blurredField, setBlurredField] = useState(null)
```
Birkbjo, line 6: Let's try to avoid using `setState`, since that will cause an additional re-render. See how a `usePrevious`-hook can be implemented eg. [here](https://usehooks.com/usePrevious/). I think that this can be used here. 

Basically:

```javascript
const previouslyActiveFieldRef = useRef(null)
const { active } = useFormState({ subscription: { active: true } })
useEffect(() => {
        if (active) { // not sure if this check is needed
            previouslyActiveFieldRef.current = active
        }
    }, [active])

// this runs before useEffect, so this contains previous value
return previouslyActiveFieldRef.current
```

src/data-workspace/indicators-table-body/use-indicator-value.js
```diff
@@ -3,43 +3,90 @@ import { useMemo, useState, useEffect } from 'react'
...
 const evaluate = (expression) => parser.parse(expression).evaluate()
-const formulaPattern = /#\{.+?\}/g
-const separator = '.'
 
-const getDataElementTotalValue = (dataElementCocs) => {
-    if (!dataElementCocs || typeof dataElementCocs !== 'object') {
+/*
+ * In an indicator expression template an `id` can be one of:
+ * 1. #{deId} for a data element
+ * 2. #{deId.cocId} for a category option combo
```
Birkbjo, line 24: Wouldn't this be the `dataElementOperand`? Eg. the `categoryOptionCombo` for a particular `dataElement`.

src/data-workspace/indicators-table-body/use-indicator-value.js
```diff
@@ -84,7 +141,11 @@ export const useIndicatorValue = ({ denominator, numerator, factor }) => {
     )
 
     useEffect(() => {
-        const dataElementId = blurredField?.split(separator)[0]
+        /**
+         * Only recompute the indicator value when a field related to
+         * a data element in the indicator expression template is blurred.
+         */
+        const dataElementId = blurredField?.split(idSeparator)[0]
```
Birkbjo, line 169: Let's use `parseFieldId`, from `get-field-id.js`.

kabaros (approved):
the PR looks great to me .. great that there are tests for the hardest bits which help understand them (and also trust the useMemo optimisations are working as expected) .

I also don't quite get how `useValueMatrix` works - a test for that hook might help me to understand what's happening.

I have a general comment/opinion about the UI: it feels to me like having the indicators in a separate table would make it easier to notice and understand, with just two columns for indicator name and value. Right now, I expect them to have totals in each column for the rows above them which is not the case. But maybe just a UI pattern I am not used to yet.

src/data-workspace/default-form.js
```diff
@@ -14,10 +15,16 @@ export function DefaultForm({ dataSet, globalFilterText }) {
     }
 
     const dataElements = selectors.getDataElementsByDataSetId(data, dataSet.id)
+    const indicators = selectors.getIndicatorsByDataSetId(data, dataSet.id)
     const groupedDataElements = selectors.getGroupedDataElementsByCatCombo(
         data,
         dataElements
     )
+    const nrColumnsInTable = groupedDataElements.reduce(
+        (max, { categoryCombo: { categoryOptionCombos } }) =>
```
Birkbjo, line 24: So we previously did this in `section.js` as well, but that can cause problems when the server returns the wrong `CategoryOptionCombos`. The table will then render empty padding-cells if the server returns more "cocs" than it should. 
Of course, this is a server-bug, but we fixed this problem in `section.js` by calculating the amount of `cocs` by the actual present `categoryOptions`:


```javascript
    // calculate how many columns in each group
    const groupedTotalColumns = groupedDataElements.map((grp) =>
        (
            selectors
                .getCategoriesByCategoryComboId(data, grp.categoryCombo.id)
                ?.map((cat) => cat.categoryOptions.length) || [1]
        ).reduce((total, curr) => total * curr)
    )

    const nrColumnsInTable = Math.max(...groupedTotalColumns)
```

Birkbjo (changes requested):
Overall looks very good! Great that we can avoid using `eval`!

src/shared/use-blurred-field.js
```diff
@@ -0,0 +1,13 @@
+import { useMemo, useRef } from 'react'
+import { useFormState } from 'react-final-form'
+
+export const useBlurredField = () => {
```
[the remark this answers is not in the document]
HendrikThePendric: So `active` returned by RFF `useFormState` represents the currently active form field ([see docs](https://final-form.org/docs/final-form/types/FormState)). By subscribing to `active` only, the `useFormState` hook will only update when the active field changes. When an active field changes, the previously active field must be the blurred field.

src/data-workspace/default-form.js
HendrikThePendric: Thanks for pointing this out @Birkbjo 🙇 

I ended up implementing this functionality in its most simple form in https://github.com/dhis2/data-entry-app/pull/89/commits/d57812b123498c59224561cc37398169a454969f. 

I liked the idea of having a cached selector, but I couldn't get this to work. I tried just copy-pasting the last two snippets you posted but that didn't run. I got a bit 🤯 ... I think the complexity lies in the fact that this selector would also need the `groupedDataElements`, and these are computed conditionally...

Also, adding `useMemo` was problematic, because we're doing the early return for `isLoading || isError`, so that makes the hook conditional. Moving everything above the early return to avoid that also seemed a bit problematic. I decided to just leave things like this for now. This change wasn't completely out-of-scope anyway, I simply happened to find it when I was adding indicator support to default forms.

Birkbjo: > I liked the idea of having a cached selector, but I couldn't get this to work. I tried just copy-pasting the last two snippets you posted but that didn't run. 

Yeah, I'm sorry, I just wrote that code directly in the comment. Was missing a parenthesis at the end at the `keySelector`-function. Also missed a `y` in `cat.categoryOptions.length`... Will test my code next time!

> Also, adding useMemo was problematic, because we're doing the early return for isLoading || isError, so that makes the hook conditional. 

We should probably remove this regardless, as that is handled in `LoadApp`, and not necessary to handle further down the tree.

Birkbjo (approved):

PR #90, opened 2022-07-07
fix: ensure empty values are not synced

kabaros:
Fixes [TECH-1232](https://dhis2.atlassian.net/browse/TECH-1232)

sync-if-not-empty-after (1)

### DIFF
--- a/src/data-workspace/inputs/generic-input.js
+++ b/src/data-workspace/inputs/generic-input.js
@@ -84,9 +84,10 @@ export const GenericInput = ({
 
     const handleBlur = () => {
         const { value } = input
+        const hasEmptySpaces = value && value.trim() === ''
         const { dirty, valid } = meta
-        if (dirty && valid && value !== lastSyncedValue) {
-            syncData(value)
+        if (dirty && valid && !hasEmptySpaces && value !== lastSyncedValue) {
+            syncData(value.trim())
         }
     }
 


Birkbjo:
How would we otherwise delete a value from a field if it's not synced? Or does this only prevent syncing if it was already empty?

kabaros:
> How would we otherwise delete a value from a field if it's not synced? Or does this only prevent syncing if it was already empty?

it does allow deleting an existing value, but that's because of a JS peculiarity. Basically `Boolean('')` is `true` but `Boolean(' ')` (with an empty space)  is `false`. So that `isEmpty` is only true if it's empty spaces, but not if it's a string of length zero.

I could change the expression to `const hasEmptySpaces = value && value !== '' && value.trim() === ''` to be more explicit?

Birkbjo:
> it does allow deleting an existing value, but that's because of a JS peculiarity. Basically `Boolean('')` is `true` but `Boolean(' ')` (with an empty space) is `false`. So that `isEmpty` is only true if it's empty spaces, but not if it's a string of length zero.

I see. But I think you mean the opposite? But yeah, the `value` would not be truthy in the case of empty string, so I understand how it works now. 

> I could change the expression to `const hasEmptySpaces = value && value !== '' && value.trim() === ''` to be more explicit?

I think the variable change could be beneficial, but I don't think the extra check for empty string is needed.

kabaros:
> I see. But I think you mean the opposite? But yeah, the `value` would not be truthy in the case of empty string, so I understand how it works now.

yeah I did mean the opposite. Updated the variable name now.

Birkbjo (approved):

PR #92, opened 2022-07-08
refactor: fix/debounce dataelement search alternative1

Mohammer5:


src/shared/use-debounce.js
```diff
@@ -12,10 +12,15 @@ export function useDebounce(value, delay) {
...
-export function useDebounceCallback(value, cb, delay = 200) {
-    const debouncedValue = useDebounce(value, delay)
+export function useDebounce(callback, delay) {
+    const timeoutIdRef = useRef(null)
 
-    useEffect(() => {
-        cb(debouncedValue)
-    }, [debouncedValue, cb])
+    return useCallback((...args) => {
+        if (timeoutIdRef.current) {
```
Birkbjo, line 19: A minor point; but I don't think this if-statement is needed. 
I think just doing `clearTimeout(timeoutIdRef.current)` should suffice, since `clearTimeOut(undefined)` does nothing.

Mohammer5: Ah yeah, that's true. What do you want to do with this PR?
We've talked about this and came to the conclusion that we could use https://github.com/xnimorz/use-debounce. What's your current opinion on this?

Birkbjo (approved):

PR #93, opened 2022-07-12
feat: add server-side validation

kabaros:
implements [TECH-1153](https://dhis2.atlassian.net/browse/TECH-1153)

## Changes in this PR
- [x] call API to get server-side validation result
- [x] implement UI to display results
- [x] UI - empty state
- [x] UI - loading state
- [x] UI - error state
- [x] implement re-run validation
- [x] add tests

### The following will be tackled in separate PRs:
- implement changed data notification
- investigate and implement comments required violations
- refactor to not have a second call for validation metadata once endpoint is implemented
- export results

## Screenshots
<p align="center">
image
</p>

### error state
<p align="center">
image
</p>

### loading state
<p align="center">
image
</p>

### empty state
<p align="center">
image
</p>


.github/workflows/dhis2-preview-pr.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             - uses: actions/checkout@v2
             - uses: actions/setup-node@v1
               with:
-                  node-version: 12.x
+                  node-version: 14.x
```
kabaros, line 27: this change is needed to be able to use `msw` library .. The current LTS is v16 so I thought it might not be a bad idea to use a newer version anyhow, but let me know if this has other implications and I can change it back.

src/test-utils/with-app-test-wrapper.js
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from 'prop-types'
+import React from 'react'
+import { MemoryRouter as Router, Route } from 'react-router-dom'
+import { QueryParamProvider } from 'use-query-params'
+import { CurrentItemProvider } from '../shared/index.js'
+import QueryClientWrapper from './query-client-wrapper.js'
+
+export const withAppTestWrapper = (
+    component,
+    initialUrl = '?dataSetId=BfMAe6Itzgt&orgUnitId=DiszpKrYNg8&periodId=202202'
```
Mohammer5, line 11: IMO this should be:

```suggestion
    initialUrl = ''
```

This special initial url should be moved to the validation results sidebar test as it's being used there only

src/test-utils/with-app-test-wrapper.js
```diff
@@ -0,0 +1,27 @@
+import { DataProvider } from '@dhis2/app-runtime'
+import PropTypes from 'prop-types'
+import React from 'react'
+import { MemoryRouter as Router, Route } from 'react-router-dom'
+import { QueryParamProvider } from 'use-query-params'
+import { CurrentItemProvider } from '../shared/index.js'
+import QueryClientWrapper from './query-client-wrapper.js'
+
+export const withAppTestWrapper = (
```
Mohammer5, line 9: I think a better strategy could be to have a custom render function: https://testing-library.com/docs/react-testing-library/setup/#custom-render.

If we need a different initial url, I'd propose to expose the individual building blocks used by the custom render function, e.g. `withAppTestWrapper` and `originalRender`, from the `test-utils.js` module as well

KaiVandivier (commented):
This looks really nice! 🙂 I’ve tested it out on a few datasets with validation rule violations on `debug/dev`(like Child Health and Mortality < 5 years; in Ngelehun for May 2022) and it seems to be working great 👍 

I noticed a few UI things in manual testing:
* When clicking on an entry field in the form, the sidebar closes (this only happens when focusing entry fields: clicking on a filter field, context selection dropdown, or other part of the main app doesn’t close the sidebar) — this might be the thing that @Jan-Gerke mentioned PR #73 will fix?
* (Very minor): when navigating to a new data set while the sidebar is open, the sidebar stays open while the new set is loading — this may not be the desired UX
* Then “Run validation” button in the bottom bar doesn’t do anything when the validation sidebar is open. This isn’t specc’ed, but I’d propose one of two things happening: 1. run validation again (which might be tricky to do given the separate components) or 2. toggle the sidebar

The testing strategy looks good 🙂 `msw` looks cool, but I don’t have strong opinion on the tool used — I’ll leave that to others. Once concern I can mention though is that checking `src/mocks/responses/metadata.js` into source control is pretty big at 45k lines 😬

Other than that, I have some suggestions that I’ll add alongside the code :)

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+    const validationQueryKey = getValidationQueryKey(datasetId, {
+        periodId,
+        orgUnitId,
+        categoryComboId,
+        categoryOptionIds,
+    })
+    const validationMetaDataQuery = getValidationMetaDataQueryKey(datasetId)
+
+    const activeMutations = useIsMutating({ mutationKey: validationQueryKey })
+    const enabled = activeMutations === 0
```
KaiVandivier, line 71: Maybe `useIsFetching` would be more semantic here?

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+        attributeOptions: categoryOptionIds,
+    } = useApiAttributeParams()
+
+    const validationQueryKey = getValidationQueryKey(datasetId, {
+        periodId,
+        orgUnitId,
+        categoryComboId,
+        categoryOptionIds,
+    })
+    const validationMetaDataQuery = getValidationMetaDataQueryKey(datasetId)
```
KaiVandivier, line 68: Some naming suggestions in this file I think would make it easier to read:

```suggestion
    const validationMetaDataQueryKey = getValidationMetaDataQueryKey(datasetId)
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+
+    const activeMutations = useIsMutating({ mutationKey: validationQueryKey })
+    const enabled = activeMutations === 0
+
+    const results = useQueries({
+        queries: [
+            { queryKey: validationQueryKey, enabled },
+            { queryKey: validationMetaDataQuery, enabled },
```
KaiVandivier, line 76:
```suggestion
            { queryKey: validationMetaDataQueryKey, enabled },
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+        ],
+    })
+
+    const refetch = useCallback(() => {
+        results.forEach((result) => result.refetch())
+    }, [results])
+
+    const [validationQuery, metaDataQuery] = results
```
KaiVandivier, line 84:
```suggestion
    const [validationQuery, validationMetaDataQuery] = results
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+        isRefetching:
+            validationQuery.isRefetching || metaDataQuery.isRefetching,
+        data:
+            validationQuery.isSuccess && metaDataQuery.isSuccess
+                ? groupValidationRules(validationQuery.data, metaDataQuery.data)
+                : undefined,
+    }
+}
+
+const groupValidationRules = (
```
KaiVandivier, line 99:
```suggestion
const groupValidationRuleViolationsByImportance = (
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+
+    return {
+        refetch,
+        error: validationQuery.error || metaDataQuery.error,
+        isLoading: validationQuery.isLoading || metaDataQuery.isLoading,
+        isRefetching:
+            validationQuery.isRefetching || metaDataQuery.isRefetching,
+        data:
+            validationQuery.isSuccess && metaDataQuery.isSuccess
+                ? groupValidationRules(validationQuery.data, metaDataQuery.data)
```
KaiVandivier, line 94:
```suggestion
                ? groupValidationRuleViolationsByImportance(validationQuery.data, metaDataQuery.data)
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+
+    const validationQueryKey = [
+        `validation/dataSet/${datasetId}`,
+        {
+            params,
+        },
+    ]
+    return validationQueryKey
+}
+export const useValidationResult = () => {
```
KaiVandivier, line 52:
```suggestion

/** Returns validation rule violations grouped by importance */
export const useValidationResult = () => {
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,118 @@
...
+
+const getValidationQueryKey = (
+    datasetId,
+    { periodId, orgUnitId, categoryComboId, categoryOptionIds }
+) => {
+    const params = {
+        pe: periodId,
+        ou: orgUnitId,
+        cc: categoryComboId,
+        cp: categoryOptionIds?.join(';'),
```
KaiVandivier, line 41: Have you verified if these two parameters work with this API? I haven't seen this in the API docs and so far I haven't found a dataset with a non-default attribute combo _and_ val'n rule violations to test it out, but what I would look for is that this filters the validation rule violations in the response by the provided attribute combo and options

Maybe I can set up a dataset to test this out

src/data-workspace/validation/validation-priority-group-unit.js
```diff
@@ -0,0 +1,85 @@
...
+    const translationOptons = {
+        level: validationConfig.text?.toLowerCase(),
+        length: validationViolations.length,
+    }
+
+    return (
+        <div data-test={`priority-group-${level}`}>
+            <div className={styles.titleWrapper}>
+                <div className={iconStyle}>
+                    <Icon />
```
KaiVandivier, line 53: Icons have a `color` prop that accepts a string (that is sadly missing from our documentation 😭) that may simplify the styling considerably -- maybe an `iconColor` property could be added to the `validation-config`?

src/data-workspace/validation/validation-priority-group-unit.module.css
```diff
@@ -0,0 +1,67 @@
+.titleWrapper {
+    display: flex;
+    align-items: center;
```
KaiVandivier, line 3:
```suggestion
    align-items: center;
    gap: 6px;
```

You might consider this instead of adding the margin to the icon 🙂

src/data-workspace/validation/validation-priority-group-unit.js
```diff
@@ -0,0 +1,85 @@
...
+} from './validation-result-prop-types.js'
+
+const ValidationPriortyGroupUnit = ({ level, validationViolations = [] }) => {
+    if (validationViolations?.length === 0) {
+        return null
+    }
+
+    const validationConfig = validationLevelsConfig[level]
+
+    const style = cx(styles.priorityGroupBox, styles[validationConfig.style])
```
KaiVandivier, line 19:
```suggestion
    const validationViolationBoxStyle = cx(styles.priorityGroupBox, styles[validationConfig.style])
```

src/data-workspace/validation/validation-priority-group-unit.js
```diff
@@ -0,0 +1,85 @@
...
+                          )
+                        : i18n.t(
+                              '{{length}} {{level}} priority alerts',
+                              translationOptons
+                          )}
+                </h1>
+            </div>
+            {validationViolations.map((validationRule) => {
+                return (
+                    <div className={style} key={validationRule.metaData.id}>
```
KaiVandivier, line 69:
```suggestion
                    <div className={validationViolationBoxStyle} key={validationRule.metaData.id}>
```

src/right-hand-panel/right-hand-panel.js
```diff
@@ -1,11 +1,17 @@
...
 import { DataDetailsSidebar } from '../data-workspace/index.js'
+import ValidationResultsSidebar from '../data-workspace/validation/validation-results-sidebar.js'
 import RightHandPanelContext from './right-hand-panel-context.js'
 import styles from './right-hand-panel.module.css'
 
 const idSidebarMap = {
-    'data-details': DataDetailsSidebar,
+    [dataDetailsSidebarId]: DataDetailsSidebar,
+    [validationResultsSidebarId]: ValidationResultsSidebar,
     'contextual-help': ContextualHelpSidebar,
```
KaiVandivier, line 15: Could maybe use a constant for the `contextual-help` key :)

src/data-workspace/validation/validation-result-prop-types.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types'
+
+export const ValidationRulePropTypes = PropTypes.shape({
```
KaiVandivier, line 3:
```suggestion
export const ValidationRuleViolationWithMetaDataPropTypes = PropTypes.shape({
```

src/data-workspace/validation/validation-results-sidebar.js
```diff
@@ -0,0 +1,86 @@
...
+import ValidationSummaryBox from './validation-summary-box.js'
+
+export default function ValidationResultsSidebar() {
+    const rightHandPanel = useRightHandPanelContext()
+    const { isLoading, isRefetching, error, data, refetch } =
+        useValidationResult()
+    const showLoader = isLoading || isRefetching
+
+    const isEmpty =
+        data && Object.values(data).every((rules) => rules.length === 0)
```
KaiVandivier, line 19:
```suggestion
        data && Object.values(data).every((ruleViolations) => ruleViolations.length === 0)
```

src/data-workspace/validation/validation-results-sidebar.module.css
```diff
@@ -0,0 +1,38 @@
+.wrapper {
+    margin: 16px
+}
+
+.butttons :global(button) {
+    margin-inline-end: 4px
+}
```
KaiVandivier, line 7: Consider: (also notice typo in "butttons")

```suggestion
.buttons {
    display: flex;
    gap: 4px;
}
```

src/data-workspace/validation/validation-results-sidebar.module.css
```diff
@@ -0,0 +1,38 @@
...
+
+.title {
+    color: var(--colors-grey900);
+    font-size: 16px;
+    line-height: 22px;
+    font-weight: 500;
```
KaiVandivier, line 13: Noticed this in the design spec 🙂

```suggestion
    font-weight: 500;
    margin-block-start: 8px;
    margin-block-end: 8px;
```

src/data-workspace/validation/validation-results-sidebar.module.css
```diff
@@ -0,0 +1,38 @@
...
+    justify-content: center;
+    align-items: center;
+    align-content: center;
+    font-size: 14px;
+    line-height: 19px;
+    color: var(--colors-grey700)
+}
+.loading .text {
+    margin-inline-start: 8px;
+}
```
KaiVandivier, line 33: Consider:

```suggestion
    color: var(--colors-grey700)
    gap: 8px;
}
```

src/data-workspace/validation/validation-summary-box.js
```diff
@@ -0,0 +1,29 @@
...
+const ValidationSummaryBox = ({ level, count, hideCounts }) => {
+    const { icon: Icon, text, style: styleName } = validationLevelsConfig[level]
+    const style = styles[styleName]
+    const classNames = cx(styles.summaryBox, style, {
+        [styles.loading]: hideCounts,
+    })
+    return (
+        <div data-test={`count-${level}`} className={classNames}>
+            <div className={styles.title}>
+                <Icon />
```
KaiVandivier, line 17: Consider using the `color` prop on the icon here to simplify styles

KaiVandivier:
One thing I noticed, which I don't think has to do with the content of this PR: the names of data elements can be inconsistent in these three places 🤔 (validation rule name, `validationRule.leftSide.description`, and the data element's display name)

Screen Shot 2022-07-20 at 6 27 27 PM

KaiVandivier:
Ah one other thing I noticed in the spec: it looks like we want to show the values of the relevant data items in the validation expression

Screen Shot 2022-07-21 at 2 32 41 PM

kabaros:
> Ah one other thing I noticed in the spec: it looks like we want to show the values of the relevant data items in the validation expression
> 
> Screen Shot 2022-07-21 at 2 32 41 PM

Updated to show the value in the validation expression

image

src/test-utils/render.js
```diff
@@ -25,14 +31,12 @@ export function TestWrapper({
...
 export function render(ui, options = {}) {
-    const queryClient = new QueryClient()
-
     return renderOrig(ui, {
         ...options,
         wrapper: ({ children }) => (
-            <TestWrapper wrapper={options.wrapper} queryClient={queryClient}>
-                {children}
-            </TestWrapper>
+            <DataProvider baseUrl="http://dhis2-tests.org" apiVersion="39">
```
kabaros, line 37: I've had to make a couple of small changes for the test-utils. The query client needed to be inside a Data Provider in order to work, so I added it here .. passing it as `wrapper` property didn't work because in `TestWrapper`, we're rendering the `content` then injecting it to the wrapper so the content doesn't have access to the provider. I could change that logic instead but this looked more straightforward; leaving DataProvider for tests that need it and it won't affect those that don't.

src/test-utils/render.js
```diff
@@ -1,18 +1,24 @@
...
 import { OuterComponents } from '../app/app-wrapper.js'
+import { useTestQueryClient } from './index.js'
 
 export function TestWrapper({
     wrapper: WrapperFromOptions,
     router = MemoryRouter,
     queryClient,
     children,
 }) {
+    const defaultTestQueryClient = useTestQueryClient()
```
kabaros, line 15: since I have to use a hook under the hood (`useDataEngine()`), then this needs to live in a React component so made it the default here. Not sure if there is a non-hook way to get hold of the data engine?

kabaros (commented):
@Mohammer5 I updated the tests to use the new render function .. there are a couple of changes I needed to make to get it to work, so please have a look

src/test-utils/render.js
Mohammer5: > Not sure if there is a non-hook way to get hold of the data engine?

Unfortunately there isn't.. 
I'm OK with the change, please import `useTestQueryClient` from `'./use-test-query-client.js`, not from the `index.js`, which is the entrypoint to this "namespace" from the outside.

kabaros: > I'm OK with the change, please import useTestQueryClient from './use-test-query-client.js, not from the index.js, which is the entrypoint to this "namespace" from the outside.

done

src/test-utils/render.js
Mohammer5: I'm wondering if it'd make more sense to use the `CustomDataProvider` here. But that's a discussion we'll have to defer until we talked about how to proceed with msw

kabaros: ok, I can use CustomDataProvider then and will keep the msw tests separately for a discussion. Switching the tests to CustomDataProvider would help me as well to see what works and doesn't in practice compared to msw

Mohammer5:
> I feel it might be better to hold off on that until we've had the chance to properly discuss the proposal.

I second this opinion upon reflecting what's happening here. For now I'd prefer if we simply use the `CustomDataProvider` as that's our current standard.

We could create a branch/PR based on this one that introduces msw which gives us time to talk about it, best practices, standards we want to follow, how to integrate the library into this app, etc.
Additionally we can start a proper discussion in our [dhis2/notes](https://github.com/dhis2/notes) repository.

src/data-workspace/validation/use-validation-result.js
KaiVandivier: Yeah I suppose that's correct -- is your intention to disable these queries while a new data value is being sent to the server, or something else? (If it's the first, maybe you could use `isMutating` with some [mutation filters](https://tanstack.com/query/v4/docs/guides/filters#mutation-filters) to pause these queries while syncing data)

kabaros: yeah I think that's the intention. I added after a remark from @Mohammer5 

> We probably should check that there are no active mutations (see the useDataValueSet hook).

So I copied the logic from useDataValueSet. But I think the mutationKey I am checking against doesn't make sense.

src/data-workspace/validation/use-validation-result.js
KaiVandivier: Sure! I tried it out on the "Mortality < 5 years" data set on debug/dev - it takes opening up the data set's options page in the Maintenance app ([link](https://debug.dhis2.org/dev/dhis-web-maintenance/index.html#/edit/dataSetSection/dataSet/pBOMPrpg1QX)), and I set the "Category combination" option to "Target vs. Result"

Then I opened up that data set in the DE app ([link on localhost](http://localhost:3000/#/?attributeOptionComboSelection=eLwL77Z9E7R-STgy27fGvaU&dataSetId=pBOMPrpg1QX&orgUnitId=DiszpKrYNg8&periodId=202202)) and ran validation:
- I noticed that the `api/validation` response returns rule violations for the `default` attribute option combo, and not for the one currently selected (i.e., not for the "Target" selection in "Target vs. Result" in the link above)
- Editing the meningitis value for the "Target" attribute option combo to create a rule violation doesn't work (or at least it isn't reflected in the `validation` response

So I think this means that this request isn't correctly filtered by attribute option combo, and we either need to either 1) figure out how to _actually_ do that with this endpoint, 2) set up a new endpoint where it's possible, and make a note  (Jira issue?) to come back and implement this once that endpoint is ready

I realized that you might not have gotten a chance to learn what things like "attribute option combo" mean since you've been here, which can be a little complex -- in case you haven't, here's an attempt at a definition: "Attribute option combo" means the "unique ID representing the combination of selected options in each of the `category` parameters (which is configured by the 'category combination' that we modified in the maintenance app) that define the context of this data set". You will see "attribute combo" and "category combo" and other places where either "attribute" or "category" is used -- "attribute" means "category defined at the _data set_ level", to distinguish from the category combinations that are used for individual data elements. Maybe that's helpful? 😅

kabaros: thanks for the explanation - I will ask for a one-to-one walkthrough as it still confuses me though! :) ... I created a Jira ticket for investigating the attribute option combo and see whether we're doing something wrong or need a new endpoint: https://dhis2.atlassian.net/browse/TECH-1287

src/data-workspace/validation/use-validation-result.js
Mohammer5: Yes, so using the `validationQueryKey` doesn't make much sense here, it'd have to be the same query key that the `useDataValueSet` hook uses. I suppose you could simply use the `useDataValueSet`'s `isLoading` state here, it shouldn't trigger another fetch.

Maybe it makes more sense to disable the "Run validation" button instead though.

src/app/app.test.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react'
+import { QueryClient } from 'react-query'
 import { render } from '../test-utils/index.js'
 import App from './app.js'
 
 describe('<App />', () => {
     it('renders without crashing', () => {
-        render(<App />)
+        render(<App />, { queryClient: new QueryClient() })
```
Mohammer5, line 8: The `render` function has a default for the query client, right? Why did you pass a plain query client here?

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,125 @@
+import { useCallback } from 'react'
+import { useIsMutating, useQueries } from 'react-query'
+import {
+    useDataSetId,
+    useOrgUnitId,
+    usePeriodId,
+} from '../../context-selection/index.js'
+import { useApiAttributeParams } from '../../shared/index.js'
+
+const getValidationMetaDataQueryKey = (datasetId) => {
```
Mohammer5, line 10: `getValidationMetaDataQueryKey` as well as `getValidationQueryKey` should go into `src/data-workspace/query-key-factory.js`.

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,125 @@
...
+        },
+    ]
+
+    return queryKey
+}
+
+const getValidationQueryKey = (
+    datasetId,
+    { periodId, orgUnitId, categoryComboId, categoryOptionIds }
+) => {
```
Mohammer5, line 36: I wouldn't separate the data set id from the other options. This probably due to the data set it being a segment in the endpoint path while the others are query parameters, but these are implementation details and should affect the api of this function. Also I'd do `dataSetId` instead of `datasetId`:

```suggestion
const getValidationQueryKey = ({
    dataSetId,
    periodId,
    orgUnitId,
    categoryComboId,
    categoryOptionIds,
}) => {
```

src/data-workspace/validation/use-validation-result.js
```diff
@@ -0,0 +1,125 @@
...
+        isLoading: validationQuery.isLoading || metaDataQuery.isLoading,
+        isRefetching:
+            validationQuery.isRefetching || metaDataQuery.isRefetching,
+        data:
+            validationQuery.isSuccess && metaDataQuery.isSuccess
+                ? groupValidationRuleViolationsByImportance(
+                      validationQuery.data,
+                      metaDataQuery.data
+                  )
+                : undefined,
```
Mohammer5, line 100: We could use stale data while refetching. Not sure if that'd have an impact on the UI though, so just a though I have:

```suggestion
            validationQuery.data && metaDataQuery.data
                ? groupValidationRuleViolationsByImportance(
                      validationQuery.data,
                      metaDataQuery.data
                  )
                : undefined,
```

src/data-workspace/validation/validation-config.js
```diff
@@ -0,0 +1,34 @@
...
+    IconErrorFilled16,
+    IconWarningFilled16,
+    IconInfo16,
+    IconErrorFilled24,
+    IconWarningFilled24,
+    IconInfo24,
+} from '@dhis2/ui'
+
+export const validationLevels = ['HIGH', 'MEDIUM', 'LOW']
+export const validationLevelsConfig = {
```
Mohammer5, line 12: I think these should be components instead of config, that'd be more "react-ish".
Configs have the nasty issue that when there's an exception (like an edge case for a particular config), that exception has to be build into the general handling of the config rather than into the component that has to handle the exception.

Mohammer5: This requires a somewhat different approach how to compose things, e.g. instead of having a single validation priority group component, there would be three (`ValidationGroupHighPriority`, `ValidationGroupMediumPriority` & `ValidationGroupLowPriority`).

On the one hand this would create a bit more boiler plate code but on the other hand it's reduce complexity and increases maintainability as changing one of these composed components wouldn't affect the complexity of the others. Everything that's shared between them can be abstracted into its own component / function.

If you want to, we can talk about this in a call / on slack a bit more in case you disagree or have questions.

src/data-workspace/validation/validation-priority-group.js
```diff
@@ -0,0 +1,96 @@
...
+                <h1>
+                    {validationViolations.length === 1
+                        ? i18n.t(
+                              '{{length}} {{level}} priority alert',
+                              translationOptons
+                          )
+                        : i18n.t(
+                              '{{length}} {{level}} priority alerts',
+                              translationOptons
+                          )}
```
Mohammer5, line 70: If there were three different components for the various levels, there wouldn't be a need for a `translationOptons` (there's a typo btw: `Optons` -> `Options`).

The singular translation would look like this: `i18n.t('1 high priority alert')`. This would allow the translators to have a better understanding of the actual string, which they don't see in this apps context but just as an string outside the context of this app in the translation service we use.

src/data-workspace/validation/validation-result-prop-types.js
```diff
@@ -0,0 +1,28 @@
...
+        id: PropTypes.string.isRequired,
+        name: PropTypes.string.isRequired,
+    }),
+})
+
+export const ImportanceLevelPropTypes = PropTypes.oneOf([
+    'LOW',
+    'MEDIUM',
+    'HIGH',
+])
```
Mohammer5, line 28: Could (should?) be constants in a `constants.js` module as well

src/mocks/responses/validation-metadata.js
```diff
@@ -0,0 +1,333 @@
+export const validationMetadata = {
```
Mohammer5, line 1: This is only being used in the `validation-results-sidebar.test.js`, so I'd co-locate this with that file. Maybe `validation-metadata.response-dummy.js`. We don't have a convention for where to put this data, I don't even remember having seen test data outside of test files. I wouldn't mind putting this into the `.test.js` file actually, code editors can collapse that code anyway, so it shouldn't be in the way when browsing the test code.

Also I don't think that this is a "mock", although I'm sure there are different definitions for what term means what out there. Referring to [this stackoverflow answer](https://stackoverflow.com/a/17810004), the `validationMetadata` is "dummy" data, if I interpret the answer correctly.

Mohammer5 (changes requested):

src/data-workspace/validation/use-validation-result.js
kabaros: I  used `useIsMutating` with the key from `useDataValueSetQueryKey` to get the behaviour we're after. Otherwise if I use `useDataValueSet`, it disables the button when refetching the data element after it was updated (not while it's updating). Please have at [the change](https://github.com/dhis2/data-entry-app/pull/93/commits/f855db6f87e814625503ec55bfeba35ca77a6bc6) and let me know if this looks right.

src/data-workspace/validation/validation-config.js
kabaros: hmm I disagree with this .. I don't think having `ValidationGroupHighPriority` `ValidationGroupMediumPriority` etc.. would reduce complexity nor is it a more "react-y" way of achieving what we want. The `importance` (high, medium or low) is a property of the component, and react way is to use `props` for this. We have a similar concept in many places in our component library, so for things like `NoticeBox`, we wouldn't have an `ErrorNoticeBox` and `WarningNoticeBox`, it's just a prop. I don't think that having a config (or a helper function) is too complex to justify duplication. And if we try to avoid that duplication by abstracting it into its own component, then we're just shifting the complexity into that shared component (it also feels a bit inheritance-like rather componentised). 

For me it helps to look at the design and its language, [the design specs](https://docs.google.com/document/d/1vS1urvJRePRugFqrPkfOEjpkbij1AfTz0BYp6RKV8LU/edit#heading=h.9xf56tvc4qp8) talk about `PriortyGroupUnit` as a component with some `titles` and `icons` for each scenario. Design changes would happen at that component level, not a particular instance of it. That's the natural level of abstraction imo, everything else is a prop to the component.

Now maybe a config is not the best way to save that map of props .. but then we could just change the config to a helper function.

src/data-workspace/validation/use-validation-result.js
kabaros: in [the design specs](https://docs.google.com/document/d/1vS1urvJRePRugFqrPkfOEjpkbij1AfTz0BYp6RKV8LU/edit#heading=h.padhp6i77jlc), we're hiding data while refetching (I think that makes sense UX-wise)

src/app/app.test.js
kabaros: this is the equivalent of how the test was before .. with the default test query client (or the previous implementation with the [QueryClientWrapper](https://github.com/dhis2/data-entry-app/blob/development/src/test-utils/query-client-wrapper.js)) then this test passes but there is a console.error in the logs from trying to get the metadata. I can fix that by passing a CustomDataProvider, or by using `new QueryClient` like the `render` used to default to.

image

image

src/data-workspace/validation/validation-result-prop-types.js
kabaros: do you mean the PropTypes or the array passed to it?

src/mocks/responses/validation-metadata.js
kabaros: yeah that was the structure recommended by msw .. I will change it now that we got rid of it, and put it closer to the test

kabaros:
@Mohammer5  I think I tackled all your comments (pending reaching a decision about the component structure).

src/data-workspace/validation/validation-result-prop-types.js
Mohammer5: I mean something like this:

```js
const LOW = 'LOW'
const MEDIUM = 'MEDIUM'
const HIGH = 'HIGH'
```

I just double checked and there are only two instances where the actual string literal is being specified:
1. `validation-config.js`
2. `validation-result-prop-types.js`

I think instead of creating constants for the string literal, you could re-use the `validationLevels` variable in `validation-result-prop-types.js`

src/data-workspace/validation/use-validation-result.js
Mohammer5: :ok_hand:

src/data-workspace/validation/validation-config.js
Mohammer5: As discussed, we'll defer that, I need to prepare this thought in a proper manner so we can properly talk about it. This topic is highly subjective and depends on past experiences, so I'll try to show how I ended up with my position. You can leave the code as is

src/data-workspace/validation/use-validation-result.js
KaiVandivier: Perfect, thanks for the issue report 🙂

src/data-workspace/query-key-factory.js
```diff
@@ -40,3 +40,49 @@ export const dataValueContext = {
...
+    periodId,
+    orgUnitId,
+    categoryComboId,
+    categoryOptionIds,
+}) => {
+    const params = {
+        pe: periodId,
+        ou: orgUnitId,
+        cc: categoryComboId,
+        cp: categoryOptionIds?.join(';'),
```
KaiVandivier, line 78:
```suggestion
        // @TODO: Revisit these params to filter by AOC https://dhis2.atlassian.net/browse/TECH-1287
        cc: categoryComboId,
        cp: categoryOptionIds?.join(';'),
```

src/data-workspace/validation/validation-result-prop-types.js
kabaros: > I think instead of creating constants for the string literal, you could re-use the validationLevels variable in validation-result-prop-types.js

updated

src/data-workspace/validation/use-validation-result.js
KaiVandivier: Hm, I don't think disabling based on the `dataValueSets` fetch is correct -- I think we should disable it based on the `dataValues` mutations, which `use-data-value-set.js` seems to be lacking too

Based on the mutation filters docs linked above, I think this might look like `useIsMutating({ mutationKey: ['dataValues'] })`

KaiVandivier: Oh, that's... confusing 🤔

src/data-workspace/validation/validation-rule-expression.js
```diff
@@ -0,0 +1,34 @@
...
+        compulsory_pair: 'is compulsory with',
+        exclusive_pair: 'is exclusive to',
+    }
+    return (
+        <>
+            {leftSide.displayDescription} ({leftsideValue}){' '}
+            {operatorDisplay[operator] ?? operator}{' '}
+            {rightSide.displayDescription} ({rightsideValue})
+        </>
+    )
```
Mohammer5, line 28: Component can return strings, there's no need to wrap this within a `Fragment`.

src/data-workspace/validation/validation-rule-expression.js
```diff
@@ -0,0 +1,34 @@
...
+    // @ToDO: confirm the human readable equivalent for the operators
+    const operatorDisplay = {
+        equal_to: '==',
+        not_equal_to: '!=',
+        greater_than: '>',
+        greater_than_or_equal_to: '>=',
+        less_than: '<',
+        less_than_or_equal_to: '<=',
+        compulsory_pair: 'is compulsory with',
+        exclusive_pair: 'is exclusive to',
```
Mohammer5, line 20: Needs `i18n.t('...')`

src/data-workspace/validation/validation-rule-expression.js
```diff
@@ -0,0 +1,34 @@
...
+
+const ValidationRuleExpression = ({ validationRule }) => {
+    const {
+        leftsideValue,
+        rightsideValue,
+        metaData: { leftSide, rightSide, operator } = {},
+    } = validationRule
+
+    // @ToDO: confirm the human readable equivalent for the operators
+    const operatorDisplay = {
```
Mohammer5, line 12: Static object, can be moved out of component

Mohammer5 (changes requested):
This looks and works really well already! I just found a few minor things.

---

Currently the sidebar is closing when changing a field. You can temporarily turn that off by commenting out the line with `rightHandPanel.hide()` in `entry-field-input.js`. In the specs it says that when changing a value in the form, the following should be displayed:

image

---

There's too much spacing below the error message when an api error occurs:

image

src/data-workspace/validation/validation-rule-expression.js
kabaros: updated

src/data-workspace/validation/validation-rule-expression.js
kabaros: > Currently the sidebar is closing when changing a field. You can ....In the specs it says that when changing a value in the form, the following should be displayed

This is tackled in a different subtask / PR: https://dhis2.atlassian.net/browse/TECH-1275


> There's too much spacing below the error message when an api error occurs:

updated ... looks like this
image


> Component can return strings, there's no need to wrap this within a Fragment.
updated

src/data-workspace/validation/use-validation-result.js
kabaros: I created a subtask to look more into this: https://dhis2.atlassian.net/browse/TECH-1289 .. so just this PR can be unblocked and I can raise PRs for the next bits related to servr-side validation

KaiVandivier: Great, thanks 🙂👍

KaiVandivier (approved):
My comments have been met or added to future todos; looks good to me! 🙂

Mohammer5 (approved):

PR #96, opened 2022-07-13
refactor: move sortedcocs-computation to selector

Birkbjo:
Fixes https://dhis2.atlassian.net/browse/TECH-1112 by refactoring the computation to a `cachedSelector`. 

This also moves the `cartesian`-computation to a selector, and thus the options will be stable between components and remounts. So this will only be recomputed once for every `categoryCombo`. Even if that `catComboTable` is unmounted and remounted again.

### DIFF
--- a/src/metadata/selectors.js
+++ b/src/metadata/selectors.js
@@ -342,6 +342,7 @@ export const getCoCByCategoryOptions = createCachedSelector(
 /**
  * @param {*} metadata
  * @param {*} categoryComboId
+ * @returns {Array.<Array.<string>>} An array with arrays of categoryOption-ids
  */
 export const getComputedCategoryOptionsByCatComboId = createCachedSelector(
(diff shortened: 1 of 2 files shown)


src/metadata/selectors.js
```diff
@@ -309,6 +309,32 @@ export const getCoCByCategoryOptions = createCachedSelector(
     }
 )((_, categoryComboId) => categoryComboId)
 
+/**
+ * @param {*} metadata
+ * @param {*} categoryComboId
+ */
+export const getComputedCategoryOptionsByCatComboId = createCachedSelector(
```
Birkbjo, line 316: Naming is hard. But this computes all the combinations of `categoryOption`s in a `catCombo`, so the result is a list with lists of `categoryOptions`-ids.

Mohammer5 (approved):

PR #97, opened 2022-07-15
fix: address rerendering issues

Mohammer5:
This PR does multiple things:

* Prevent refetching the entire data values when changing values
   this is something I'd like to discuss next Tuesday, I already got different opinions on this from @HendrikThePendric and @ismay. This is something I'd like to discuss with the entire team
* Wraps some variables & children with `useMemo` to prevent unnecessary rerenders
* Wraps the `SectionFormSection` component with `React.memo(...)`
* Splits the right hand panel into multiple providers, just like we did with the current item provider
* Renames "current item" to "highlighted field" as we now create the highlighted item only in the areas where used (sidebar and bottom bar)
* Moves static props out of final form wrapper to prevent prop changes
* Moves `useDataValueSet` to `src/shared` as it's now used in multiple places
* Does NOT!!! address performance issues in custom forms, @HendrikThePendric wants to do that separately in one of his branches/PRs

Todos
* [ ] Address linting issues (will do before merging to prevent unnecessary work)


src/shared/highlighted-field/index.js
```diff
@@ -0,0 +1,4 @@
+export * from './highlighted-field-context.js'
+export { default as HighlightedFieldIdsProvider } from './highlighted-field-ids-provider.js'
+export * from './use-highlighted-field-context.js'
+export { default as useCreateHighlightedFieldIds } from './use-highlighted-field.js'
```
Birkbjo, line 4: Why is this name different from the `default` function name? `useHighlightedFieldId`?

src/shared/highlighted-field/highlighted-field-context.js
```diff
@@ -1,12 +1,12 @@
 import { createContext } from 'react'
 
-export const CurrentItemContext = createContext({
+export const HighlightedFieldIdsContext = createContext({
```
Birkbjo, line 3: Why is `ids` plural? From the name it seems like this contain multiple `ids`.

src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -45,24 +45,25 @@ function mapDataValuesToFormInitialValues(dataValues) {
     return formInitialValues
 }
 
+const select = (data) => {
```
Birkbjo, line 48: It might be valuable to memoize this function, eg. with `reselect`. This is run on every render.

src/right-hand-panel/right-hand-panel-provider.js
```diff
@@ -10,7 +14,11 @@ export default function RightHandPanelProvider({ children }) {
 
     return (
         <RightHandPanelContext.Provider value={value}>
-            {children}
+            <ShowRightHandPanelContext.Provider value={show}>
```
Birkbjo, line 17: As far as I can see, it's not necessary to split `show` and `hide` into multiple contexts. These do not change, and can safely be in the same context. They also seem to be used together in the same component. You would probably need to `useMemo` for the context-value though. 


```javascript
// in RightHandPanelProvider
const setters = useMemo(() => {
    const show = setId
    const hide = () => setId('')
   return { show, hide } 
}, [setId])

<SetRightHandPanelContext.Provider value={setters}>


const useSetRightHandPanel = useContext(SetRightHandPanelContext)

// usage
const { show, hide } = useSetRightHandPanel()
```

It's only worth splitting the context from values that change often, and not together.

Birkbjo (changes requested):

src/data-workspace/section-form/section.js
```diff
@@ -9,18 +9,18 @@ import {
...
-import React, { useState, useMemo } from 'react'
+import React, { useMemo, useState } from 'react'
 import { useMetadata, selectors } from '../../metadata/index.js'
 import { CategoryComboTableBody } from '../category-combo-table-body/index.js'
 import { getFieldId } from '../get-field-id.js'
 import { IndicatorsTableBody } from '../indicators-table-body/indicators-table-body.js'
 import styles from './section.module.css'
 
-export const SectionFormSection = ({
+export const SectionFormSection = React.memo(function SectionFormSection({
```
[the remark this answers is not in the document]
Birkbjo: I'm pretty sure this is due to the parent re-rendering, and the parent causing it is the `FinalFormWrapper` in `DataWorkspace`. I think we should remove this `React.memo` to the `EntryForm`-component. That way your re-rendering optimization should benefit all form-types.

src/data-workspace/section-form/section.js
```diff
@@ -46,18 +47,48 @@ export const SectionFormSection = ({
...
+            section.showRowTotals,
+            section.showColumnTotals,
+            groupedDataElements,
+            filterText,
+            globalFilterText,
+            maxColumnsInSection,
+            greyedFields,
+        ]
+    )
+
```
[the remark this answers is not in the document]
Birkbjo: What kind of input did you do to observe the `children` change? 
I'm pretty sure this is not necessary - since react will use the `key`-prop and see that no dom-changes are necessary. The `children`-prop may change in the profiler, but I don't think this should matter. 
I would rather use `React.memo` on the `CategoryComboTableBody` if this kind of optimization is necessary.

src/metadata/selectors.js
```diff
@@ -74,8 +74,10 @@ export const getCategoryOptionsByCategoryOptionComboId = createCachedSelector(
         const categoryCombos = Object.values(metadata.categoryCombos)
```
Birkbjo, line 74:
```suggestion
        const categoryCombos = Object.values(getCategoryCombos(metadata))
```

src/metadata/selectors.js
```diff
@@ -74,8 +74,10 @@ export const getCategoryOptionsByCategoryOptionComboId = createCachedSelector(
...
 
         for (const categoryCombo of categoryCombos) {
-            for (const curCategoryOptionCombo of categoryCombo.categoryOptionCombos) {
-                if (curCategoryOptionCombo.id === categoryOptionComboId) {
+            const curCategoryOptionCombos =
+                categoryCombo?.categoryOptionCombos || []
```
Birkbjo, line 78: How come this `?.` is necessary? I would not expect `categoryCombos` to contain a null or undefined value? 
It's a minor thing, I'm just curious.

src/shared/use-data-value-set/use-data-value-set.js
Mohammer5: This will happen anyway. A change will optimistically update the response, which causes `data` to change, which in turn will trigger the `select` function.

src/shared/highlighted-field/highlighted-field-context.js
Mohammer5: Because a highlighted field's "id" is essentially the combination of the data element's id as well as the coc's id

src/metadata/selectors.js
Mohammer5: I don't remember to be honest.. I just encountered this once, but that might be due to my trying things out or due to some hot reloading weirdness

src/shared/highlighted-field/highlighted-field-context.js
Birkbjo: I see where you're coming from, but we're referring to that concatenation as the `fieldId`. Check `get-field-id.js`, and functon `getFieldId()`. So I would still argue for this to be called `fieldId`.

src/shared/use-data-value-set/use-data-value-set.js
Birkbjo: That's not what I mean. The select function is re-run even though the data for the query doesn't change. For every component that is mounted and uses the `useDataValueSet`, or re-renders, the select function is run.

src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -45,24 +45,25 @@ function mapDataValuesToFormInitialValues(dataValues) {
...
+    const dataValues = mapDataValuesToFormInitialValues(data.dataValues)
+    const minMaxValues = data.minMaxValues || {}
+    return { dataValues, minMaxValues }
+}
```
Birkbjo, line 52:
```suggestion
const select = createSelector(
    (data) => data,
    (data) => {
        const dataValues = mapDataValuesToFormInitialValues(data.dataValues)
        const minMaxValues = data.minMaxValues || {}
        return { dataValues, minMaxValues }
    }
)
```

Mohammer5: According to the [documentation](https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option), my solution should work:

> Selectors like the one above will also run on every render, because the functional identity changes (it's an inline function)
> If your transformation is expensive, you can memoize it either with useCallback, or by extracting it to a stable function reference

I moved it outside of the component, giving it a stable reference

src/shared/use-data-value-set/use-data-value-set.js
Mohammer5: See the comment below your suggestion. Let's keep talking there, will resolve this one

src/shared/use-data-value-set/use-data-value-set.js
Mohammer5: That hook is only being used in two places (`data-workspace.js` & `use-highlighted-field.js`), the fields don't use it anymore. So I'm wondering what's more expensive:
* Re-computing this when the data values change
* checking if the data has changed on every render

I suppose the re-computing as an operation itself is more expensive, but checking for a change happens more often, so it could be more expensive the accumulative time it take on every render. Unless this is a real performance issue, I wouldn't use reselect here. The data structure of this request seems simple to me, unlike the metadata select callback for example.

I don't mind if we profile this though, if you're worried about this!

src/shared/highlighted-field/highlighted-field-context.js
Mohammer5: Ah, fair enough, will change it

src/shared/use-data-value-set/use-data-value-set.js
Birkbjo: > I moved it outside of the component, giving it a stable reference. That means it won't run the select function on re-renders.

It's not enough, just test to print something in the selector. It's atleast run for every mount.

We are doing this for other selectors, like in use-metadata.


> checking if the data has changed on every render

It's a simple reference check, so I wouldn't worry about this at all.

Birkbjo: But yeah I tested it some and it doesn't seem like it is recomputed that much, but it is atleast computed 3 times in app load.

Mohammer5: Alright, I've wrapped it

Birkbjo (approved):

PR #99, opened 2022-07-21
refactor(app wrapper): improve jest test utils

Mohammer5:
I'm working on updating tests in #73. For this I need the components to be wrapped with the providers that wrapp the app component. @kabaros also had a similar need and ended up with a [`withAppTestWrapper`](https://github.com/dhis2/data-entry-app/blob/19a73e3637/src/test-utils/with-app-test-wrapper.js) HOC.

I already made a comment that I think I'd prefer the [official way](https://testing-library.com/docs/react-testing-library/setup/#custom-render) of doing this, so I created this PR which
* should support @kabaros's case (overriding the Router with `MemoryRouter`)
* exports the `TestWrapper` component as an escape hatch in case tests need to do something special
* updates all existing tests, showing that the changes just work

See [`src/app/app.test.js`](https://github.com/dhis2/data-entry-app/pull/99/files#diff-e7cb7ff6e3b917804d472df26e224c232bea4a09ad1f60500da1757a588a5e49) for how this removes test boilder plate code

### DIFF
--- a/src/app/app-wrapper.js
+++ b/src/app/app-wrapper.js
@@ -1,7 +1,8 @@
 import { CssReset, CssVariables } from '@dhis2/ui'
+import PropTypes from 'prop-types'
 import React from 'react'
-import { QueryClientProvider } from 'react-query'
-import { HashRouter as Router, Route } from 'react-router-dom'
+import { QueryClient, QueryClientProvider } from 'react-query'
+import { HashRouter, Route } from 'react-router-dom'
 import { QueryParamProvider } from 'use-query-params'
 import PrintAreaProvider from '../data-workspace/print-area/print-area-provider.js'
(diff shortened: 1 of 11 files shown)


src/test-utils/@testing-library/react/index.js
```diff
@@ -0,0 +1,45 @@
...
+            <OuterComponents queryClient={queryClient} router={router}>
+                {children}
+            </OuterComponents>
+        )
+
+        if (!WrapperFromOptions) {
+            return content
+        }
+
+        return <WrapperFromOptions>{content}</WrapperFromOptions>
```
Mohammer5, line 27: `WrapperFromOptions -> OuterComponents -> children` or `OuterComponents -> WrapperFromOptions -> children`?

PR #100, opened 2022-07-25
fix(context selection): implement improvements TECH-1238

Mohammer5:
Closes [TECH-1238](https://dhis2.atlassian.net/browse/TECH-1238)
Closes subtask [TECH-1239](https://dhis2.atlassian.net/browse/TECH-1239)
Closes subtask [TECH-1240](https://dhis2.atlassian.net/browse/TECH-1240)
Closes subtask [TECH-1241](https://dhis2.atlassian.net/browse/TECH-1241)
Closes subtask [TECH-1242](https://dhis2.atlassian.net/browse/TECH-1242)
Closes subtask [TECH-1248](https://dhis2.atlassian.net/browse/TECH-1248)
(TECH-1243 seems to have been implemented already, so marked it as done with "Won't do" in Jira)

Commit messages say exactly what's been done in this PR


src/context-selection/menu-select/menu-select.module.css
```diff
@@ -1,6 +1,6 @@
 .menuSelect {
     width: 400px;
-    max-height: 100vh;
+    max-height: 70vh;
```
tomzemp, line 3: This could be 80vh to match the max-height that is used in org-unit-selector-bar-item.module.css (or they could both be set to 70vh)

src/context-selection/attribute-option-combo-selector-bar-item/use-should-component-render-null.js
```diff
@@ -13,6 +13,7 @@ export default function useShouldComponentRenderNull(
     const [orgUnitId] = useOrgUnitId()
 
     return (
+        categoryCombination?.isDefault ||
```
tomzemp, line 16: check in lines 23-24:  `(!categoryCombination.isDefault && !categoryCombination.categories.length)` seems like it can now be simplified to `!categoryCombination.categories.length`

tomzemp (commented):
@Mohammer5: looks good. I had some questions about the updates on the period selector and some minor comments on the other issues

**TECH-1239**
looks good. I'd suggest making the max-height for the menuSelect class the same as for the OrgUnit selector (itemContentContainer class), i.e. 80vh (or make them both 70vh)

**TECH-1240**
looks good. I think the logic for the check in useShouldComponentRenderNull can now be simplified a bit.

**TECH-1241**
looks good. Maybe a design question for Joe, but if the attribute option combo name is really long, we might want to truncate it?

**TECH-1238**
I have a couple of comments (but apologies if I'm misunderstanding the ticket)

- current logic for calculating future periods only looks at the next year. I think this needs to be redone as while loop or similar (e.g. so if dataset is monthly and there are 30 openFuturePeriods, the future periods might span the next 3 years)
- there does not appear to be any logic for adding future years if dataset is yearly and there are openFuturePeriods
- the logic for determining acceptable future periods now seems (to me) to be split between useFuturePeriods (in use-future-periods) and the createFilterFuturePeriods (in period-menu), which seems less optimal for code maintenance over time. 
Since the filterFuturePeriods is only relevant for the current year at this point, it seems like useFuturePeriods could be modified to return currentYearAndFuturePeriods; that way, the previous years using getPeriods could just not run the filter. Alternately, maybe it would make sense to focus on determining the last possible endDate (rather than all future periods), so that this could be passed to a filter function definition and the logic there can be simplified to be: is the period between the earliest possible start date and the last possible end date
- future/current periods are being determined based on browser time, rather than server time (very edge case issue, but just pointing that out in case we want to address)
- I'm assuming that there are other tickets for the complexities of dataInputPeriods and openPeriodsAfterCoEndDate, such that we don't need to consider their impact at the moment


**TECH-1248**
looks good. one disadvantage, however, is that the getYearlyPeriodType function in fixed-periods is only returning the last 10 years, so with the correct logic you've implement, a user cannot go more than 10 years back in time. Unless there was some UX decision about this, maybe this function can just be modified the same startYear constant of 1970 that's in year-navigator.js?

Mohammer5:
Thanks @tomzemp 

> TECH-1239
> looks good. I'd suggest making the max-height for the menuSelect class the same as for the OrgUnit selector (itemContentContainer class), i.e. 80vh (or make them both 70vh)

Done

> TECH-1240
> looks good. I think the logic for the check in useShouldComponentRenderNull can now be simplified a bit.

Done

> TECH-1241
> looks good. Maybe a design question for Joe, but if the attribute option combo name is really long, we might want to truncate it?

Yes, this is something @cooper-joe should decide. We could do that in an extra PR so this one isn't blocked. Could you create a Jira issue and assign it to epic [TECH-1187](https://dhis2.atlassian.net/browse/TECH-1187)?

> * current logic for calculating future periods only looks at the next year. I think this needs to be redone as while loop or similar (e.g. so if dataset is monthly and there are 30 openFuturePeriods, the future periods might span the next 3 years)
> * there does not appear to be any logic for adding future years if dataset is yearly and there are openFuturePeriods

Good catch, will implement that now!

> the logic for determining acceptable future periods now seems (to me) to be [...] less optimal for code maintenance over time.

Will take a look at this one, too

> only returning the last 10 years

I'm wondering about what's the best solution here. Maybe it makes sense to simply show all possible years. It's the quickest solution right now. I'll just ahead with that implementation right now. @cooper-joe if you prefer to do this differently, please create a jira Issue and add it either to the epic for the next release ([TECH-1152](https://dhis2.atlassian.net/browse/TECH-1152)) or to the one for the following release ([TECH-1187](https://dhis2.atlassian.net/browse/TECH-1152))

---

@tomzemp In case I don't manage to get this PR merged before I leave, do you mind taking it over as you've already spent some time digging through the code & changes this PR should add?

Mohammer5:
@tomzemp I think I've addressed all your points. The user can now select any year when the type is yearly and all future periods should be covered. I also tried to untie couple code a bit, but it still feels like we could improve the code quite a bit. Handing over the PR to you as I'll be on vacation

src/context-selection/period-selector-bar-item/use-periods.js
```diff
@@ -0,0 +1,79 @@
...
+            periods = Array.from(Array(yearsCount)).map((_, index) => {
+                const year = index + 1970
+                const periodId = getYearlyPeriodIdForTypeAndYear(
+                    periodType,
+                    year
+                )
+                return parsePeriodId(periodId)
+            })
+        } else if (year > currentYear) {
+            return futurePeriods.filter(
```
tomzemp, line 56: this should be `periods = futurePeriods.filter(//etc)` as currently the future periods are not being reversed (whereas the current/past periods are reversed)

tomzemp:
This looked good. I liked the hooks for controlling the period logic, and it seems like there is not duplicated logic of testing for future period validity! I looked into why the cypress test was failing in CI (and not locally) which then led me to discover some issues (see below), and then I tested on the other periodTypes and found a few more things to clean up. Changes I've made are summarized below.

## Cypress test failure (January 2022 appearing in 2021 months)

**TL;DR:** this is fixed by not using moment in the app. I've removed moment and made some other small updates to account for JavaScript's worse handling of date addition/subtraction.

The cypress test for number of months was failing in CI because of a difference in date instantion between JavaScripts native Date constructor and moment.js. 

FYI: moment appears to instantiate a date from a string in the local timezone, so moment.date('2022-01-31') would be `2022-01-31 00:00:00 (CET)` whereas JavaScript `Date()` is treating the dateString as UTC time, so `Date('2022-01-31') ` would be `2022-01-31 01:00:00 (CET)`.

The test then failed because we had logic to account for the possibility of dates spanning years (so we looked up through the first startDate in the new year). When running locally in CET time zone, we'd check that the January 2022 period was valid by checking that the period's end date was less than our maximum date and it did not because of where we were using moment/`Date()`
`2022-01-31 01:00:00 (CET)` <= 2022-01-31 00:00:00 (CET)`

However, when the test was executed in CI mode, it seems to run in UTC and the test ends up as
`2022-01-31 00:00:00 (UTC) <= 2022-01-31 00:00:00 (UTC)`, which passes and then we end up with January 2022 showing up in our list of 2021 months.

## Period end date
Throughout fixed-periods, we record the endDate for a period as the day before the subsequent startDate, e.g. for August 2022 month, the endDate is 2022-08-31. In the logic in getFixedPeriodsForTypeAndDateRange, we instantiated a date with Date(endDate) which means we could have a user log in on 2022-08-31 at 10am and be shown August 2022. I've updated the logic to make the endDate be end of day, so that users could not, for example, see August 2022 until the period actually ends and they log on in September 2022 or later.

_[This would be something to consider for Approvals app as well]_


## Future period calculations
I like the approach of iteratively building up an array of future periods (getFollowingFixedPeriodsForPeriod), as I think that's the surest way to get the exact number of periods needed. However, since this used called to getFixedPeriodsForTypeAndDateRange which in turn called getFixedPeriodsByTypeAndYear, we could end up in a situation where we evaluated many periods repeatedly. For example, when iteratively getting openFuturePeriods for daily periodType, we seem to be getting a year's worth of days and then using a forEach to evaluate each of those periods, so e.g. if we need 500 future daily periods, we do something like 365*500 evaluations. 

Because of that, I've just switched to trying to find the startDate of the period after the final open future period and then calling getFixedPeriodsForTypeAndDateRange with that set as the end date. This should be more performant and worked in all tests I tried. After making that change, it was no longer necessary to have a separate calculation of futurePeriods, so in my last commit I changed the logic to remove the separate useFuturePeriods hook; if we wanted to keep this differentiation, I could revert the last commit. 


## Additional cypress tests
I added in some additional tests to cover the other periodTypes available in the Sierra Leone database.


## Yearly period types
Year selector was being hidden only for `periodType=='yearly'`. I've hidden for the other yearly period types (e.g. FinancialYearOct)

## Periods accross years
The future periods had a slightly different logic for presenting periods that crossed over year boundaries. I've standardized to make all years display periods that cross a given year (e.g. SixMonthlyApril for 2022 would show October 2021-March 2022, April 2022 - September 2022, October 2022 - March 2023)

## Six-monthly November
This period was missing from fixed-periods.js. I've added it in as the backend allows datasets to be created with this periodType.
_[Something to consider for other uses of this code. Also the developer documentation does not include all periods]_

## Miscellaneous cleanup

- some of the updates to fixed-periods.js resulted in some errors (e.g. like const period = {startDate: '2022', iso: period.startDate})
- there were some inconsistencies in period definition, e.g. correcting 2022October to 2022Oct for FinancialYearOct periodTypes
- I removed some calls to parsePeriodId() where the function was just being called to add in the periodType (and cleaned up corresponding unit tests)


## Issues outside the scope of this PR

- Backend seems to allow for negative values on openFuturePeriods (it's set as an Integer). In the last commit, I enforced logic to ensure we had only "future" periods and didn't get disabled "past" periods. The user docmentation does not mention any possibility of negative values, so I believe this is just a bug.
- biweekly data periods in past years sometimes lead to situations where certain dates are inaccessible, e.g. 2020 Bi-Week 26 is 2020-12-14 to 2020-12-27 and 2021 Bi-Week 1 iss 2021-01-04 to 2021-01-17 which seems to imply there is a week where data entry is not possible?

HendrikThePendric (approved):
Thanks @tomzemp for such a thorough and well documented PR. I've ran the branch locally and everything is working as expected. Your changes make a lot of sense as well. So as far as I am concerned this is good to merge.

You rightfully pointed out that some other apps would need these changes too. While I fully agree with this, I think we can defer this for now. There is an ongoing discussion on how we should deal with period logic. It is being used on the server, in the android SDK and in the browser. Ideally we'd like to have one single re-usable implementation, but this is technically challenging. If this turns out to be impossible and we are to implement a browser specific solution, we should create a dedicated repo/package for this and the code in the data entry app could be the starting point for that.

> Backend seems to allow for negative values on openFuturePeriods (it's set as an Integer). In the last commit, I enforced logic to ensure we had only "future" periods and didn't get disabled "past" periods. The user docmentation does not mention any possibility of negative values, so I believe this is just a bug.

Would you mind creating a JIRA ticket for this?

> biweekly data periods in past years sometimes lead to situations where certain dates are inaccessible, e.g. 2020 Bi-Week 26 is 2020-12-14 to 2020-12-27 and 2021 Bi-Week 1 is 2021-01-04 to 2021-01-17 which seems to imply there is a week where data entry is not possible?
I've added this to the agenda of tomorrow's team-platform meeting.

cypress/integration/context-selection/period/index.js
```diff
@@ -140,11 +163,39 @@ Then('the user should only see options of that type for "monthly"', () => {
 })
 
 Then('the user should only see options of that type for "yearly"', () => {
+    const currentYear = new Date().getFullYear()
+    const yearsToBeDisplayed = currentYear - 1970 // year of 0-timestamp
```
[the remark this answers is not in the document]
tomzemp: I tried importing this with `import { startYear } from '../../../../src/shared/index.js'` (which seems to be the right path as I checked it a couple times)
but, I get an error that I cannot figure out:
```
SyntaxError: Unexpected token
    at EventEmitter.handler (/Users/thomasrz/Library/Caches/Cypress/9.7.0/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/util.js:69:27)
    at EventEmitter.emit (node:events:390:28)
    at ChildProcess.<anonymous> (/Users/thomasrz/Library/Caches/Cypress/9.7.0/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/util.js:19:22)
    at ChildProcess.emit (node:events:390:28)
    at emit (node:internal/child_process:917:12)
    at processTicksAndRejections (node:internal/process/task_queues:84:21)
``` 
I haven't had any luck figuring this out. @kabaros: do you have any ideas?

cypress/integration/context-selection/period/index.js
```diff
@@ -151,3 +174,28 @@ Then('the user should only see options of that type for "yearly"', () => {
...
+        cy.get('[data-test="period-selector-menu"] li').should(
+            'have.length',
+            yearsToBeDisplayed
+        )
+    }
+)
+
+// count varies by year; for 2021: 53 weeks
+Then('the user should only see options of that type for "weekly"', () => {
+    cy.get('[data-test="period-selector-menu"] li').should('have.length', 53)
```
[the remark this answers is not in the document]
tomzemp: The weekly period selection has been implemented with `periodId: 2021W52` as a constant to avoid this issue. I think that's probably okay rather than having a dynamic test that updates yearly? I think there's not a way to get around the yearly periodTypes being dynamic.

src/context-selection/attribute-option-combo-selector-bar-item/categories-menu.js
```diff
@@ -10,6 +11,29 @@ export default function CategoriesMenu({
     selected,
     onChange,
 }) {
+    if (categories.length === 1) {
+        const category = categories[0]
+        const values = category.categoryOptions.map(({ id, displayName }) => ({
```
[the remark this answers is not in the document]
tomzemp: Jan-Gerke wrote this logic, so I can't speak for him directly, but I think it's okay because every categoryCombo will have at least 1 category, and every category will have at least 1 categoryOption. So I don't think it's likely to end up in a situation where we need to be defensive (it could happen if they screw around in their database...but then maybe it's better to just have the app fail at that point rather than to add handling for that?)

src/context-selection/period-selector-bar-item/use-date-limit.test.js
```diff
@@ -0,0 +1,120 @@
...
+ * Should add future periods, even when spanning over several years
+ */
+describe('computeDateLimit', () => {
+    afterEach(() => {
+        getCurrentDate.mockClear()
+    })
+
+    it('should be current date if no openFuturePeriods', () => {
+        getCurrentDate.mockImplementation(() => {
+            return new Date()
```
tomzemp, line 19: I noticed in running these tests that they are a bit flakey with this implementation; the actual and expected get executed at slightly different times, so sometimes on my machine they fail the equality check as they differ by 1 ms. So, I think this test should be modified to return a constant time

tomzemp: That's neat. Jan-Gerke actually added the getCurrentDate function with a note about not using it once jest.setSystemTime was working, so I wonder if he was running into problems with that before? I'm going to just return a constant from the mocked getCurrentDate for now, but will make a note to follow up if we should get rid of this function and just use jest.setSystemTime

PR #101, opened 2022-07-28
fix: untie circular dependencies

Mohammer5:
I've encountered some nasty bugs while trying to fix some tests which were due to circular dependencies.
This meant that I had to move some things into the `shared` folder, which are actually "shared" things, like metadata.
The downside of this is that it's not fully "ducks"-mode anymore.

The folders on the top level (context-selection, data-workspace) are now for the respective UI area only, the `use-context-selection` is now part of the `shared` folder. Right now I believe that this is actually a good decision, but I can believe that this could start quite a controversy and maybe I'm overlooking something that we really shouldn't do but I did anyway in this PR.
So instead of adding a commit to #73, I made this a separate PR so we can talk about it


kabaros (approved):
lgtm .. I don't understand why it's a controversial change though (making the change to get rid of circular references is a good reason for me) so maybe someone else with more experience with the project should check too

kabaros:
is `madge` giving zero circular references now?

Mohammer5:
> I don't understand why it's a controversial change though

The reason is that I've moved a couple of things into the shared folder. Some of these violate the "modules"/"ducks" approach. The `useContextSelection` hooks are now decoupled from the `src/context-selection` folder, which I think is good actually, but I can see disagreement on that part and wanted to see if there are alternative solutions other have in mind.

> is madge giving zero circular references now?

image

PR #105, opened 2022-08-02
fix(validation): tweaks for validation right hand panel

kabaros:
## Changes in this PR

- When the right hand panel is open, and the "Run Validation" button in the bottom bar is clicked, then the the panel is toggled closed
- Ensure that the right panel is hidden when the dataset changes
- Get rid of `0` that appeared in comments validation section when no violations exist

## Screenshots

**Closing right hand panel when dataset is changed**
| Before | After
| --- | --- |
|  close_right_hand_on_change_dataset_before | close_right_hand_on_change_dataset_after |

**Clicking Run validation in the bottom menu again closes an already open menu**
| Before | After
| --- | --- |
| toggle_right_hand_before |  toggle_right_hand_after |

### DIFF
--- a/src/data-workspace/use-close-right-hand-panel-on-selection-change.js
+++ b/src/data-workspace/use-close-right-hand-panel-on-selection-change.js
@@ -1,18 +1,20 @@
 import { useEffect } from 'react'
-import { useParams } from 'react-router-dom'
+import {
+    useAttributeOptionComboSelection,
+    useDataSetId,
+    useOrgUnitId,
+    usePeriodId,
+} from '../context-selection/index.js'
 import { useRightHandPanelContext } from '../right-hand-panel/index.js'
 
 export default function useCloseRightHandPanelOnSelectionChange() {
-    const { show } = useRightHandPanelContext()
-    const searchParams = useParams()
+    const { hide } = useRightHandPanelContext()
+    const [dataSetId] = useDataSetId()
+    const [periodId] = usePeriodId()
+    const [orgUnitId] = useOrgUnitId()
+    const [attributeOptionComboSelection] = useAttributeOptionComboSelection()
 
-    useEffect(
-        () => {
-            show('')
-        },
-        // search params are part of the dependency array as we need to hide
-        // the sidebar when the selection changes. There's no need to use the
-        // search params in the useEffect though
-        [searchParams, show]
-    )
+    useEffect(() => {
+        hide()
+    }, [dataSetId, periodId, orgUnitId, attributeOptionComboSelection, hide])
 }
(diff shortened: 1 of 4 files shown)


src/data-workspace/data-workspace.js
```diff
@@ -14,12 +14,14 @@ import { useDataValueSet } from '../shared/index.js'
...
 import { EntryForm } from './entry-form.js'
 import { FinalFormWrapper } from './final-form-wrapper.js'
+import useCloseRightHandPanelOnSelectionChange from './use-close-right-hand-panel-on-selection-change.js'
 
 export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
     const [{ dataSetId }] = useContextSelection()
     const { data } = useMetadata()
     const initialDataValuesFetch = useDataValueSet()
     const isValidSelection = useIsValidSelection()
+    useCloseRightHandPanelOnSelectionChange()
```
kabaros, line 24: this hook needed to be a level higher in the hierarchy otherwise it errors and doesn't close if the new selection is invalid

src/data-workspace/use-close-right-hand-panel-on-selection-change.js
```diff
@@ -3,16 +3,16 @@ import { useParams } from 'react-router-dom'
...
     useEffect(
         () => {
-            show('')
+            hide()
         },
         // search params are part of the dependency array as we need to hide
         // the sidebar when the selection changes. There's no need to use the
         // search params in the useEffect though
-        [searchParams, show]
+        [searchParams, hide]
```
Mohammer5, line 16: I realize this hook was using `useParams`. Other hooks do this differently, see `src/context-selection/section-filter-selector-bar-item/use-on-dependent-params-change.js`. This way the user can change the section filter without closing the right hand panel. Plus we can add any query param without having to worry about the sidebar being closed

src/data-workspace/data-workspace.js
Mohammer5: Wondering if it'd make more sense to put this into the component in `right-hand-panel.js`. Just a thought, not a proposal

Mohammer5 (changes requested):
Looks good, just 1 change request

src/data-workspace/data-workspace.js
kabaros: > Wondering if it'd make more sense to put this into the component in `right-hand-panel.js`. Just a thought, not a proposal

yeah that makes sense .. moved it to `right-hand-panel.js`

src/data-workspace/use-close-right-hand-panel-on-selection-change.js
kabaros: updated

Mohammer5 (approved):

PR #106, opened 2022-08-03
feat: update validation comments violations UI and pair formula texts

kabaros:
implements [TECH-1292](https://dhis2.atlassian.net/browse/TECH-1292)

## Changes in this PR
- Update design for required comments to match updated design specs
- Update formula text for special cases (compulsory_pair and exclusive_pair)
- Update design slightly to make the formula's right and left side values italic

# Screenshots

image


src/data-workspace/validation/validation-comments-violations.js
```diff
@@ -1,27 +1,47 @@
...
+
+    if (!commentsCount) {
+        return null
+    }
+
+    const title =
+        commentRequiredViolations?.length === 1
+            ? i18n.t('{{commentsCount}} comment required', {
+                  commentsCount,
+              })
```
Mohammer5, line 18:
```suggestion
            ? i18n.t('1 comment required')
```

src/data-workspace/validation/validation-rule-expression.js
```diff
@@ -20,9 +10,45 @@ const ValidationRuleExpression = ({ validationRule }) => {
...
+        ),
+    }
+
+    if (operator === 'compulsory_pair' || operator === 'exclusive_pair') {
+        return (
+            <span className={styles.formulaValue}>
+                {operatorDisplay[operator]}
+            </span>
+        )
+    }
```
Mohammer5, line 39: Just a suggestion: There are already conditions checking whether it's one of these "special" operators, so we could keep the static ones outside of this component to keep the component itself smaller / more readable. I wouldn't mind having a component for the "pair" operators, not sure what a good name would be though (`<ConditionalRelationshipOperator />` sounds awkward and is quite long...).

Mohammer5 (approved):

src/data-workspace/validation/validation-rule-expression.js
kabaros: updated .. not sure about the naming either, but I think regardless of the operator, one of them is an expression with right and left-side values, while the other one is a kind of static description. Have [a look](https://github.com/dhis2/data-entry-app/compare/42cbfb42774908ec0c4c1d5ad338ca962f1c07d0..cf9798cacf002dfdcaf29040ef643a40f6dce9a0) and let me know what you think.

cooper-joe (approved):

Mohammer5 (approved):

PR #108, opened 2022-08-08
fix(validation): disable running validation when a mutation is ongoing

kabaros:
implements [TECH-1289](https://dhis2.atlassian.net/browse/TECH-1289)

Changes in this PR 
---

Disable running and re-running validation when there is an ongoing mutation.

In this PR, I am using [useDataValueSetQueryKey](https://github.com/dhis2/data-entry-app/blob/development/src/shared/use-data-value-set/use-data-value-set-query-key.js#L5) as the mutation key, which is used internally by [useDataValueMutation](https://github.com/dhis2/data-entry-app/blob/e327a277ff754c9de3167115fb1084143fa324bc/src/data-workspace/use-data-value-mutation/use-data-value-mutation.js#L11). [This other bit](https://github.com/dhis2/data-entry-app/blob/e327a277ff754c9de3167115fb1084143fa324bc/src/bottom-bar/main-tool-bar.js#L13) that is related and unchanged.

**Note**: This PR is a continuation of [this discussion](https://github.com/dhis2/data-entry-app/pull/93#discussion_r930139503).

### DIFF
--- a/src/data-workspace/validation/use-validation-result.js
+++ b/src/data-workspace/validation/use-validation-result.js
@@ -1,5 +1,5 @@
 import { useCallback } from 'react'
-import { useIsMutating, useQueries } from 'react-query'
+import { useQueries } from 'react-query'
 import {
     useApiAttributeParams,
     useDataSetId,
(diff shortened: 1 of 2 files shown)


KaiVandivier (approved):
Looks good! Tested successfully 🙂

Once I refactor the data value mutations, I'll revisit this to maybe use that mutation key instead of the `dataValueSetQueryKey` 👍

PR #109, opened 2022-08-10
chore: use logical CSS properties instead of physical properties throughout the app

kabaros:
implements [TECH-1294](https://dhis2.atlassian.net/browse/TECH-1294)

### Changes in this PR

Use logical CSS properties instead of physical properties to make localisation to right-to-left (and vertical) languages easier. 

#### Background and useful links

- [Mapping for margins and borders](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Margins_borders_padding#mappings_for_margins_borders_and_padding)

- [Mapping for floating and positions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Floating_and_positioning)

- Browser support - https://caniuse.com/css-logical-props

- [Why we need logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Basic_concepts#why_do_we_need_logical_properties)

#### Note

A followup PR will enforce the use of logical operators in CI checks

### DIFF
--- a/src/app/load-app.module.css
+++ b/src/app/load-app.module.css
@@ -1,3 +1,3 @@
 .center {
-    margin-top: var(--spacers-dp48)
+    margin-block-start: var(--spacers-dp48)
 }
--- a/src/bottom-bar/data-item-bar.module.css
+++ b/src/bottom-bar/data-item-bar.module.css
@@ -2,12 +2,12 @@
     display: flex;
     padding: 4px 12px;
     align-items: center;
-    border-bottom: 1px solid var(--colors-grey400);
+    border-block-end: 1px solid var(--colors-grey400);
 }
 
 .name {
     display: block;
-    margin-right: 8px;
+    margin-inline-end: 8px;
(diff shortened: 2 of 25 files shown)


cooper-joe:
> _Note:_ I didn't add changes for [sizing mappings](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Sizing) .. they affect vertical languages only, and I personally found the mapping hard to follow and rationalise (but maybe subconscious bias since I know rtl languages but not vertical ones) .. do we support such languages right now?

Good question. There's an overview of the translation/localisation coverage on [this page on DHIS2.org](https://dhis2.org/localization/). Chinese and Korean are included, which I know were traditionally vertical languages. I'm not sure if this is the standard for localized webapps though, would be worth looking into. The changes to the interface components to support vertical directionality would be a big lift for a lot of apps though.

PR #112, opened 2022-08-16
fix(aoc selector): implement openPeriodsAfterCoEndDate

tomzemp:
This implements logic for `openPeriodsAfterCoEndDate` which "extends" the endDate of the categoryOption by the number of periods specified by `openPeriodsAfterCoEndDate` (e.g. if categoryOption should expire on `2022-01-15`, but it's being used on a Monthly dataset with `openPeriodsAfterCoEndDate:3`, the categoryOption would effectively expire on `2022-04-15` for the purposes of data entry on that data set)

The logic check for categoryOption already existed in the code (https://github.com/dhis2/data-entry-app/blob/development/src/shared/metadata/selectors.js#L394): `periodEndDate > endDate`, which I have adapted to accommodate for `openPeriodsAfterCoEndDate`

I initially extended the endDate's of the individual categoryOptions, but I switched in the version I am submitting to instead reduce the dataset's periodEndDate. This seems a little less logically intuitive than extending individual categoryOptions' endDates, but it should be mathematically equivalent (i.e. `periodEndDate > (endDate + X periods)` = `(periodEndDate - X periods) > endDate`. I thought this approach was ultimately better because:

- if there are a lot of categoryOptions with endDates, this saves calculation time because we only do the adjustment for `openPeriodsAfterCoEndDate` once on the periodEndDate rather than multiple times on each of the categoryOptions' endDates
- I think this could better avoid some peculiarities that could arise with regards to end of period categoryOption endDates (e.g. if we have a Monthly dataset and our categoryOption has endDate of `2021-04-30` and we extend it because  openPeriodsAfterCoEndDate:3, we get `2021-07-30` which is probably not what a user would want? Instead if we reduce the periodEndDate, if we are in 2021 July, we would get an updated periodEndDate of `2021-04-30` to compare against)

### DIFF
--- a/src/shared/metadata/selectors.js
+++ b/src/shared/metadata/selectors.js
@@ -3,6 +3,7 @@ import { createSelector } from 'reselect'
 import {
     addFullPeriodTimeToDate,
     parsePeriodId,
+    removeFullPeriodTimeToDate,
 } from '../fixed-periods/index.js'
 import { cartesian } from '../utils.js'
 // Helper to group array items by an identifier
@@ -378,8 +379,6 @@ const isOptionWithinPeriod = ({
     periodStartDate,
     periodEndDate,
     categoryOption,
-    openPeriodsAfterCoEndDate,
-    periodType,
 }) => {
     // option has not start and end dates
     if (!categoryOption.startDate && !categoryOption.endDate) {
@@ -395,12 +394,7 @@ const isOptionWithinPeriod = ({
     }
 
     if (categoryOption.endDate) {
-        let endDate = new Date(categoryOption.endDate)
-        if (openPeriodsAfterCoEndDate) {
-            for (let i = 0; i < openPeriodsAfterCoEndDate; i++) {
-                endDate = addFullPeriodTimeToDate(endDate, periodType)
-            }
-        }
+        const endDate = new Date(categoryOption.endDate)
         if (periodEndDate > endDate) {
             // option end date is before period end date
             return false
@@ -433,6 +427,7 @@ export const getCategoriesWithOptionsWithinPeriod = createCachedSelector(
             metadata,
             dataSet?.id
         )
+
         const categoryOptions = getCategoryOptions(metadata)
         const relevantCategoriesWithOptions = resolveCategoryOptionIds(
             relevantCategories,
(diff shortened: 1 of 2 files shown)


src/shared/metadata/selectors.js
```diff
@@ -412,22 +416,46 @@ const resolveCategoryOptionIds = (categories, categoryOptions) => {
...
         const relevantCategoriesWithOptions = resolveCategoryOptionIds(
             relevantCategories,
             categoryOptions
         )
         const period = parsePeriodId(periodId)
         const periodStartDate = new Date(period.startDate)
-        const periodEndDate = new Date(period.endDate)
+
+        // periodEndDate is up to following start date
+        let periodEndDate = addFullPeriodTimeToDate(
```
Birkbjo, line 440: Forgive my ignorance, I'm not too familiar with the date stuff here. But what is the difference between this (and the `periodEndDate.setDate(periodEndDate.getDate() - 1)` below), and using the previous `new Date(period.endDate)`.

src/shared/metadata/selectors.js
```diff
@@ -412,22 +416,46 @@ const resolveCategoryOptionIds = (categories, categoryOptions) => {
...
+            periodStartDate,
+            period?.periodType?.type
+        )
+
+        // reduce perfiodEndDate by openPeriodsAfterCoEndDate
+        const openPeriodsAfterCoEndDate = Math.max(
+            dataSet?.openPeriodsAfterCoEndDate || 0,
+            0
+        )
+        for (let i = 0; i < openPeriodsAfterCoEndDate; i++) {
```
Birkbjo, line 450: Would it make sense to implement a `removePeriodsTimeToDate`-function in `fixed-periods.js`?

`const removePeriodsTimeToDate = periods => createUpdateFullPeriodTimeToDate(-1 * periods).`

Birkbjo (approved):
Looks good to me! Great with the context in the description and the tests!

I agree that reducing the `endDate` makes sense, as opposed to extending the option end dates.

src/shared/metadata/selectors.js
tomzemp: @Birkbjo: so previously we were just taking the endDate for the period, which was easy because we could just create a date object from the endDate associated with the period, so we create a date object for April 2022 from `new Date('2022-04-30')`, for example.

When we need to reduce the date however, we can get into a situation where just reducing the endDate itself will cause problems. For example, if we decrement the month on `2022-04-30`, we get `2022-03-30`, which is now not the end of the period March 2022. For that reason, in this code, I'm initiating the endDate as the next period's start date (e.g. instead of `new Date('2022-04-30')`, I used `new Date('2022-05-01')`, then decrement the periods, e.g. decrement 1 month to get `2022-04-01`, then subtract a day to get `2022-03-31` (i.e. we get to the endDate for the previous period).

Agree (in light of the comment below), that it would be better to move this logic elsewhere

Birkbjo: Ah, of course. That makes perfect sense!

PR #113, opened 2022-08-16
feat: show grey triangle when fields have a comment

HendrikThePendric:
It took me a while to find a somewhat suitable way to introduce this. Therefore I have left some comments in the PR to explain some of the quirks I encountered.

#### Known issue
The fact that currently POSTing values, comments, etc. for a dataset with an AOC is broken, also affects this feature. You can test datasets with just a default AOC. Custom forms work too.

### DIFF
--- a/src/data-workspace/data-workspace.js
+++ b/src/data-workspace/data-workspace.js
@@ -16,10 +16,16 @@ import {
 import styles from './data-workspace.module.css'
 import { EntryForm } from './entry-form.js'
 import { FinalFormWrapper } from './final-form-wrapper.js'
+import { useHasCommentContext } from './has-comment/has-comment-context.js'
 
 export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
     const { data: metadata } = useMetadata()
-    const initialDataValuesFetch = useDataValueSet()
+    const { populateHasCommentContextForDataSetValues } = useHasCommentContext()
+    const initialDataValuesFetch = useDataValueSet({
+        onSuccess: (data) => {
+            populateHasCommentContextForDataSetValues(data.dataValues)
+        },
+    })
     const isValidSelection = useIsValidSelection()
     const [dataSetId] = useDataSetId()
     // used to reset form-state when context-selection is changed
--- a/src/data-workspace/has-comment/has-comment-context.js
+++ b/src/data-workspace/has-comment/has-comment-context.js
@@ -1,0 +1,32 @@
+import { useContext, createContext } from 'react'
+
+export const HasCommentContext = createContext({
+    populateHasCommentContextForDataSetValues: (dataSetValues) => {
+        throw new Error(
+            `Context function "populateHasCommentContextForDataSetValues" not set. Provided dataSetValues of type "${typeof dataSetValues}".`
+        )
+    },
+
+    updateHasCommentContext: (
+        dataElementId,
+        categoryOptionComboId,
+        comment
+    ) => {
+        throw new Error(
+            `Context function "updateHasCommentContext" not set. Tried to get dataElementId "${dataElementId}", categoryOptionComboId "${categoryOptionComboId}" and comment "${comment}".`
+        )
+    },
+
+    hasComment: (fieldname) => {
+        throw new Error(
+            `Context function "hasComment" not set. Tried to get fieldname "${fieldname}".`
+        )
+    },
+})
+
+export const useHasCommentContext = () => useContext(HasCommentContext)
+
+export const useHasComment = (fieldname) => {
+    const { hasComment } = useHasCommentContext()
+    return hasComment(fieldname)
+}
--- a/src/data-workspace/has-comment/has-comment-provider.js
+++ b/src/data-workspace/has-comment/has-comment-provider.js
@@ -1,0 +1,66 @@
+import PropTypes from 'prop-types'
+import React, { useState, useCallback } from 'react'
+import { getFieldId } from '../get-field-id.js'
+import { HasCommentContext } from './has-comment-context.js'
+
+export const createCommentLookupFromDataSetValues = (dataSetValues) =>
+    Object.values(dataSetValues).reduce((lookup, dataElementCocs) => {
+        for (const [
+            categoryOptionComboId,
+            { dataElement: dataElementId, comment },
+        ] of Object.entries(dataElementCocs)) {
+            if (comment) {
+                lookup.add(getFieldId(dataElementId, categoryOptionComboId))
+            }
+        }
+        return lookup
+    }, new Set())
+
+export default function HasCommentProvider({ children }) {
+    const [lookup, setLookup] = useState(new Set())
+
+    const populateHasCommentContextForDataSetValues = useCallback(
+        (dataSetValues) => {
+            // Make sure to trigger a re-render when new dataSetValues are received
+            // because the cells are not updated otherwise
+            setLookup(createCommentLookupFromDataSetValues(dataSetValues))
+        },
+        []
+    )
+
+    const updateHasCommentContext = useCallback(
+        (dataElementId, categoryOptionComboId, comment) => {
+            const fieldname = getFieldId(dataElementId, categoryOptionComboId)
+
+            // Just mutate the exiting Set here, because adding/editing a comment
+            // will cause the cells to update anyway
+            if (comment) {
+                lookup.add(fieldname)
+            } else {
+                lookup.delete(fieldname)
+            }
+        },
+        [lookup]
+    )
+
+    const hasComment = useCallback(
+        (fieldname) => lookup.has(fieldname),
+        [lookup]
+    )
+
+    const value = {
+        populateHasCommentContextForDataSetValues,
+        updateHasCommentContext,
+        hasComment,
+    }
+
+    return (
+        <HasCommentContext.Provider value={value}>
+            {children}
+        </HasCommentContext.Provider>
+    )
+}
+
+HasCommentProvider.propTypes = {
+    children: PropTypes.node.isRequired,
+}
--- a/src/shared/use-data-value-set/use-data-value-set.js
+++ b/src/shared/use-data-value-set/use-data-value-set.js
@@ -69,7 +69,7 @@ const select = createSelector(
  *
  * TODO: This is no longer using the dataValueSet endpoint; should rename.
  */
-export const useDataValueSet = () => {
+export const useDataValueSet = ({ onSuccess } = {}) => {
     const isValidSelection = useIsValidSelection()
     const queryKey = useDataValueSetQueryKey()
     const activeMutations = useIsMutating({ mutationKey: ['dataValues'] })
@@ -91,6 +91,7 @@ export const useDataValueSet = () => {
             return !!query.meta.isHydrated
         },
         meta: { persist: true },
+        onSuccess,
     })
 
     return result
--- a/src/app/app-wrapper.js
+++ b/src/app/app-wrapper.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { HashRouter, Route } from 'react-router-dom'
 import { QueryParamProvider } from 'use-query-params'
+import HasCommentProvider from '../data-workspace/has-comment/has-comment-provider.js'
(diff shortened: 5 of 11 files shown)


src/data-workspace/data-details-sidebar/comment.test.js
```diff
@@ -145,7 +145,7 @@ describe('<Comment />', () => {
             return {
                 mutate: () => {
                     // fire after returning
-                    setTimeout(onSuccess, 0)
+                    setTimeout(() => onSuccess({}), 0)
```
HendrikThePendric, line 148: This test got broken because I am now passing the saved data value object to the `onSucces` handler, so had to apply this fix.

src/data-workspace/data-workspace.js
```diff
@@ -15,12 +15,17 @@ import {
...
+import { useHasCommentContext } from './has-comment/has-comment-context.js'
 
 export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
     const { data } = useMetadata()
     const [dataSetId] = useDataSetId()
-    const initialDataValuesFetch = useDataValueSet()
+    const { populateHasCommentContextForDataSetValues } = useHasCommentContext()
+    const initialDataValuesFetch = useDataValueSet((data) => {
+        populateHasCommentContextForDataSetValues(data.dataValues)
+    })
```
HendrikThePendric, line 26: The `useDataValueSet` hook now accepts a param which represents the `onSuccess` property of the `react-query` `useQuery`. I guess it could be good to make this more explicit, like so:
```suggestion
    const initialDataValuesFetch = useDataValueSet({
        onSuccess: (data) => {
            populateHasCommentContextForDataSetValues(data.dataValues)
        },
    })
```

But I'm not actually sure about that, as it might create the expectation that you could add lots of different options here.

In any case, this solution definitely wasn't my first choice to begin with. I would have much preferred to have implemented a `useEffect` hook here with a dependency to `dataValueSet`. However, since that object gets refreshed way too often that would have caused more performance issues.

src/data-workspace/data-workspace.js
```diff
@@ -15,11 +15,17 @@ import {
...
 export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
     const { data } = useMetadata()
     const [dataSetId] = useDataSetId()
-    const initialDataValuesFetch = useDataValueSet()
+    const { populateHasCommentContextForDataSetValues } = useHasCommentContext()
+    const initialDataValuesFetch = useDataValueSet({
+        onSuccess: (data) => {
+            populateHasCommentContextForDataSetValues(data.dataValues)
+        },
+    })
```
HendrikThePendric, line 28: The `useDataValueSet` hook now accepts a param which represents the `onSuccess` property of the `react-query` `useQuery`. I guess it could be good to make this more explicit, like so:
```suggestion
    const initialDataValuesFetch = useDataValueSet({
        onSuccess: (data) => {
            populateHasCommentContextForDataSetValues(data.dataValues)
        },
    })
```

But I'm not actually sure about that, as it might create the expectation that you could add lots of different options here.

In any case, this solution definitely wasn't my first choice to begin with. I would have much preferred to have implemented a `useEffect` hook here with a dependency to `dataValueSet`. However, since that object gets refreshed way too often that would have caused more performance issues.

src/data-workspace/data-workspace.js
```diff
@@ -15,11 +15,15 @@ import {
...
+import { useHasCommentContext } from './has-comment/has-comment-context.js'
 
 export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
     const { data } = useMetadata()
     const [dataSetId] = useDataSetId()
-    const initialDataValuesFetch = useDataValueSet()
+    const { populateHasCommentContextForDataSetValues } = useHasCommentContext()
+    const initialDataValuesFetch = useDataValueSet((data) => {
+        populateHasCommentContextForDataSetValues(data.dataValues)
+    })
```
HendrikThePendric, line 26: The `useDataValueSet` hook now accepts a param which represents the `onSuccess` property of the `react-query` `useQuery`. I guess it could be good to make this more explicit, like so:
```suggestion
    const initialDataValuesFetch = useDataValueSet({
        onSuccess: (data) => {
            populateHasCommentContextForDataSetValues(data.dataValues)
        },
    })
```

But I'm not actually sure about that, as it might create the expectation that you could add lots of different options here.

In any case, this solution definitely wasn't my first choice to begin with. I would have much preferred to have implemented a `useEffect` hook here with a dependency to `dataValueSet`. However, since that object gets refreshed way too often that would have caused more performance issues.

src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -57,7 +57,12 @@ const select = createSelector(
 
 const meta = { persist: true }
 
-export const useDataValueSet = () => {
+// TODO: investigate why the same data is fetch twice.
+// A. When refreshing the page it fires once when the form is rendered and the second
+// time when the form is focussed ()
+// B. When making a new selection it simply fires twice without any user-interaction
+// Also: it would make sense if this would return a 304, but we're getting a 200
```
HendrikThePendric, line 64: This worries me 😢

src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -67,6 +72,7 @@ export const useDataValueSet = () => {
         // Only fetch whilst offline, to prevent optimistic updates from being overwritten
         networkMode: 'online',
         meta,
+        onSuccess,
```
HendrikThePendric, line 75: I initially wanted to populate the has-comment-context directly from this hook (just implement the `onSuccess` handler here), since it is something that is directly related to loading the data values for the dataset. However, I found that when doing it that way the `onSuccess` handler would be called way too often. It turns out this behaviour is intentional in `react-query`, and the reasoning behind it is explained in https://github.com/TanStack/query/issues/1969. By implementing this as a parameter, `onSuccess` is only called by the instance of the hook which passes it, i.e. the one in the data-workspace.

src/shared/use-data-value-set/use-data-value-set.js
Birkbjo: This is a simple fix. Just needs `refetchOnMount: false` in the `useQuery`-options.

src/data-workspace/has-comment/has-comment-context.js
```diff
@@ -0,0 +1,32 @@
+import { useContext, createContext } from 'react'
+
+export const HasCommentContext = createContext({
```
Birkbjo, line 3: I'm a bit confused by the need for a context for this. Is it to prevent "subscribing" to `dataValues`?
I haven't looked too much into this, and I reckon you have a good reason for doing it like this. But to me it seems like this is something we should be able to compute pretty easily.

And to me it seems any change to the comment will re-render all fields regardless, since they subscribe to this context?

Birkbjo: And if it really is because we cannot reach into the dataValues in a performant way, I think its time to introduce something like https://github.com/pmndrs/zustand . 
I don't like that we have to keep the `dataValues` and this context in sync.

HendrikThePendric: Yeah I wasn't too happy with the implementation I ended up with, but there were some reasons why I resorted to it. I've left this PR a bit too long so I'm not a 100% sure if memory is serving me correctly, but I've checked a few things again and believe the following to be the case:
- The comment field is included the `dataValueSet` items in the `DataWorkSpace`, but that object changes a lot, i.e. whenever a field is changed and then blurred.
- The `DataEntryField` does not have direct access to its comment, and currently the data-entry fields don't all re-render every time the `dataValueSet` object changes. This is obviously behaviour we don't want to change.
- As such, deriving the `hasComment` value from the `dataValueSet` is problematic. The only way I could think of to avoid deriving the comment status from the `dataValueSet` was to take a more imperative approach which gives us much more granular control of when the `hasComment` values are updated.

So yeah, the problem was basically as you have already suggested in your second comment: we can't use `dataValueSet`. I also agree with your dislike to the fact that we will essentially be storing duplicate/overlapping information in two places and will have to ensure this will match. But to me this is the core issue and personally I can't think of away around this, while keeping re-renders limited. From what I've seen, if we were to use the `final-form-set-field-data` package you linked to, we'd still have to imperatively update the field's metadata right? If this is the case, I think adding this additional package would not really add much benefit over using a context....

Perhaps I am failing to see something here.....

HendrikThePendric:
I returned this PR to draft status because I discussed a slightly more elegant approach with @Birkbjo yesterday. It will need to be reworked quite a bit.

src/data-workspace/data-workspace.js
Birkbjo: If we are to keep this I would prefer your second example. We could just pass in an `options`-param, and spread it to `useQuery`.

HendrikThePendric:
I explored the alternative solution turned out to come with some complications of its own. After discussing again with @Birkbjo we did identify some possible avenues for refactoring this app, but we also realised that in the current state of the codebase the current context-based solution introduced in this PR probably makes sense.

src/shared/use-data-value-set/use-data-value-set.js
Birkbjo: I think we can remove this comment since it's fixed in #123 .

src/data-workspace/has-comment/has-comment-provider.js
```diff
@@ -0,0 +1,66 @@
...
+            { dataElement: dataElementId, comment },
+        ] of Object.entries(dataElementCocs)) {
+            if (comment) {
+                lookup.add(getFieldId(dataElementId, categoryOptionComboId))
+            }
+        }
+        return lookup
+    }, new Set())
+
+export default function HasCommentProvider({ children }) {
```
Birkbjo, line 19: Should we add a TODO or a comment about wanting to investigate better query-keys and how that might mitigate the use of this context?

src/data-workspace/data-workspace.js
Birkbjo, line 26: Is this not something we always want to do? Would it not make sense to move this logic to the `useDataValueSet`-hook? 
I guess one downside is that every component that uses that hook would also be subscribed to the context (since you would need `useContext` inside).

On that note, we could also do as we do with some of the other contexts - splitting it into two contexts, so you have one "setter"-context, and one "value"-context. That way this component would not need to update when the value is changed. This can be done regardless if we go the route above.

Could potentially also create a "composing"-hook, like `useDataValueSetWithComment` or something?

src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -57,7 +57,12 @@ const select = createSelector(
 
 const meta = { persist: true }
 
-export const useDataValueSet = () => {
+// TODO: investigate why the same data is fetch twice.
+// A. When refreshing the page it fires once when the form is rendered and the second
+// time when the form is focussed ()
+// B. When making a new selection it simply fires twice without any user-interaction
+// Also: it would make sense if this would return a 304, but we're getting a 200
+export const useDataValueSet = (onSuccess) => {
```
Birkbjo, line 65: As noted above, I think it's cleaner to pass in an "override"-object, or at least an object that we destructure. You never know what we need in the future, and if we were to go this route we would need to change the API.

Birkbjo (changes requested):
Nothing major, but a few points worth discussing I think.

src/shared/use-data-value-set/use-data-value-set.js
Birkbjo: I think the update in #123 prevents this, but I'm not totally sure.

src/data-workspace/has-comment/has-comment-provider.js
HendrikThePendric: No we should not. Please see [this Slack thread](https://dhis2.slack.com/archives/CJF3MDK8Q/p1661859412925369) for some additional info. In the thread I suggest writing a JIRA issue, but in this case we should first actually discuss within the team how we would like to improve/refactor this app, so a JIRA ticket isn't very suitable yet. Instead I have created #133 where we can discuss improvements. Once we finalise that discussion we can create a JIRA issue/epic for it. And then during the v2.40 cycle we can see if we can allocate some time to work on it.

Birkbjo (approved):

PR #115, opened 2022-08-17
fix: cc and cp in dv mutations & refactor

KaiVandivier:
A regression was introduced that disabled the attribute parameters for data value mutations, this is intended to fix it

Includes considerable refactoring to the structure of the mutations, hopefully to make it easier to interpret and maintain


src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -81,15 +74,15 @@ CommentEditForm.propTypes = {
...
 export default function Comment({ item }) {
     const [open, setOpen] = useState(true)
     const [editing, setEditing] = useState(false)
-    const setDataValueComment = useSetDataValueCommentMutation(() =>
-        setEditing(false)
-    )
+    const setDataValueComment = useSetDataValueMutation({
+        deId: item.dataElement,
+        cocId: item.categoryOptionCombo,
+    })
```
KaiVandivier, line 85: Moved `onSuccess` callback to `mutate()` invocation

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -3,8 +3,8 @@ import { CircularLoader, Button, ButtonStrip, TextAreaFieldFF } from '@dhis2/ui'
 import PropTypes from 'prop-types'
 import React, { useState } from 'react'
 import { Form, Field } from 'react-final-form'
-import { ExpandableUnit, useContextSelection } from '../../shared/index.js'
-import { useSetDataValueCommentMutation } from '../use-data-value-mutation/index.js'
+import { ExpandableUnit } from '../../shared/index.js'
+import { useSetDataValueMutation } from '../data-value-mutations/index.js'
```
KaiVandivier, line 7: Integrated the `comment` mutation into `useSetDataValueMutation`

src/data-workspace/inputs/file-inputs.js
```diff
@@ -78,7 +79,7 @@ export const FileResourceInput = ({
         input.onChange('')
         input.onBlur()
         setSyncStatus({ syncing: true, synced: false })
-        deleteFile(dataValueParams, {
+        deleteFile(null, {
```
KaiVandivier, line 82: no variables needed here; the mutation function handles the dv params

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -92,27 +88,22 @@ export function EntryFieldInput({
...
         }),
-        [
-            fieldname,
-            dataValueParams,
-            disabled,
-            setSyncStatus,
-            onFocus,
-            onKeyDown,
-        ]
+        [fieldname, de, coc, disabled, setSyncStatus, onFocus, onKeyDown]
```
KaiVandivier, line 106: dataValueParams isn't needed at the input-level anymore since the mutation functions handle it, so now just passing `deId` and `cocId` to the input components

src/data-workspace/data-value-mutations/data-value-mutations.js
```diff
@@ -0,0 +1,113 @@
...
+    optimisticallySetDataValue,
+    optimisticallySetFileDataValue,
+} from './data-value-optimistic-updates.js'
+import {
+    useDeleteDataValueMutationFunction,
+    useSetDataValueMutationFunction,
+    useUploadFileDataValueMutationFunction,
+} from './mutation-functions.js'
+import useApiError from './use-api-error.js'
+import { useDataValueParams } from './use-data-value-params.js'
```
KaiVandivier, line 14: This file is the main meat of the refactor

src/data-workspace/data-value-mutations/optimistic-update-helpers.js
```diff
@@ -0,0 +1,97 @@
...
+    previousDataValues,
+    dataValueMutationParams
+) {
+    const { co, de, ou, pe } = dataValueMutationParams
+
+    return previousDataValues.findIndex(
+        (dataValue) =>
+            dataValue.categoryOptionCombo === co &&
+            dataValue.dataElement === de &&
+            // not sure the following two matchers are necessary
```
tomzemp, line 11: I don't see why matching on the orgUnit and period would be necessary given that the previous data values should only be relevant to the given dataset/period/orgunit/attribute options. I guess I would vote to remove these extra criteria or add in the ones for attribute options

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -13,21 +13,13 @@ const errorMessage = i18n.t(
...
-            co: item.categoryOptionCombo,
-            de: item.dataElement,
-            comment: values.comment,
-        }
-
-        return syncComment(variables)
+        return syncComment(
+            // Don't send `undefined` (or 'undefined' will be stored as the comment)
+            { comment: values.comment || '' },
+            { onSuccess: closeEditor }
```
tomzemp, line 21: do we need to have the editor close onMutate? In offline mode, it seems it won't close until the mutation goes through, which I assume is not what we want?

src/data-workspace/data-value-mutations/use-data-value-params.js
```diff
@@ -0,0 +1,33 @@
...
+        ou: orgUnitId,
+        pe: periodId,
+    }
+
+    // Add attribute params to context params if attribute is not default
+    // (useApiAttributeParams helper returns undefined props if attribute
+    // is default)
+    if (attributeCombo) {
+        // Get a ';'-separated listed of attribute options
+        const attributeOptionIdList = attributeOptions.join(';')
```
tomzemp, line 27: a bit of an esoteric point here...but the way the attributeOptions information is being stored does not guarantee order (https://github.com/dhis2/data-entry-app/blob/a29c303083067162deb75917527f33d1c705fc7c/src/context-selection/attribute-option-combo-selector-bar-item/use-selected.js#L10), so in ART Monthly Summary, for example, if you click on "Implementing Partner" and then "Project" you'd get one order like `{LFsZ8v5v7rq: 'B3nxOazOO2G', yY2bQYqNt0o: 'i4Nbp8S2G6A'}`, then if you navigate to a different form, this gets cleared, and if you go back and then select "Project" first and then "Implementing Partner", you will get `{yY2bQYqNt0o: 'i4Nbp8S2G6A', LFsZ8v5v7rq: 'B3nxOazOO2G'}`...
so useApiAttributeParams just uses Object.values and then we join them, this can result in non-matching queryKeys (e.g. `{cp: 'i4Nbp8S2G6A;B3nxOazOO2G'}!={cp: 'B3nxOazOO2G;i4Nbp8S2G6A'}`. This then causes problems in the caching of these data values (notable e.g. in offline mode). Maybe we could alphabetize when generating queryKeys/mutationKeys?

tomzemp (commented):
Looks good to me @KaiVandivier! Yay for attribute options 🎉 and the general refactor looks good for maintainability going forward 😀 

I've added some comments about some minor issues that I noticed. I think maybe the attribute option strings in queryKeys should be addressed (if you agree), so that it doesn't cause problems later. The point on the comments editor not closing is minor, so probably not worth spending time on right now

src/data-workspace/data-details-sidebar/comment.js
KaiVandivier: I think you're right, good point!

src/data-workspace/data-value-mutations/use-data-value-params.js
KaiVandivier: Astute observation! 😮 Alphabetizing them during query/mutation key formation would work, but what do you think about designing the attribute option selector to return the order of the `{ [category]: <attributeOptionId> }` object in the same order as the categories in the category combo? (e.g. always "Implementing Partner" before "Project")

tomzemp: that sounds good! A bit more work to get them in that set order, but probably more understandable too if anyone looks at the api requests

HendrikThePendric:
> much nicer and DRYer imo .. I would love to see us getting rid of these `de`, `cc` `cp` etc... abbreviations, and only have them in the last possible stage (`mutation-functions` I guess) but that can be done in the future in a separate PR

I agree with this @kabaros. We should use full names where possible in our app and only switch to using abbreviations where the API requires us to do so. But I also agree this can and should be done separately.

src/data-workspace/data-value-mutations/use-data-value-params.js
KaiVandivier: Made a Jira issue for this to address in another PR: https://dhis2.atlassian.net/browse/TECH-1342

Birkbjo: > designing the attribute option selector to return the order of the { [category]: <attributeOptionId> } object in the same order as the categories in the category combo? (e.g. always "Implementing Partner" before "Project")

If we were to do this, we would need to change the entire API, since it cannot be an object. JavaScript objects are not ordered - even if it kind of works to create an array, sort the properties and create the object - the order of object-properties are never guaranteed to be in order. 

So I think the easiest/best way is to just `.sort()` the `attributeOptionIdList` when it's used for the params.

HendrikThePendric (approved):
Thanks for this.

src/data-workspace/data-entry-cell/inner-wrapper.js
```diff
@@ -38,21 +41,25 @@ CommentIndicator.propTypes = { isComment: PropTypes.bool }
...
+    const dataValueParams = useDataValueParams({ deId, cocId })
     const activeMutations = useIsMutating({
-        predicate: (mutation) => {
-            const { de, co } = mutation.options.variables
-            return deId === de && cocId === co
-        },
+        mutationKey: getDataValueMutationKey(dataValueParams),
     })
 
+    // todo: maybe use mutation state to improve this style handling
```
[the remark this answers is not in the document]
KaiVandivier: Good point to include the Jira issue! I agree with you here

This todo has to do with https://dhis2.atlassian.net/browse/TECH-1316 -- I'll add it in the comment

src/data-workspace/data-value-mutations/use-data-value-params.js
KaiVandivier: Chose to sort these here: https://github.com/dhis2/data-entry-app/blob/29740a6de1aa21deb89d4b3cdfb08bb1479d9dce/src/shared/use-api-attribute-params.js#L25-L26

Will close the Jira above

HendrikThePendric (approved):
Thanks for adding JIRA issues to the todos @KaiVandivier

PR #116, opened 2022-08-22
fix: default form prop order

DNR800:
Jira ticket: [TECH-1318](https://dhis2.atlassian.net/browse/TECH-1318)

Have data elements display alphabetical order by displayFormName on default forms

Data table

### DIFF
--- a/src/shared/metadata/selectors.js
+++ b/src/shared/metadata/selectors.js
@@ -201,18 +201,21 @@ export const getDataElementsByDataSetId = createCachedSelector(
         }
 
         return dataSet.dataSetElements
-            .map((dse) => {
-                const de = dataElements[dse.dataElement.id]
+            .map((dataSetElement) => {
+                const dataElement = dataElements[dataSetElement.dataElement.id]
 
-                const categoryCombo = dse.categoryCombo ?? de.categoryCombo
+                const categoryCombo =
+                    dataSetElement.categoryCombo ?? dataElement.categoryCombo
 
                 return {
-                    ...de,
+                    ...dataElement,
                     categoryCombo,
                 }
             })
-            .sort((deA, deB) =>
-                deA.displayFormName?.localeCompare(deB.displayFormName)
+            .sort((dataElementA, dataElementB) =>
+                dataElementA.displayFormName?.localeCompare(
+                    dataElementB.displayFormName
+                )
             )
     }
 )((_, dataSetId) => dataSetId)
--- a/src/shared/metadata/selectors.js
+++ b/src/shared/metadata/selectors.js
@@ -48,6 +48,15 @@ export const getCategoryComboById = (metadata, id) =>
     getCategoryCombos(metadata)[id]
 
 /**
+ * Alphabetically sort the order of data elements on displayFormName
+ * @param {*} dataElements
+ */
+export const sortDataElementsByDisplayFormName = (dataElements) =>
+    dataElements.sort((deA, deB) =>
+        deA.displayFormName?.localeCompare(deB.displayFormName)
+    )
+
+/**
  *
  * @param {*} metadata
  * @param {*} categoryComboId
--- a/src/shared/metadata/selectors.js
+++ b/src/shared/metadata/selectors.js
@@ -220,7 +220,7 @@ export const getDataElementsByDataSetId = createCachedSelector(
  * @param {*} metadata
  * @param {*} dataSetId
  */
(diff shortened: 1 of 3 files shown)


src/shared/metadata/selectors.js
```diff
@@ -47,6 +47,15 @@ export const getSectionsByDataSetId = (metadata, dataSetId) => {
 export const getCategoryComboById = (metadata, id) =>
     getCategoryCombos(metadata)[id]
 
+/**
+ * Alphabetically sort the order of data elements on displayFormName
+ * @param {*} dataElements
+ */
+export const sortDataElementsByDisplayFormName = (dataElements) =>
```
Birkbjo, line 54: This doesn't really comply to the same API as the rest of the selectors, where we actually pass in the `metadata`.

I appreciate how easy this makes to test it though, but I don't think `selectors` is the right place for this. 

Could either move this and have this function in the `default-form.js`-file, or even make it inline (as a chained `.sort` on `selectors.getDataElementsByDataSetId(data, dataSet.id)`, even though that would make it more difficult to test. 

Another approach would be to make it an actual selector, something like this:

```javascript
export const getSortedDataElementsByDataSetId = createSelector(
    getDataElementsByDataSetId,
    (dataElements) =>
        dataElements.sort((deA, deB) =>
            deA.displayFormName?.localeCompare(deB.displayFormName)
        )
)
```

HendrikThePendric (changes requested):
It'd be good to see the comments by @Birkbjo and @kabaros addressed. Feel free to ping us on Slack to move the re-review along.

src/shared/metadata/selectors.js
DNR800: Thats a good point thanks Birk. I wasn't sure if as a stand alone function this was a good place for it to live. I had pulled this out partly to make the test a bit more straight forward but looking at the usage of `getDataElementsByDataSetId` it does seem to be used only in `default-form.js` so I'll look at chaining the `.sort` in `getDataElementsByDataSetId`. I think I can get something place in the tests to support that easy enough.

src/shared/metadata/selectors.js
```diff
@@ -200,16 +200,20 @@ export const getDataElementsByDataSetId = createCachedSelector(
             return undefined
         }
 
-        return dataSet.dataSetElements.map((dse) => {
-            const de = dataElements[dse.dataElement.id]
+        return dataSet.dataSetElements
+            .map((dse) => {
+                const de = dataElements[dse.dataElement.id]
```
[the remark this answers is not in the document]
Birkbjo: > could we rename these variables from de and dse to dataElement and dataSetElement .. I know the abbreviated forms are all over the codebase, but we have a ticket to remove those, and ideally we won't add to them 


I don't necessarily disagree - but I feel like there's a difference between having APIs and objects that pass these abbreviated forms around, and having these abbreviations in arrow-functions like above, where you can see the full name one line above (`dataSet.dataSetElements`).

src/shared/metadata/selectors.js
Birkbjo: It's actually also used in `getDataElementsBySection`-selector, but that uses it only to find the elements and the order is determined by the `section.dataElements`. So it should be fine. At the same time you would do an unnecessary sort though. Not crucial, but I still think this should either be a different selector (like above), or just do it inline in `DefaultForm`.

src/shared/metadata/selectors.js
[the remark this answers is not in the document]
DNR800: I've made the change here anyway - was just small and I thought the full names did actually make make it easier to understand especially after working though the tests

src/shared/metadata/selectors.js
DNR800: Ok I'll just go with doing it inline in the DefaultForm

Birkbjo (approved):
Just a minor comment, you can decide if you agree or not.

src/shared/metadata/selectors.js
DNR800: @Birkbjo Just pushed those changes. Gone with the name `getDataElementsByDataSetIdSortedOnDisplayFormName` its long but descriptive

HendrikThePendric (approved):

src/shared/metadata/selectors.js
DNR800: Chatting with @Birkbjo decide to shorten this to just `getDataElementsByDataSetIdSorted`

PR #117, opened 2022-08-23
fix(min/max validators): allow entry of zero,decimals,negatives

tomzemp:
This fixes the issue where zeros, negatives, decimal values could not be entered (even when valid): https://dhis2.atlassian.net/browse/TECH-1327

I noticed the save button was clickable when there were errors (but no post request was sent), so I have disabled it when there are errors

### DIFF
--- a/src/data-workspace/inputs/validators.js
+++ b/src/data-workspace/inputs/validators.js
@@ -81,10 +81,14 @@ export const validatorsByValueType = {
     [VALUE_TYPES.URL]: url,
 }
 
-// minimum and maximum values are restricted to integers
-export const validatorsByValueTypeMinMax = {
-    ...validatorsByValueType,
+export const minMaxValidatorsByValueType = {
+    [VALUE_TYPES.INTEGER]: integer,
+    [VALUE_TYPES.INTEGER_POSITIVE]: integerPositive,
+    [VALUE_TYPES.INTEGER_NEGATIVE]: integerNegative,
+    [VALUE_TYPES.INTEGER_ZERO_OR_POSITIVE]: integerZeroOrPositive,
+    // backend restricts minimum and maximum to integers
     [VALUE_TYPES.NUMBER]: integer,
+    [VALUE_TYPES.PERCENTAGE]: integerZeroOrPositive,
 }
 
 // This is an internal helper of the ui-forms library,
--- a/src/data-workspace/inputs/validators.js
+++ b/src/data-workspace/inputs/validators.js
@@ -81,6 +81,12 @@ export const validatorsByValueType = {
     [VALUE_TYPES.URL]: url,
 }
 
+// minimum and maximum values are restricted to integers
+export const validatorsByValueTypeMinMax = {
+    ...validatorsByValueType,
+    [VALUE_TYPES.NUMBER]: integer,
+}
+
 // This is an internal helper of the ui-forms library,
 // can be removed once `createLessThan` and `createMoreThan` have been moved to
 // the @dhis2/ui-forms library
--- a/src/data-workspace/inputs/validators.js
+++ b/src/data-workspace/inputs/validators.js
@@ -61,6 +61,12 @@ export const integerZeroOrPositive = composeValidators(
 export const integerNegative = composeValidators(integer, createMaxNumber(-1))
 
 export const percentage = createNumberRange(0, 100)
+const percentageInteger = composeValidators(
+    integer,
+    createMinNumber(0),
+    createMaxNumber(100)
+)
+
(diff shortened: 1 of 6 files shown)


tomzemp:
Actually, it seems the backend is rejecting values with decimals with the following message:

Cannot deserialize value of type `java.lang.Integer` from String \"-3.5\": not a valid `java.lang.Integer` value\n at [Source: (org.springframework.util.StreamUtils$NonClosingInputStream); line: 1, column: 101] (through reference chain: org.hisp.dhis.webapi.webdomain.datavalue.MinMaxValueDto[\"minValue\"])

So, I'll add a note on the ticket, and ask someone re: backend implications.

We don't have frontend messaging about the error either, which would be good to add

src/data-workspace/inputs/validators.js
```diff
@@ -95,7 +101,9 @@ export const createLessThan = (key, description) => {
     )
 
     return (value, allValues) =>
-        isEmpty(value) || isEmpty(allValues[key]) || value < allValues[key]
+        isEmpty(value) ||
+        isEmpty(allValues[key]) ||
+        parseFloat(value) < parseFloat(allValues[key])
```
tomzemp, line 106: parsing to number is required to make correct comparison of -7 < -1 (for example)

tomzemp:
Since noticing the issue with the backend restriction of min/max values to integers, I have
- Made validation for valueType:NUMBER be the same as INTEGER_NEGATIVE_OR_POSITIVE
- fixed error alerts (noticed when testing that error messages weren't displayed)
- moved parsing to ensure correct number propType in relevant components

Should now be ready for review

src/data-workspace/inputs/validators.js
```diff
@@ -81,6 +81,12 @@ export const validatorsByValueType = {
     [VALUE_TYPES.URL]: url,
 }
 
+// minimum and maximum values are restricted to integers
+export const validatorsByValueTypeMinMax = {
+    ...validatorsByValueType,
+    [VALUE_TYPES.NUMBER]: integer,
+}
```
KaiVandivier, line 88: I admit I was a little confused why all the other validators are in this object if limits can only be integers, but I realized it's because the existing `integer`, `integerPositive`, `integerNegative` etc. validators are functional without changing -- I have a few suggestions for clarity for [ignorant] people like me 😁 feel free to take or leave them as you like:

1. Include in this object only the value types that can have limits
2. Add to the comment above: `// (not all values can have min/max limits though)`
3. Maybe call the object `minMaxValidatorsByValueType`?

(And as mentioned in DMs, `percentage` seems to need a special min-max validator here)

KaiVandivier (commented):
Looks good! Just a comment on `validatorsByValueTypeMinMax` 🙂

src/data-workspace/inputs/validators.js
HendrikThePendric: I agree with @KaiVandivier's comment above. Especially this one:
> 1. Include in this object only the value types that can have limits

And as an additional point, add a comment above the `VALUE_TYPES.NUMBER` property to emphasise the fact that this is treated as an integer, which is different than how it is treated in `validatorsByValueType`.

HendrikThePendric (approved):
Optionally it would be good to implement the suggestion left by @KaiVandivier, but otherwise this looks great. Thanks.

src/data-workspace/inputs/validators.js
tomzemp: Thanks @KaiVandivier and @HendrikThePendric! I've incorporated the comments above. @KaiVandivier also followed up to point out that I should restrict the PERCENTAGE to integers in 0-100 range.

When doing that, I also noticed that the logic we were using on the Inputs for these numbers was resulting in 0 values not being displayed (i.e. `format: (value) => (value ? value.toString() : '')` made 0 display as `''`; I've updated this `format: (value) => (value !== undefined ? value.toString() : '')` to avoid that issue. (I thought that should probably be fixed in the context of this PR)

tomzemp:
Other issues noted (and not addressed) when working on this ticket:

- Backend allows min/max values for valueType:PERCENTAGE to be negative (even though the values themselves are restricted to 0-100)
- Backend will not reject a POST in which a non-integer number is a number, e.g. if you send something like `{maxValue: 50.1}`, this is accepted by the backend and saved as `maxValue: 50` without an error message. If you send a string, `{maxValue: '50.1'}`, you get the error that Java can't cast as integer.
- Frontend is not resetting the input values if one is in edit mode and clicks on a new dataElementOperand (I think it should probably just cancel the edit operation when a new operand is selected?)

Birkbjo (approved):

PR #118, opened 2022-08-23
fix(options menu): hide print options if no form is selected

tomzemp:
Hides print options when non are selected https://dhis2.atlassian.net/browse/TECH-1333


Birkbjo (approved):
LGTM

KaiVandivier (approved):

PR #120, opened 2022-08-23
fix(section select): autoselect 1of1 sections

tomzemp:
autoselects first sections if there is only 1 (and hides "All sections") from menu


tomzemp:
Note: this is in draft mode as the current behaviour of selecting 'All sections' is specified in the design docs. Waiting feedback on ticket: https://dhis2.atlassian.net/browse/TECH-1323

tomzemp:
Closed because it was determined that the existing design was acceptable

PR #121, opened 2022-08-24
feat: implement followup logic

kabaros:
implements [TECH-1280](https://dhis2.atlassian.net/browse/TECH-1280)

mark_for_followup

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2022-08-16T13:42:05.663Z\n"
-"PO-Revision-Date: 2022-08-16T13:42:05.663Z\n"
+"POT-Creation-Date: 2022-08-26T05:32:30.241Z\n"
+"PO-Revision-Date: 2022-08-26T05:32:30.241Z\n"
 
 msgid "Not authorized"
 msgstr "Not authorized"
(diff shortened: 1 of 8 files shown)


tomzemp (commented):
Looks good @kabaros!

Two things I noticed (beyond the small comments):

1. it seems that the backend when a value is deleted will remove the followUp flag. We can see this, for example in this sequence

- curl -X POST 'https://debug.dhis2.org/dev/api/39/dataValues?de=RR538iV9G1X&co=HllvX50cXC0&ds=rsyjyJmYD4J&ou=ImspTQPwCqd&pe=2020April&value=42' -u admin:district
- curl -X POST 'https://debug.dhis2.org/dev/api/39/dataValues?de=RR538iV9G1X&co=HllvX50cXC0&ds=rsyjyJmYD4J&ou=ImspTQPwCqd&pe=2020April&followUp=true' -u admin:district
- curl -X GET 'https://debug.dhis2.org/dev/api/dataAnalysis/followup?ds=rsyjyJmYD4J&startDate=2020-04-01&endDate=2021-03-31&ou=ImspTQPwCqd' -u admin:district
- curl -X DELETE 'https://debug.dhis2.org/dev/api/39/dataValues?de=RR538iV9G1X&co=HllvX50cXC0&ds=rsyjyJmYD4J&ou=ImspTQPwCqd&pe=2020April'  -u admin:district
- curl -X GET 'https://debug.dhis2.org/dev/api/dataAnalysis/followup?ds=rsyjyJmYD4J&startDate=2020-04-01&endDate=2021-03-31&ou=ImspTQPwCqd' -u admin:district

I think we need to probably need to include logic in the optimistic update for deletes to remove these followUps. (It seems like restricting the "Unmark for follow-up" to cases where there is a value would be insufficient because they could then add a followUp, delete the value, add back a value and it could appear as though there is a followUp once the value was added back whereas it would have been removed by the delete operation on the backend)

2. The following isn't an issue with your code, but I noticed it during review, so I'll just mention it here: when I set my connection to offline and mark an item for followUp, I see the optimistic update sets the button to "Unmark for follow-up". When I then go back online, the POST request with the followUp goes through, but the UI updates and the "Mark for follow-up" comes back (I observe similar with some of the other updates, so I'll open a new ticket if we don't already have one)

tomzemp (approved):
Updates look good @kabaros and sounds like a good plan to address the other issue in a separate PR 👍 

I also noticed that the followUps aren't posting for fileTypes...which I think is a backend problem?

PR #122, opened 2022-08-24
docs: add model diagram

Birkbjo:
Not sure about the name of the file though. Any feedback is also welcome.


Birkbjo:
> thanks for this @Birkbjo .. I would just ask if you can add in parentheses the abbreviations we use in the API for these concepts .. like cc cp co etc... they confuse the hell out of me! 

I can, but I don't know how useful it is, since this doesn't explain the `attributeOptionCombo`. `cc` and `cp` is used to refer to the selected `categoryCombo` on the `dataSet` itself (we tend to call this an `attribute category combo`, and the resulting selection `attributeOptionCombo`, to distinguish from `categoryOptionCombo` on `dataElements`). This `attributeOptionCombo (aoc)` is selected in the context-selection, (eg. for data set `Emergency Response`), which would have aocs `Target` and `Result`. 

So only `co (category option combo)` and `de (data element) ` from this diagram is actually used in the API-calls.


To be honest I wish we updated the name of those parameters... Either fully type them out like `attributeCategoryCombo (cc)`, `attributeCategoryOptions (cp)`, `categoryOptionCombo (co) `, or update them to more correct/descriptive abbreviations. eg. `ac`, `aco`, `coc`, respectively.  
However, there's some value in keeping the API consistent with older/other APIs. 


> also we can show or link the file directly in README.md so that it is easier to find?

Will add a link :)

README.md
```diff
@@ -4,3 +4,11 @@
 
 We're using use-query-params to store part of the app's state in the url query
 parameters. This allows users to bookmark and share specific states of the app.
+
+## Model overview
+
+The category model can be fairly confusing, so we've created a few diagrams to help explain and get an overview of how DHIS2 models are used in the app.
+
+[Models used in form](./docs/category-combo-diagram.png)
```
tomzemp, line 12: shouldn't the links in github markup just be `(docs/category-combo-diagram.png)`?

tomzemp:
Looks good to me and provides a really nice visual overview of these concepts! I'm not totally sure we'd want it in the app's main readme in the future, but I think it's useful, so if others think it fits, sounds good to me

tomzemp (approved):

Birkbjo:
>  I'm not totally sure we'd want it in the app's main readme in the future, but I think it's useful, so if others think it fits, sounds good to me

I agree. I think having a `README` in `/docs`, and having some more documentation and context there would be useful. And maybe a link from the root `README` to the docs under a "contribution"-section.

PR #123, opened 2022-08-24
fix: data values duplicate request

Birkbjo:
[TECH-1320](https://dhis2.atlassian.net/browse/TECH-1320)

This also fixes persisted queries. Previously, they would only be retrieved from  the cache "every other reload of the app", because the `meta`-property on the query was not persisted - so just hydrate queries would not be persisted again, due to lacking `meta.persist`. This should now be fixed by adding this property to all dehydrated queries.

### DIFF
--- a/src/app/query-client/configured-query-client-provider.js
+++ b/src/app/query-client/configured-query-client-provider.js
@@ -22,6 +22,8 @@ const persistOptions = {
         defaultOptions: {
             queries: {
                 meta: {
+                    // meta-property is not persisted, so this makes sure dehydrated-queries will
+                    // be persisted again
                     persist: true,
                     // can be used to check if the query originates from the persisted-store
                     isHydrated: true,
--- a/src/app/query-client/configured-query-client-provider.js
+++ b/src/app/query-client/configured-query-client-provider.js
@@ -18,6 +18,17 @@ const persistOptions = {
             return isSuccess && shouldPersist
         },
     },
+    hydrateOptions: {
+        defaultOptions: {
+            queries: {
+                meta: {
+                    persist: true,
+                    // can be used to check if the query originates from the persisted-store
+                    isHydrated: true,
+                },
+            },
+        },
+    },
 }
 
 export const ConfiguredQueryClientProvider = ({ children, queryClient }) => {
--- a/src/shared/highlighted-field/use-highlighted-field.js
+++ b/src/shared/highlighted-field/use-highlighted-field.js
@@ -49,10 +49,17 @@ export default function useHighlightedField() {
 
     useEffect(() => {
         const { de, coc } = item
-        setHighlightedFieldId(
-            gatherHighlightedFieldData({ de, coc, dataValueSet })
-        )
-    }, [item, dataValueSet])
+        // only update if it's different from currentItem
+        // prevents unnecessary re-render on first-render as well
+        if (
+            de.id !== currentItem?.dataElement ||
+            coc.id !== currentItem?.categoryOptionCombo
+        ) {
+            setHighlightedFieldId(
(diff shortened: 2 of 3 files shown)


src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -64,6 +64,13 @@ export const useDataValueSet = () => {
         // Only enable this query if there are no ongoing mutations
         enabled: isValidSelection,
         select,
+        refetchOnMount: (query) => {
+            // only refetch on mount if the query was just hydrated
+            // this should only happen during initial app-load.
+            // If we were to return 'false' the query would not be refetch on first mount
+            // because it's mounted with hydrated-state
+            return query.meta.isHydrated
+        },
```
HendrikThePendric, line 73: Related to my previous comment, I presently failing to see how `isHydrated` would ever get any other value than `true` here.

HendrikThePendric (changes requested):
It does not seem to work for me and I also don't really get why this would work. Perhaps I just need some guidance.

src/app/query-client/configured-query-client-provider.js
```diff
@@ -18,6 +18,19 @@ const persistOptions = {
...
+                meta: {
+                    // meta-property is not persisted, so this makes sure dehydrated-queries will
+                    // be persisted again
+                    persist: true,
+                    // can be used to check if the query originates from the persisted-store
+                    isHydrated: true,
+                },
+            },
+        },
+    },
```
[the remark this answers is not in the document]
Birkbjo: Yeah, as Ismay say these props are entirely custom. The `defaultOptions` here just make sure that every query that is hydrated from the persisted-store, will have these meta-properties set. Since they are not persisted, `meta` will not have any properties. 

What issues are you seeing?

src/shared/use-data-value-set/use-data-value-set.js
Birkbjo: When the `dataValueSet` is updated, it will not have that property because the entire query is "new" and overridden.

src/shared/highlighted-field/use-highlighted-field.js
```diff
@@ -49,10 +49,17 @@ export default function useHighlightedField() {
...
+        // prevents unnecessary re-render on first-render as well
+        if (
+            de.id !== currentItem?.dataElement ||
+            coc.id !== currentItem?.categoryOptionCombo
+        ) {
+            setHighlightedFieldId(
+                gatherHighlightedFieldData({ de, coc, dataValueSet })
+            )
+        }
+    }, [currentItem, item, dataValueSet])
```
Birkbjo, line 62: I noticed that this was causing an additional render, and was the "cause" of the duplicated request. However, I wouldn't think of this as a proper solution, since we shouldn't have to worry about excess request due to excessive renders.

src/app/query-client/configured-query-client-provider.js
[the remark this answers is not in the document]
Birkbjo: So after some more testing, I think I see what you mean. There's still an excessive request the first time the app is loadedd. I'm not totally sure about what causes it. But you can still see that this fix will prevent the same duplicated request if you eg. change the period, then highlight a cell.

I will look into why this happens. But a `staleTIme` would also prevent this. Which is probably something we should discuss across the board. I guess having a `staleTime` of a few seconds on each request won't hurt, and will prevent duplicate requests like this.

HendrikThePendric (approved):
https://github.com/dhis2/data-entry-app/pull/123/commits/2624dd65ec1e843651b9d09449e316a852e7ee99 seems to have fixed the issue. Awesome.

PR #124, opened 2022-08-26
feat(lock datasets): add handling for locked datasets

tomzemp:
### Known issues

- backend does not currently provide information on lockExceptions and dataset approval status, so the full logic for locking forms cannot be yet implemented; this PR implements logic for dataInputPeriods and expiryDays. However, it seems easier to submit the PR so that the ability to lock is in place in advance of all necessary information being returned from the backend.
- the logic implemented in this PR uses browser time. The check for dataInputPeriods on the back end uses server time. I've opened a ticket to address the overall issue of using server time in the app (https://dhis2.atlassian.net/browse/TECH-1348) as I think it's better addressed in one PR for all applicable period/locking logic
- design specs call for using a lock icon in the NoticeBox informing about a form's locked status. The NoticeBox component does not seem to support custom icons, so I am just using the default icons for the time being. I assume it's not worth updating the component or writing a separate implementation of NoticeBox just for the purposes of using the lock icon.

### Overall approach/technical notes
- I changed the app-level `selectionHasNoFormMessage` state to a NoFormOrLocked context/provider. Happy for other ideas here; it seemed the easiest for accessing information about whether the form was locked throughout the app 
- I moved the logic for setting the highlighted field id from EntryFieldInput component to InnerWrapper component because the disabled inputs could not be focused; this necessitated some change in some of the logic for setting the highlighted id. I couldn't get the InnerWrapper to focus on the fully disabled inputs (e.g. radio buttons), but it was possible to click on the InnerWrapper to set the highlighted field id.
- we decided on our call (31 August) that follow-up and min/max edits should be allowed when locked, but comment should be disabled. (now implemented)
- there's no datasets with expiryDays or dataInputPeriods in the Sierra Leone database, which I'm not sure is ideal for testing. I was thinking cypress and intercepting the api/metadata/dataEntry response, but held off on this for the time being.


tomzemp:
### Know issues

- backend does not currently provide information on lockExceptions and dataset approval status, so the full logic for locking forms cannot be yet implemented; this PR implements logic for dataInputPeriods and expiryDays. However, it seems easier to submit the PR so that the ability to lock is in place in advance of all necessary information being returned from the backend.
- the logic implemented in this PR uses browser time. The check for dataInputPeriods on the back end uses server time; the backend does not seem to restrict based on expiryDays (to confirm). I've opened a ticket to address the overall issue of using server time (https://dhis2.atlassian.net/browse/TECH-1348) as I think it's better addressed in one PR for all applicable period/locking logic
- design specs call for using a lock icon in the NoticeBox informing about a form's locked status. The NoticeBox component does not seem to support custom icons, so I am just using the defaults for the time being. I assume it's not worth updating the component or writing a separate implementation of NoticeBox just for the purposes of using the lock icon.

### Overall approach/technical notes
- I changed the app-level `selectionHasNoFormMessage` state to a NoFormOrLocked context/provider. Happy for other ideas here; it seemed the easiest for accessing information about whether the form was locked throughout the app 
- I moved the logic for setting the highlighted field id from EntryFieldInput component to InnerWrapper component because the disabled inputs could not be focused; this necessitated some change in some of the logic for setting the highlighted id
- in the existing app, you are allowed to edit min/max values, edit comments in the ui for data values in locked datasets. These will then be selectively rejected on the backend (rejected for expired data input periods, not rejected for expiryDays (consistent with lack of apparent check for expiryDays)). I think it seems more logical to restrict this on the front-end in general, but I can follow up on the original ticket to ask Lars if there's some reason for different behaviour for different locking reasons

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -63,56 +57,21 @@ export function EntryFieldInput({
...
-        setHighlightedFieldId({ de, coc })
-    }, [de, coc, setHighlightedFieldId])
-
     const sharedProps = useMemo(
         () => ({
             fieldname,
             dataValueParams,
-            disabled,
+            disabled: disabled,
+            locked,
```
tomzemp, line 71: I thought it would be easier to pass the `locked` boolean to each of the possible inputs from here rather than get it from the context, but that choice was fairly arbitrary

src/data-workspace/entry-form.js
```diff
@@ -38,6 +44,15 @@ export const EntryForm = React.memo(function EntryForm({ dataSet }) {
...
         <>
+            {noFormOrLocked?.message && noFormOrLocked?.inForm && (
+                <NoticeBox
+                    className={styles.warningBox}
+                    error={noFormOrLocked?.error}
+                    title={noFormOrLocked?.title}
+                >
+                    {noFormOrLocked?.message}
+                </NoticeBox>
+            )}
```
Mohammer5, line 55: This will never be displayed as the same condition is used in `data-workspace.js` which won't render the `<EntryForm />` when the condition is true.

Mohammer5 (changes requested):
I find the pattern of using a config (`no-form-and-locked-states.js`) confusing:

1. every option other than `OPEN` has `locked: true`, even when the form is technically not locked but shouldn't be displayed for other reasons (e.g. when some categories don't have options) which effectively causes that property to be the same as checking for `message === ''`.
2. The config conflates UI and state (message/title & error/inForm/locked). This architecture can leads to cases where adding an "exceptional" feature to a particular option increases the code's complexity
3. The context's name indicates that it's being used for multiple purposes (no form & locked). My preference would be to change the name to `NoFormContext`. A non-`OPEN` value already expresses that we want to lock the form

Considering that we can remove the display from the entry form, which leaves only one place where we render the `<NoticeBox />`, I'd prefer if we have components with `<NoticeBox />`s for each case. This allows each component to accept individual props (e.g. the display names of the categories without options; which is one of those "exceptional" features I was talking about). Instead of storing a config in the state, we can simply store a constant for which case should be active and let the components handle what they should render.
If you want to, you could add another hook (`useIsFormLocked`), which returns true/false, depending on what case is in the context's state. With the current implementation, it would simply return `context.value !== OPEN` but could be adjusted easily when this behavior gets more complex.

tomzemp:
Thanks for the review @Mohammer5

I've:

- refactored to keep Locked / NoForm separate (this involves reverting the NoForm handling back to what we previously had).
- refactored the locked-states.js to be exclusively focused on state, and have moved UI to EntryForm component (the UI differences are just around the message in the NoticeBox, so I have not defined separate components for these)
- removed the client-side check of expiryDays (the logic in the existing PR was wrong as users can still have access to forms past a dataSet's expiryDays if they have the `F_EDIT_EXPIRED` authority. If we want to mimic the old data entry app which checks expiryDays on both the front- and back-end, we need to do a check of user authorities. I've held off on this for now and have just relied on the backend check returned by api/dataEntry/dataValues
- I've added in handling on the `lockStatus` returned from `api/dataEntry/dataValues`. I'm not, to be honest, sure how to best organize these. The client-side checks can occur during context selection, so I've left the client-side useCheckLockStatus in the ContextSelector and populated it with the lockStatus cached in useDataValueSet() if there is already a violation other than data input period. But, maybe, it would be better to have the client-side checks also be in data-workspace.js? I found it a bit confusing how to best arrange these given that we need to check the dataInputPeriods client-side, the approval status needs to come from the backend, and the expiryDays comes from the backend (but could also be executed client-side). I'm open to ideas for how to rearrange
- I've also updated to allow users to run validations in `locked` state (consistent with the old app)

Let me know if you have more ideas for refactoring. Thanks!

**Further note**: I seem to see that adding a single data input period to one period (e.g. July 2022) is also resulting in the backend locking other periods without defined data input periods (e.g. May 2022 also gets locked). I will follow up on this as I believe this may be a bug on the back end

src/shared/locked-status/use-check-lock-status.js
```diff
@@ -0,0 +1,84 @@
...
+                const openingDate = new Date(dataInputPeriod.openingDate)
+                const closingDate = new Date(dataInputPeriod.closingDate)
+
+                if (openingDate > now || closingDate < now) {
+                    // mark as invalid for data input period
+                    setLockStatus(LockedStates.LOCKED_DATA_INPUT_PERIOD)
+                    return
+                }
+            }
+        }
```
Mohammer5, line 41: Less nesting, easier to parse (imo):

```suggestion
        // should be outside of the hook
        const isDataInputPeriodLocked = (dataSetId, periodId) => {
            const dataInputPeriod = selectors.getApplicableDataInputPeriod(
                metadata,
                dataSetId,
                periodId
            )
            
            if (!dataInputPeriod) {
                return false
            }
            
            const openingDate = new Date(dataInputPeriod.openingDate)
            const closingDate = new Date(dataInputPeriod.closingDate)
            const now = getCurrentDate()
            
            return openingDate > now || closingDate < now
        }

        if (isDataInputPeriodLocked(dataSetId, periodId)) {
            // mark as invalid for data input period
            setLockStatus(LockedStates.LOCKED_DATA_INPUT_PERIOD)
            return
        }
```

Mohammer5 (approved):
> * refactored to keep Locked / NoForm separate (this involves reverting the NoForm handling back to what we previously had).

:ok_hand: 

> * refactored the locked-states.js to be exclusively focused on state, and have moved UI to EntryForm component (the UI differences are just around the message in the NoticeBox, so I have not defined separate components for these)

:ok_hand: 

> * removed the client-side check of expiryDays (the logic in the existing PR was wrong as users can still have access to forms past a dataSet's expiryDays if they have the F_EDIT_EXPIRED authority. If we want to mimic the old data entry app which checks expiryDays on both the front- and back-end, we need to do a check of user authorities. I've held off on this for now and have just relied on the backend check returned by api/dataEntry/dataValues
I've added in handling on the lockStatus returned from api/dataEntry/dataValues. I'm not, to be honest, sure how to best organize these. The client-side checks can occur during context selection, so I've left the client-side useCheckLockStatus in the ContextSelector and populated it with the lockStatus cached in useDataValueSet() if there is already a violation other than data input period. But, maybe, it would be better to have the client-side checks also be in data-workspace.js? I found it a bit confusing how to best arrange these given that we need to check the dataInputPeriods client-side, the approval status needs to come from the backend, and the expiryDays comes from the backend (but could also be executed client-side). I'm open to ideas for how to rearrange

I think you're right, you could move the `useCheckLockStatus` into the `<DataWorkspace />` component. At least that's the code that's dealing with forms, the context selection should only have to worry about managing the "context".

In terms of how you've arranged the code, I think it's exactly they it's supposed to be. Everything that's shared between different "views" is in the shared folder, everything that's not shared is inside the individual "view"-modules.

> * I've also updated to allow users to run validations in locked state (consistent with the old app)

:ok_hand: 

---

Changes look very good to me!

src/shared/locked-status/use-check-lock-status.js
tomzemp: updated

src/shared/locked-status/locked-info-context.js
```diff
@@ -0,0 +1,8 @@
+import { createContext } from 'react'
+
+export const LockedContext = createContext({
+    // sets above properties
```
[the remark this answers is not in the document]
tomzemp: fixed to reference the `lockStatus` and `locked` and remove comment

tomzemp:
In addition to the refactor due to comments above, 
@HendrikThePendric's work on [TECH-1296](https://dhis2.atlassian.net/browse/TECH-1296) (https://github.com/dhis2/data-entry-app/pull/142/files) meant that some of the code in this PR (for allowing clicking of disabled cells) was no longer necessary, so the code has been updated to remove that

PR #127, opened 2022-08-29
fix(final-form-wrapper): stable initialValues

Birkbjo:
This prevents re-render of the entire form when a `dataValue` is updated.

From my limited testing this does not seem to cause any new issues. But it's worth testing more.

### DIFF
--- a/src/data-workspace/data-workspace.js
+++ b/src/data-workspace/data-workspace.js
@@ -9,7 +9,7 @@ import {
     useMetadata,
     selectors,
     useDataSetId,
-    useSelectionKey,
+    useContextSelectionKey,
     useDataValueSet,
     useIsValidSelection,
 } from '../shared/index.js'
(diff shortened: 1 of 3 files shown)


HendrikThePendric (changes requested):
This introduces a bug. To reproduce:

1. Make a valid context selection
2. Fill in some values
3. Change the context selection, i.e. change a period
4. See that the values which you entered in step 2 are filled in for this new period
5. Now refresh and see that the form is empty

So the issue now is that these initial values are not being reinitialised when the the selection changes.

FYI: this issue only occurs when if the context selection changes, without having an intermediate invalid state. So for example, if you were to click "clear selection" after step 2 above and then made a new selection, the issue would not occur.

Birkbjo:
> This introduces a bug. To reproduce:
> 
> 1. Make a valid context selection
> 2. Fill in some values
> 3. Change the context selection, i.e. change a period
> 4. See that the values which you entered in step 2 are filled in for this new period
> 5. Now refresh and see that the form is empty
> 
> So the issue now is that these initial values are not being reinitialised when the the selection changes.
> 
> FYI: this issue only occurs when if the context selection changes, without having an intermediate invalid state. So for example, if you were to click "clear selection" after step 2 above and then made a new selection, the issue would not occur.

Thanks for testing this. To be more specific this only seems to happen for already cached periods, or else the form will load and clear the state.

This also seems to be somewhat related to the cells keeping the highlighting. [TECH-1316](https://dhis2.atlassian.net/browse/TECH-1316?jql=project%20%3D%20%22TECH%22%20AND%20component%20%3D%20%22%5BAPP%5D%20Data%20Entry%22)

HendrikThePendric (approved):
Great solution! Very much approved.

KaiVandivier (approved):

PR #128, opened 2022-08-30
fix: post only once when boolean radio is cleared

kabaros:
implements [TECH-1338](https://dhis2.atlassian.net/browse/TECH-1338)

### Changes in this PR
We had an extra handler for `onBlur` that caused the boolean radio to sync twice. It's unnecessary to have it so it's removed in this PR.


tomzemp (approved):

tomzemp:
Yeah. I'm not sure the best way to deal with that. I don't know if we can check if the last value is '' onBlur to only trigger that POST? Anyway, we could open a separate JIRA for this if it's not easy to fix as part of this PR

PR #129, opened 2022-08-30
fix(org unit selector): restrict org unit tree

tomzemp:
This PR restricts Organisation Units to those assigned to the selected dataset, see https://dhis2.atlassian.net/browse/TECH-1298.

Joe might slightly update the final design, but I think this would be within the overall approach of the ticket


Mohammer5:
[This comment](https://github.com/dhis2/data-entry-app/blob/7fea2c6e2ecfd6b5366f141b61023fd1fc4d23c3/src/context-selection/org-unit-selector-bar-item/org-unit-selector-bar-item.js#L59:L64) can be removed with these changes.

src/context-selection/org-unit-selector-bar-item/org-unit-selector-bar-item.js
```diff
@@ -106,12 +131,36 @@ export default function OrganisationUnitSetSelectorBarItem() {
...
                         </div>
+                        <Divider margin="0" />
+                        <div className={css.labelContentContainer}>
+                            <IconBlock16 />
+                            <span className={css.label}>
+                                {i18n.t(
+                                    'Dataset is not assigned to this organisation unit'
+                                )}
+                            </span>
+                        </div>
```
Mohammer5, line 163: I found this a bit confusing until I realised that this is a "legend". I'm wondering if we should add a headline or if we should move this to the `Help` menu which contains a legend for cells as well as keyboard shortcuts already (cc @cooper-joe)

Mohammer5 (approved):
Code-wise this looks good! Just one question regarding a text

src/context-selection/org-unit-selector-bar-item/org-unit-selector-bar-item.js
tomzemp: The styling/legend here is copied from the approvals app, but I'm definitely happy to update as people want

@cooper-joe do you want us to hold off merging this until you've finalized the design? Or should I merge and then you can revisit the design later? 

This is what it looks like now:
org_unit_restrictions

Mohammer5: Ah, if this is an established pattern, then I don't mind.

> @cooper-joe do you want us to hold off merging this until you've finalized the design? Or should I merge and then you can revisit the design later?

I'm for merging this and address design changes later. It will result in the same amount of work as we would have to do now anyway, so I don't see a reason to defer merging this PR

cooper-joe: Agreed that it is slightly confusing when there's only a single legend. The pattern works fine in Data Approvals, but is a bit sub-optimal here. However, agreed that we should merge this PR now and revisit the design separately (if it's a priority).

PR #131, opened 2022-09-01
fix(audit history): update to new backend format, clean up

tomzemp:
This PR updates the audit history to use the current backend implementation (there are, however, some issues with that implementation).

Also fixes some minor issues (duplicate keys, overflowing div element, invalid proptype)

See https://dhis2.atlassian.net/browse/TECH-1281


Mohammer5:
Could you add the reason or a todo list to the issue's description so we know why it's in draft mode and what's left to do

tomzemp:
In draft mode, as backend issues were revealed while editing:

- https://dhis2.atlassian.net/browse/TECH-1355
- https://dhis2.atlassian.net/browse/TECH-1352

The fixes on these might necessitate further changes on the front end, so unsure if it makes sense to implement an intermediary front end fix

tomzemp:
_before (development branch)_
image

_after (this PR)_
image

src/data-workspace/data-details-sidebar/audit-log.js
```diff
@@ -74,56 +74,64 @@ export default function AuditLog({ item }) {
...
-                                <DataTableCell>
-                                    <div>
-                                        {auditType === 'DELETE' && (
-                                            <DeletedValue
-                                                previousValue={previousValue}
-                                            />
+                            return (
+                                <DataTableRow key={key}>
+                                    <DataTableCell>
+                                        {moment(created).format(
```
Mohammer5, line 107: Oh I overlooked that one in my PR: This is in server time, right? Could you use the new date utils and display the client date here?

Mohammer5 (changes requested):

src/data-workspace/data-details-sidebar/audit-log.js
tomzemp: updated @Mohammer5 . I needed to use `fromServerDate` util function to avoid variable number of hook renders

Mohammer5: I think we can safely use `useClientServerDate` here as the hook never re-sets the internal state and therefore never causes a re-render (see [`use-client-server-date.js`](https://github.com/dhis2/aggregate-data-entry-app/blob/development/src/shared/date/use-client-server-date.js))

Mohammer5: My bad! We're using `.map` here, so my suggestion doesn't make any sense!

Mohammer5 (approved):

PR #135, opened 2022-09-01
feat: add formatting functionality for dealing with leading zeros and decimal places

DNR800:
Jira: [TECH-1246](https://dhis2.atlassian.net/browse/TECH-1246)

Adds in a formatting function onBlur on the general input that deals with values like..

- `.123` converting to `0.123`
- `0.` or `123.` converting to `0` or `123` respectfully
-  `000.123` converting to `0.123`

This specifically does the format for all Number types 

leadingZeros

### DIFF
--- a/src/data-workspace/inputs/generic-input.js
+++ b/src/data-workspace/inputs/generic-input.js
@@ -1,6 +1,6 @@
 import cx from 'classnames'
 import PropTypes from 'prop-types'
-import React, { useState, useEffect } from 'react'
+import React, { useState } from 'react'
 import { useField } from 'react-final-form'
 import { NUMBER_TYPES, VALUE_TYPES } from '../../shared/index.js'
 import { useSetDataValueMutation } from '../data-value-mutations/index.js'
@@ -27,7 +27,27 @@ export const GenericInput = ({
     onFocus,
     disabled,
 }) => {
-    const [lastSyncedValue, setLastSyncedValue] = useState('')
+    const formatValue = (value) => {
+        if (value === undefined) {
+            return ''
+        } else if (
+            value.trim() &&
+            NUMBER_TYPES.includes(valueType) &&
+            Number.isFinite(Number(value))
+        ) {
+            return Number(value).toString()
+        } else {
+            return value.trim()
+        }
+    }
+    const { input, meta } = useField(fieldname, {
+        validate: validatorsByValueType[valueType],
+        subscription: { value: true, dirty: true, valid: true },
+        format: formatValue,
+        formatOnBlur: true,
+        parse: (value) => value ?? '',
+    })
+    const [lastSyncedValue, setLastSyncedValue] = useState(input.value)
     const { mutate } = useSetDataValueMutation({ deId, cocId })
     const syncData = (value) => {
         // todo: Here's where an error state could be set: ('onError')
@@ -42,35 +62,6 @@ export const GenericInput = ({
             }
         )
     }
-
-    const formatValue = (value) => {
-        if (value === undefined) {
-            return ''
-        } else if (
-            value.trim() &&
-            NUMBER_TYPES.includes(valueType) &&
-            Number.isFinite(Number(value))
-        ) {
-            return Number(value).toString()
-        } else {
-            return value.trim()
-        }
-    }
-
-    const { input, meta } = useField(fieldname, {
-        validate: validatorsByValueType[valueType],
-        subscription: { value: true, dirty: true, valid: true },
-        format: formatValue,
-        formatOnBlur: true,
-    })
-
-    // need to initialize lastSyncedValue
-    useEffect(() => {
-        if (input.value !== undefined) {
-            setLastSyncedValue(input.value)
-        }
-    }, []) // eslint-disable-line react-hooks/exhaustive-deps
-
     const handleBlur = () => {
         const { value } = input
         const formattedValue = formatValue(value)
@@ -83,7 +74,7 @@ export const GenericInput = ({
     return (
         <input
             {...input}
-            value={input.value ?? ''}
+            value={input.value}
             className={cx(styles.basicInput, {
                 [styles.alignToEnd]: NUMBER_TYPES.includes(valueType),
             })}
--- a/src/data-workspace/inputs/generic-input.js
+++ b/src/data-workspace/inputs/generic-input.js
@@ -17,12 +17,6 @@ const htmlTypeAttrsByValueType = {
     [VALUE_TYPES.URL]: 'url',
 }
 
-const formatNumberTypes = [
-    VALUE_TYPES.NUMBER,
-    VALUE_TYPES.UNIT_INTERVAL,
-    VALUE_TYPES.PERCENTAGE,
-]
-
 export const GenericInput = ({
     fieldname,
     deId,
@@ -54,7 +48,7 @@ export const GenericInput = ({
 
         if (
             value &&
-            formatNumberTypes.includes(valueType) &&
+            NUMBER_TYPES.includes(valueType) &&
             Number.isFinite(Number(value))
         ) {
             formattedValue = Number(value).toString()
--- a/src/data-workspace/inputs/generic-input.js
+++ b/src/data-workspace/inputs/generic-input.js
@@ -48,6 +48,7 @@ export const GenericInput = ({
 
         if (
             value &&
+            value.trim() &&
             NUMBER_TYPES.includes(valueType) &&
             Number.isFinite(Number(value))
         ) {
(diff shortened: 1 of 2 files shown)


src/data-workspace/inputs/generic-input.js
```diff
@@ -43,23 +43,51 @@ export const GenericInput = ({
         )
     }
 
+    const formatValue = (value) => {
+        let formattedValue = value
+
+        if (value && valueType === VALUE_TYPES.NUMBER) {
```
tomzemp, line 49: also relevant for UNIT_INTERVAL and PERCENTAGE valueTypes

src/data-workspace/inputs/generic-input.js
```diff
@@ -43,23 +43,51 @@ export const GenericInput = ({
...
+            if (value.startsWith('.')) {
```
tomzemp, line 50: I think this will lead to `.` interpreting to `0.` on blur?

src/data-workspace/inputs/generic-input.js
```diff
@@ -43,23 +43,51 @@ export const GenericInput = ({
...
     })
 
     const handleBlur = () => {
         const { value } = input
         const hasEmptySpaces = value && value.trim() === ''
         const { dirty, valid } = meta
-        if (dirty && valid && !hasEmptySpaces && value !== lastSyncedValue) {
-            syncData(value.trim())
+        if (
+            value &&
```
tomzemp, line 77: this seems to prevent the post of an empty value (i.e. a deletion of the data value)?

tomzemp:
Cool. I like the idea of fixing the format on blur as that seems most user-friendly! (though maybe from recent discussion, Lars actually wants to just enforce the leading 0)?

I made some individual comments, and I was just wondering if there was a reason you implemented with these successive checks as opposed to just relying on the `Number().toString()` (with handling for `NaN`) that you have in the last step? It seems that `Number()` will convert to the desired format generally (1 leading zero, no unnecessary trailing zeros)

DNR800:
Thanks for the feedback Tom. I had added the if statements just as I was hacking away at this and just adding the different cases as I went. Good point about `Number().toString()` working for all the switch statements here thats an easy enough change

src/data-workspace/inputs/generic-input.js
DNR800: Its something I found myself unexpected having to add this as a safety check for the use of `value.trim()` I was occasionally getting errors as value has `undefined` -  was a bit of a surprise as I had expect that `valid` would maybe deal with that but it didn't always seem to be the case.

src/data-workspace/inputs/generic-input.js
DNR800: Yes doing this onBlur is where it would happen, but maybe only doing it if there were additional digits following would have been good to. I think a more general use of `Number().toString()` like you suggest in the comment below would cover this too.

src/data-workspace/inputs/generic-input.js
DNR800: Thanks thats useful - I was wondering if over value types should be considered here too

src/data-workspace/inputs/generic-input.js
DNR800: Looking at the previous behaviour here though it looks like it it should still submit if the field is empty so I'll see about changing this accordingly

src/data-workspace/inputs/generic-input.js
Mohammer5: I think this applies to all values types in either `CAN_HAVE_LIMITS_TYPES` or `NUMBER_TYPES` (see [`value-types.js`](https://github.com/dhis2/data-entry-app/blob/TECH-1246/handling-leading-zeros-and-decimal-points/src/shared/value-types.js)).

DNR800: Awesome :-) thanks @Mohammer5 I'll make that change

HendrikThePendric (commented):
One thing I noticed:
You are having to deal with `undefined` in line 78 and 85 after you have introduced a `format` prop. This is because the `useField` hook will apply a default formatter if none is provided and, as it mentions [here](https://final-form.org/docs/react-final-form/types/FieldProps#format), this default formatter will take care of transforming `undefined` to `''`. So when you introduce your own `format` function you have to also ensure it does the same. Once you have done this, you should be able to restore line 85 to its former glory and remove the ternary on line 70.

I've left some suggestions in the code.

src/data-workspace/inputs/generic-input.js
```diff
@@ -43,23 +43,41 @@ export const GenericInput = ({
...
+            value.trim() &&
+            NUMBER_TYPES.includes(valueType) &&
+            Number.isFinite(Number(value))
+        ) {
+            formattedValue = Number(value).toString()
+        }
+
+        return formattedValue
+    }
```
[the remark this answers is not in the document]
HendrikThePendric: No need to call `.trim()` on line 54 because:
```js
Number(' .0001  ').toString()
// result: '0.0001'
Number(' 0000.0001  ').toString()
// result: '0.0001'
```

src/data-workspace/inputs/generic-input.js
```diff
@@ -27,7 +27,7 @@ export const GenericInput = ({
     onFocus,
     disabled,
 }) => {
-    const [lastSyncedValue, setLastSyncedValue] = useState()
+    const [lastSyncedValue, setLastSyncedValue] = useState('')
```
DNR800, line 30: lastSyncedValue this currently the cause of a couple of issues. Just trying to see what I can do about that

src/data-workspace/inputs/generic-input.js
```diff
@@ -43,23 +43,47 @@ export const GenericInput = ({
...
         validate: validatorsByValueType[valueType],
         subscription: { value: true, dirty: true, valid: true },
+        format: formatValue,
+        formatOnBlur: true,
     })
 
+    // need to initialize lastSyncedValue
+    useEffect(() => {
+        if (input.value !== undefined) {
+            setLastSyncedValue(input.value)
```
DNR800, line 70: I found myself doing this to get the lastSyncedValue to behave nicely - if the initial value isn't set the user can't delete the the value from the form.

src/data-workspace/inputs/generic-input.js
HendrikThePendric: Good observation. We need to initialise the  `lastSyncedValue` with `input.value`

src/data-workspace/inputs/generic-input.js
```diff
@@ -43,23 +43,41 @@ export const GenericInput = ({
...
-            syncData(value.trim())
+        if (dirty && valid && formattedValue !== lastSyncedValue) {
+            syncData(formattedValue)
         }
     }
 
     return (
         <input
             {...input}
+            value={input.value ?? ''}
```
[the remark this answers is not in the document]
HendrikThePendric: Another good find. I could also reproduce this. Strangely enough this doesn't happen when you remove the `format` function from the options passed to `useField`. And apart from the way you did it, it can also be fixed by adding a `parse` function. I'd argue that's marginally cleaner because we're now preventing the `input.value` from ever becoming `undefined` rather than having to deal with it being `undefined`.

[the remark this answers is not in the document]
DNR800: I did try using `parse` on `useField` as I would have preferred to use that - but that didn't fix the problem. I think `parse` might only execute on the inputs onChange event

src/data-workspace/inputs/generic-input.js
```diff
@@ -56,21 +40,28 @@ export const GenericInput = ({
...
     })
-
-    // need to initialize lastSyncedValue
-    useEffect(() => {
-        if (input.value !== undefined) {
-            setLastSyncedValue(input.value)
-        }
-    }, []) // eslint-disable-line react-hooks/exhaustive-deps
-
+    const [lastSyncedValue, setLastSyncedValue] = useState(input.value)
```
DNR800, line 50: Thanks @HendrikThePendric. This was actually the first thing I did before looking at using the `useEffect` hook to set the initial value on mount.

I think what I was seeing though was that the `lastSyncValue` continued to update beyond the setting the initial value. Everytime `input.value` changed `lastSyncValue` updated as well. Doing the change on mount with `useEffect` meant that this only update `lastSyncValue` once. Form a user point of view it seemed to work but I was worried about potentially for introducing some form of race condition with future development.

I'll double check this on my machine.

I did have one more idea for how to address this but I need to double check how the `entry-field-input.js` is working with the generic input first. I don't know if we can pass the the initial value for the field as a prop on the generic input at the entry field input level so that it only sets `lastSyncValue` with a nice clean contract.

DNR800: forget what I said in the comment above only updates once as we would like.

 `const [lastSyncedValue, setLastSyncedValue] = useState(input.value)` 

does work perfectly fine and I work be happy for that to go in :-)

DNR800: Actually testing this further I am seeing a problem with it.. let me just do a screen recording

DNR800: No screenshot needed, had a conversation with @HendrikThePendric instead. Pushed the final fix we arrived at :-)

HendrikThePendric (approved):

PR #139, opened 2022-09-06
fix: resume queries when reloading app

Birkbjo:
[TECH-1356](https://dhis2.atlassian.net/browse/TECH-1394)
Also fixes [TECH-1393](https://dhis2.atlassian.net/browse/TECH-1393)

We didn't `resumePausedMutations` after they were hydrated from the store, so we lost those values after the user reloaded the app and went online again.
This needed some additional changes, because the mutations are now not necessarily mounted - and thus each mutation needs a `mutationDefault` - that describes how to send that stored mutation (since the `mutateFn` themselves are not stored/serialized). 
One complication of that, is that the `mutateFn` did not have access to the `queryKey` (only the `variables` passed to mutate!). Which would make it so we would have to include the mutationKey params to all the variable-objects. However, after digging around I found that you have access to the `mutationKey` in the `this-context` of `mutateFn`. So we now 
leverage this to gather those params. Therefore I could also delete some the passing of those variables to the hooks. 

Also introduced a new structure for the `mutationKeys`. See [this blog post](https://tkdodo.eu/blog/leveraging-the-query-function-context#object-query-keys) for where I got the inspiration from. However, the reason to do this was that the `defaultMutations` needs different `mutateFn` for `update`, `delete` and `uploadFile` requests. The cleanest way to do this is to separate this into different `mutationKeys` - since they are not the same mutation! 

A cool bonus is that file uploads work out of the box while offline across reloads! Since we are using indexed-db, it's able to store the file object with the data, while you would normally have to serialize the file-data manually.

### DIFF
--- a/src/data-workspace/data-value-mutations/data-value-mutations.js
+++ b/src/data-workspace/data-value-mutations/data-value-mutations.js
@@ -107,15 +107,3 @@ export function useDeleteDataValueMutation({ deId, cocId }) {
         optimisticUpdateFn: optimisticallyDeleteDataValue,
     })
 }
-
-export function useSetDataValueMutationDefaults(queryClient) {
-    queryClient.setMutationDefaults(mutationKeys.file(), {
-        mutationFn: useUploadFileDataValueMutationFunction(),
-    })
-    queryClient.setMutationDefaults(mutationKeys.update(), {
-        mutationFn: useSetDataValueMutationFunction(),
-    })
-    queryClient.setMutationDefaults(mutationKeys.delete(), {
-        mutationFn: useDeleteDataValueMutationFunction(),
-    })
-}
--- a/src/data-workspace/data-value-mutations/data-value-mutations.js
+++ b/src/data-workspace/data-value-mutations/data-value-mutations.js
@@ -10,14 +10,12 @@ import {
     useSetDataValueMutationFunction,
     useUploadFileDataValueMutationFunction,
 } from './mutation-functions.js'
+import { mutationKeys } from './mutation-key-factory.js'
 import useApiError from './use-api-error.js'
 import { useDataValueParams } from './use-data-value-params.js'
 
-export function getDataValueMutationKey(dataValueParams) {
-    return ['dataValues', { params: dataValueParams }]
-}
-
 function useSharedDataValueMutation({
+    mutationKey,
     dataValueParams,
     mutationFn,
     optimisticUpdateFn,
@@ -28,7 +26,7 @@ function useSharedDataValueMutation({
 
     return useMutation(mutationFn, {
         retry: 1,
-        mutationKey: getDataValueMutationKey(dataValueParams),
+        mutationKey: mutationKey,
 
         onMutate: async (variables) => {
             // Cancel any outgoing refetches (so they don't overwrite our optimistic update)
@@ -73,10 +71,11 @@ function useSharedDataValueMutation({
  */
 export function useSetDataValueMutation({ deId, cocId }) {
     const dataValueParams = useDataValueParams({ deId, cocId })
-    const mutationFn = useSetDataValueMutationFunction(dataValueParams)
+    const mutationFn = useSetDataValueMutationFunction()
 
     return useSharedDataValueMutation({
         dataValueParams,
+        mutationKey: mutationKeys.update(dataValueParams),
         mutationFn,
         optimisticUpdateFn: optimisticallySetDataValue,
     })
@@ -87,9 +86,10 @@ export function useSetDataValueMutation({ deId, cocId }) {
  */
 export function useUploadFileDataValueMutation({ deId, cocId }) {
     const dataValueParams = useDataValueParams({ deId, cocId })
-    const mutationFn = useUploadFileDataValueMutationFunction(dataValueParams)
+    const mutationFn = useUploadFileDataValueMutationFunction()
 
     return useSharedDataValueMutation({
+        mutationKey: mutationKeys.file(dataValueParams),
         dataValueParams,
         mutationFn,
         optimisticUpdateFn: optimisticallySetFileDataValue,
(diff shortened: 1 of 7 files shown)


src/data-workspace/data-value-mutations/data-value-mutations.js
```diff
@@ -98,11 +98,24 @@ export function useUploadFileDataValueMutation({ deId, cocId }) {
...
+    queryClient.setMutationDefaults(mutationKeys.file(), {
+        mutationFn: useUploadFileDataValueMutationFunction(),
+    })
+    queryClient.setMutationDefaults(mutationKeys.update(), {
+        mutationFn: useSetDataValueMutationFunction(),
+    })
+    queryClient.setMutationDefaults(mutationKeys.delete(), {
+        mutationFn: useDeleteDataValueMutationFunction(),
+    })
+}
```
KaiVandivier, line 121: is this following the rules of hooks? 😅

src/data-workspace/data-value-mutations/index.js
```diff
@@ -1,7 +1,9 @@
 export {
-    getDataValueMutationKey,
     useSetDataValueMutation,
     useDeleteDataValueMutation,
     useUploadFileDataValueMutation,
+    useSetDataValueMutationDefaults,
 } from './data-value-mutations.js'
 export { useDataValueParams } from './use-data-value-params.js'
+export * from './mutation-functions.js'
```
KaiVandivier, line 8: I don't see any of these mutation functions imported from other files -- is this export needed?

src/data-workspace/data-value-mutations/mutation-functions.js
```diff
@@ -28,6 +28,16 @@ const UPLOAD_FILE_MUTATION = {
     data: (data) => data,
 }
 
+export const createSharedMutateFn = (engine, mutationObj) =>
```
KaiVandivier, line 31: Seems like this doesn't need exporting?

src/data-workspace/data-value-mutations/mutation-functions.js
```diff
@@ -28,6 +28,16 @@ const UPLOAD_FILE_MUTATION = {
...
+    function mutateFn(variables) {
+        const { mutationKey } = this
+        // gather params from mutationKey
+        const { params } = mutationKey[1]
```
KaiVandivier, line 35:
```suggestion
        // A trick: get params for the mutation (de, co, ds, cc, cp) from the
        // `mutationKey`, which is set in the mutation options object and can
        // be accessed on `this` in the mutation function.
        // This saves passing `dataValueParams` everywhere
        const { mutationKey } = this
        // gather params from mutationKey
        const { params } = mutationKey[1]
```

Just an explanation suggestion for posterity

src/data-workspace/data-value-mutations/data-value-mutations.js
Birkbjo: Yup, it might look weird since normally you would have it above. But there's no conditional heres, they're just nested in the argument. I could move them above though. 

```
const uploadFileMuateFn =  useUploadFileDataValueMutationFunction()
etc...
```

src/data-workspace/data-value-mutations/index.js
Birkbjo: This one is used in `query-client` to set the defaults. I didn't touch the other ones above.

src/data-workspace/data-value-mutations/mutation-functions.js
Birkbjo: True. At first I imported this in `use-query-client`, before I created a hook for setting the default functions. 
I guess that can be removed now. It's also the reason for splitting this out into a new function - so it could be used imperatively, outside of hooks. That's not strictly needed anymore, but I think it's clean, and it could be valuable in the future.

src/data-workspace/data-value-mutations/mutation-key-factory.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+    return [...rootKey, { ...theState }]
+}
+
+export const mutationKeys = {
+    all: createSubKey(),
+    update: createSubKey({ method: 'update' }),
+    file: createSubKey({ method: 'update', entity: 'file' }),
+    delete: createSubKey({ method: 'delete' }),
+}
```
KaiVandivier, line 18: Maybe I'm dumb lol, but took me a while to figure out what these keys would look like given the higher-order function -- suggestion to make it easy to understand for other dummies:

```suggestion
/**
 * Example keys:
 * mutationKeys.all() ==> ['dataValues']
 * mutationKeys.update(dvParams)
 *     ==> ['dataValues', { method: 'update', params: dvParams }]
 * mutationKeys.file(dvParams)
 *     ==> ['dataValues', { method: 'update', entity: 'file', params: dvParams }]
 * etc.
 */
export const mutationKeys = {
    all: createSubKey(),
    update: createSubKey({ method: 'update' }),
    file: createSubKey({ method: 'update', entity: 'file' }),
    delete: createSubKey({ method: 'delete' }),
}
```

Birkbjo: Agreed, was looking to comment this a bit more. A bonus is that you can use them "without" the params. Eg. `mutationKeys.update()` targets all `update`-mutations.

src/data-workspace/data-value-mutations/index.js
Birkbjo: Ah, true. Sorry I must've read this wrong. Yeah, I'm sure this can be removed.

KaiVandivier (approved):
Nicely done, thanks Birk! 🙌

src/data-workspace/data-value-mutations/data-value-mutations.js
Mohammer5: > is this following the rules of hooks? sweat_smile

I'm wondering the same. It's a bit odd to me that we set defaults every time a consumer of `useSetDataValueMutationDefaults` re-renders. I suppose it works technically right now, but conceptually it seems wrong to me and more frail than doing this in a `useEffect`, which is meant to be used for side-effects like these.

src/data-workspace/data-value-mutations/mutation-key-factory.js
Mohammer5: What's wrong with simply writing out these keys?

```js
const sharedKey = 'dataValues'

/**
 * MutationKeys for dataValues.
 * These are all functions that can be called with the dataValueParams
 * to narrow down the mutation
 */
export const mutationKeys = {
    all: (params) => [sharedKey, { params }],
    update: (params) => [sharedKey, { method: 'update', params }],
    file: (params) => [sharedKey, { method: 'update', entity: 'file', params }],
    delete: (params) => [sharedKey, { method: 'delete', params }],
}
```

That seems to be a lot easier to read (at least to me) and reduces the complexity of this module.

src/data-workspace/data-value-mutations/data-value-mutations.js
Birkbjo: Yeah, should probably add this in an effect. Should probably do the same with the setup in `useQueryClient`. Right now this is only called once, and that component is never re-render - so it works fine, but I agree - we shouldn't really rely on that.

Mohammer5: That's true. I'll leave it up to you how you want to proceed. In case you want to keep it as is, please add a TODO comment and a jira issue so we track this potential issue somewhere

src/data-workspace/data-value-mutations/mutation-key-factory.js
Birkbjo: I don't want `undefined` params, since when I tested it - it didn't seem to work properly with key-matching. 

Your second solution doesn't really work, since you're spreading the entire `params` object, not creating a nested `param` property.

Mohammer5: I got confused by the name "key" and didn't think of it referring to the "mutation key" but as the key of a vanilla js object.. Everything clear to me now!

Mohammer5 (approved):

PR #140, opened 2022-09-06
fix: remove followUp flag from deleted data value

kabaros:
implements [TECH-1354](https://dhis2.atlassian.net/browse/TECH-1354)

### Changes in this PR
When a data value is deleted (using DELETE or POST empty value), the API removes any flags set on it. We need to make sure that the same behaviour happens in the optimistic update, otherwise the flag remains on the item until the user refreshes the page.


Mohammer5 (approved):

tomzemp (approved):
🙌

PR #141, opened 2022-09-07
fix: deduplicate identical mutations

Birkbjo:
**NOTE**: This is a proof of concept. This does introduce bugs!

Eg. changing comments when offline, then changing value (or any other order) - will result in only the last action happening.
It's probably an easy fix once we use queryKeys properly, and can update individual values easily.

### DIFF
--- a/src/app/query-client/use-query-client.js
+++ b/src/app/query-client/use-query-client.js
@@ -25,6 +25,25 @@ const useQueryClient = () => {
         },
     })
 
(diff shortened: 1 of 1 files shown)


HendrikThePendric:
When you talk about _identical_ mutations, you mean _multiple mutations for the same data value_ right? You might want to change the PR title a bit because I don't think _identical_ is very apt.

Mohammer5:
By @Birkbjo:
> will come back to this, but pretty low priority

PR #142, opened 2022-09-07
fix: highlight by clicking next to input and keep when focussing sidebar

HendrikThePendric:


src/data-workspace/data-entry-cell/inner-wrapper.js
```diff
@@ -51,9 +52,11 @@ export function InnerWrapper({
     syncStatus,
 }) {
     const hasComment = useHasComment(fieldname)
+    const { item } = useHighlightedFieldIdContext()
+    const active = item && deId === item.de.id && cocId === item.coc.id
```
HendrikThePendric, line 56: This part is a potential performance issue.

tomzemp:
@HendrikThePendric: should this be added on option-set input type as well as clicking on the white space next to the option selector will not focus the cell?

HendrikThePendric:
Some features I added after the approval:
- Added support for this to the option-set input too
- Visually distinguish between highlighted+focussed and highlighted+unfocussed
- Clear the highlighted cell when a filter input gets focus

PR #144, opened 2022-09-07
feat: enable validation for min/max limits on generic-input

DNR800:
Jira: [TECH-1246](https://dhis2.atlassian.net/browse/TECH-1246)

If numeric limits are set the details panel the generic input using the min and max limits for validation

Min-Max-limits

First pass at this, I've tried to use an existing validator `createNumberRange` from the ui library and take advantage of `CAN_HAVE_LIMITS_TYPES`

I've ended up creating a small hook `useMinMaxLimits` which could potentially put to use in a refactor of the  [\<Limits \/\>](https://github.com/dhis2/data-entry-app/blob/ac2831d459b60d3cb8e6bb71981176dcb23627be/src/data-workspace/data-details-sidebar/limits.js) component too.


src/data-workspace/inputs/use-min-max-limits.js
```diff
@@ -0,0 +1,16 @@
+import { useDataValueSet } from '../../shared/index.js'
+
+export const useMinMaxLimits = (dataElementId) => {
```
DNR800, line 3: I've ended up creating a small hook useMinMaxLimits which could potentially put to use in a refactor of the [<Limits />](https://github.com/dhis2/data-entry-app/blob/ac2831d459b60d3cb8e6bb71981176dcb23627be/src/data-workspace/data-details-sidebar/limits.js) component too.

DNR800:
I think I might be seeing a bug when trying to delete limits by clicking `Delete Limits` in the Details panel

I get the following message

Trouble-deleteing-limits

Looking at the console looks like I'm getting a 409, anyone else come across this?

```
Request URL: https://debug.dhis2.org/dev/api/39/dataEntry/minMaxValues?dataElement=PwgKFzGtxU2&orgUnit=ImspTQPwCqd&categoryOptionCombo=HllvX50cXC0
Request Method: DELETE
Status Code: 409 
```
Payload
```
dataElement: PwgKFzGtxU2
orgUnit: ImspTQPwCqd
categoryOptionCombo: HllvX50cXC0
```
Response
```
{
	"httpStatus": "Conflict",
	"httpStatusCode": 409,
	"status": "ERROR",
	"message": "Data element not found or not accessible: `null`",
	"errorCode": "E1100"
}
```

src/data-workspace/inputs/use-min-max-limits.js
DNR800: I don't think the refactor of `\<Limits \/\>` is a lot of work, I can do that now and we can see what it looks like - we can always drop the commit if we don't like it

src/data-workspace/inputs/validators.test.js
```diff
@@ -0,0 +1,66 @@
+import { VALUE_TYPES, CAN_HAVE_LIMITS_TYPES } from '../../shared/index.js'
+import { validateByValueTypeWithLimits } from './validators.js'
+
+describe(`validateByValueTypeWithLimits`, () => {
+    describe(`when limits are defined`, () => {
```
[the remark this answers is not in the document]
DNR800: I think my main concern here was that the test also is covering the implementation of `createNumberRange` which comes from the UI library - I could mock this but then I general prefer to test the code in a way that would reflect real output for a user and also I think there is generally better value in testing the real code rather than overly relying on mocks

HendrikThePendric (commented):
I like this solution, but there is one issue with it:
What is annoying about the `useDataValueSet` hook is that currently a new object is created whenever a field is changed, and then blurred. And since the generic-input is using your new hook, it will cause every input to re-render every time that happens.
having said that, we have also merged a few other PRs lately that suffer from the same problem. So if performance doesn't seem too bad I suppose we can just merge this.

DNR800:
> I like this solution, but there is one issue with it: What is annoying about the `useDataValueSet` hook is that currently a new object is created whenever a field is changed, and then blurred. And since the generic-input is using your new hook, it will cause every input to re-render every time that happens. having said that, we have also merged a few other PRs lately that suffer from the same problem. So if performance doesn't seem too bad I suppose we can just merge this.

I had just take what was happening in the \<Limits \/\> component as a general lead on how to access the min and max values that we need. I can't say that I have noticed any degradation in performance whilst I have been testing this and working on it myself. If there is another way of accessing those min and max values I'd be happy to look at introducing that as potential performance - I just don't know what would like like at this point. Maybe there is room for a potential follow up ticket to spike out other avenues to improve the here?

HendrikThePendric (approved):
I first tested the development version of this branch locally at 4x CPU throttle and found the app almost unusable. However, I then tested the built version of the app (via the netlify link) and actually things were still working pretty OK at 4x throttle. So I'm happy to merge this in.

PR #145, opened 2022-09-07
fix: change radio button to sync on change rather than on blur

kabaros:
fixes [TECH-1353](https://dhis2.atlassian.net/browse/TECH-1353)

### Changes in this PR 

**Before the change**: for boolean radios, when a user clicks an option (yes or no), then immediately clicks the `clear` button, two POSTs are sent simultaneously one with the selected value, and another with the `empty value` from the Clear. This results in a race condition between the two POSTs and the result is not predictable.

Syncing on `blur` is a needed optimisation when it comes to input fields (we don't want to be syncing every character as the user types) but for inputs that just have a single action, it would be more user-friendly to sync the action immediately (same as we do with the clear button).

**After the change**: the sync of the radio buttons happens on click. This fixes the race condition issue, and more inline with what the user would expect - once they click yes or no, that value should be saved (not only when they tab to the next field).

(note: this behaviour is also inline with the other component with a `clear` button - the option set component)

### DIFF
--- a/src/data-workspace/inputs/boolean-radios.js
+++ b/src/data-workspace/inputs/boolean-radios.js
@@ -54,7 +54,12 @@ export const BooleanRadios = ({
         subscription: { value: true },
     })
 
-    const [lastSyncedValue, setLastSyncedValue] = useState()
+    const {
+        input: { value: fieldvalue },
+        meta,
+    } = useField(fieldname)
+    const [lastSyncedValue, setLastSyncedValue] = useState(fieldvalue)
+
     const { mutate } = useSetDataValueMutation({ deId, cocId })
     const syncData = (value) => {
         // todo: Here's where an error state could be set: ('onError')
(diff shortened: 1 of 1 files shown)


tomzemp:
Looks good @kabaros!

At the moment, it seems that we won't get the cell formatting until the cell is blurred, which is now a bit misleading because the value has actually been synced upon click of the radio button:
boolean_click

Should we blur on the click of the yes/no?

I think it is a bit undesirable (e.g. from a user training perspective) that we would have this input behave differently than other inputs in terms of sync upon value update/rather than upon user-initiated blur, but probably better than refactoring more as I imagine that would be time-consuming (and the current implementation blurs/syncs on hitting clear)

HendrikThePendric (approved):
Looks good.

cooper-joe:
I think this is OK, and fits into the system of cell states:
- If cell is synced and not active: green background, green triangle.
- If cell is synced and active: white background, green triangle.

There is some inconsistency, as @tomzemp says, with the handling of sync `onblur` vs. `onChange`, but I think there's enough difference in the way users interact with text inputs vs. selection controls that the inconsistency is trumped by "doing the expected thing".

Possible fix for the future, if this proves problematic, would be to move to a debounced sync `onChange` for text values, I think, rather than the other way around.

tomzemp (approved):
I defer to Joe on design issues 😀 and had no technical objections, so looks good to me!

PR #146, opened 2022-09-08
fix: sync value when editing a field to its original value

kabaros:
fixes [TECH-1397](https://dhis2.atlassian.net/browse/TECH-1397)

### Context

Before this change, if you edit a value from its original value `x` to another value `y` then back to the original value `x`, the last change does not cause a sync. This is because we were relying on `dirty` from React-Final-Form (RFF) to decide whether to sync or not, but `dirty` is [not set to true when the value is changed back to the initial value of the form](https://final-form.org/docs/final-form/types/FormState#dirty).

### Changes in this PR

- we now use what is, in effect, our own implementation of `dirty` where we compare the `lastSyncedValue` with the current value, and drop the use of `dirty` from RFF.

- `lastSyncedValue` also had a bug with some field types, where we were not setting the initial value of the field when it's loaded. This is also fixed in this PR

### DIFF
--- a/src/data-workspace/inputs/generic-input.js
+++ b/src/data-workspace/inputs/generic-input.js
@@ -65,8 +65,8 @@ export const GenericInput = ({
     const handleBlur = () => {
         const { value } = input
         const formattedValue = formatValue(value)
-        const { dirty, valid } = meta
-        if (dirty && valid && formattedValue !== lastSyncedValue) {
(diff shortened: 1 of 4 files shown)


HendrikThePendric (approved):
Good work. Thanks for explaining the problem so clearly in the PR description.

PR #148, opened 2022-09-08
feat: add completion functionality

Mohammer5:
Completes [TECH-1325](https://dhis2.atlassian.net/browse/TECH-1325)

This PR moves some things:
* `data-workspace/data-value-mutations` -> `shared/data-value-mutations`
  * Some parts of the data value mutations are being re-used in modules other than the data-workspace, so I moved it into the `shared` folder
* `data-workspace/validation` -> `shared/validation` & `bottom-bar/validation-results-sidebar`
* `data-workspace/constants.js` -> `shared/constants.js`
  * I kept the original file as there are some constants that are used by the data-workspace module itself


src/bottom-bar/complete-button.js
```diff
@@ -0,0 +1,92 @@
...
+)
+const completingFormFailedMessage = i18n.t(
+    "Couldn't validate the form, please try again later"
+)
+const incompletingFormFailedMessage = i18n.t(
+    "Couldn't incomplete the form, please try again later"
+)
+
+// @TODO: Get the "isComplete" value from some response?
+// @TODO: Optimistically update the source of the "isComplete" value
```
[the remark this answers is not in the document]
Mohammer5: These have been implemented. I removed the comments

src/bottom-bar/main-tool-bar.js
```diff
@@ -58,7 +54,7 @@ export default function MainToolBar() {
                 </span>
             </button>
 
-            {isComplete && (
+            {data?.completeStatus.lastUpdatedBy && (
```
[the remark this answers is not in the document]
Mohammer5: The `completeStatus` object is always present. If the form never has been completed, the object simply looks like this:

```js
{ complete: false }
```

src/data-workspace/query-key-factory.js
```diff
@@ -41,30 +41,31 @@ export const dataValueContext = {
     ],
 }
 
-export const getValidationMetaDataQueryKey = (datasetId) => {
-    const queryKey = [
-        `validationRules`,
+export const getCompleteFormQueryKey = ({
```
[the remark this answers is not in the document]
Mohammer5: Remnants from the past.. I've removed them

src/shared/validation/use-imperative-validate.js
```diff
@@ -0,0 +1,52 @@
...
+            categoryComboId,
+            categoryOptionIds,
+        })
+        const validationMetaDataQueryKey =
+            getValidationMetaDataQueryKey(dataSetId)
+
+        const vQ = client.fetchQuery(validationQueryKey)
+        const vMDQ = client.fetchQuery(validationMetaDataQueryKey)
+
+        return Promise.all([vQ, vMDQ]).then(
```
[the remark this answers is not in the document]
Mohammer5: Errors are handled by the `<CompleteButton />` so we can show an appropriate message to the user and set the local loading state to false in case of an error

src/shared/validation/query-key-factory.js
```diff
@@ -0,0 +1,46 @@
...
+    dataSetId,
+    periodId,
+    orgUnitId,
+    categoryComboId,
+    categoryOptionIds,
+}) => {
+    const params = {
+        pe: periodId,
+        ou: orgUnitId,
+        // @TODO: Revisit these params to filter by AOC https://dhis2.atlassian.net/browse/TECH-1287
```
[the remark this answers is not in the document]
Mohammer5: Removed

Mohammer5:
> We are completing the form now regardless of whether validation fails or not. But there is a field in maintenance app complete allowed only if validation passes that we should respect. The field does come in the metadata so we should use it, but also ideally, the BE should refuse completing the form (not sure if the complete endpoint calls validation internally or not).

Has been addressed

> another thing, maybe for the future - from my testing, I think the change in the UI from Mark complete to Mark incomplete is too subtle .. it would be good if we display any other message that completion happened successfully - maybe just useAlert with a success background, same as we do when an error happens. cc @cooper-joe

Yes, this should be a separate issues / PR

src/shared/completion/use-set-form-completion-mutation.js
```diff
@@ -0,0 +1,94 @@
...
+                    complete: variables.completed,
+                },
+            }))
+
+            const context = { previousDataValueSet, dataValueSetQueryKey }
+            return context
+        },
+        onError: (event, _, context) => {
+            showErrorAlert(
+                `Something went wrong while deleting the min-max limits: ${event.message}`
```
tomzemp, line 85: wrong message, and should be translated

Mohammer5: Good find, addressed!

src/shared/completion/use-set-form-completion-mutation.js
```diff
@@ -0,0 +1,94 @@
...
+
+            // Snapshot the previous value
+            const previousDataValueSet =
+                queryClient.getQueryData(dataValueSetQueryKey)
+
+            // Optimistically overwrite the completion status
+            queryClient.setQueryData(dataValueSetQueryKey, () => ({
+                ...previousDataValueSet,
+                completeStatus: {
+                    ...previousDataValueSet.completeStatus,
```
tomzemp, line 78: this spread operator seems to leave the lastUpdated and lastUpdatedBy unchanged? Which among other things, leads to the condition in main-tool-bar.js (`data.completeStatus.lastUpdatedBy`) not displaying until the information is refreshed from the backend

tomzemp (commented):
Looks good overall!

I left some small comments. 

I also wanted to point out that the `isLoading` is being set to false in finally() statements, which means that this is just staying in loading state when offline? Could we just put the loader in the button or something, so it doesn't block the rest of the app? (completion is technically available offline in the old app but seems not to work)

On a side note (but not related to your code): I really don't like that we are enforcing `validCompleteOnly` client-side but not server side 😬

src/shared/completion/use-set-form-completion-mutation.js
```diff
@@ -0,0 +1,108 @@
...
+            orgUnitId,
+            categoryComboId,
+            categoryOptionIds,
+            completed,
+        }
+
+        return engine.mutate(MUTATION_SET_FORM_COMPLETION, { variables })
+    }
+
+    const dataValueSetQueryKey = useDataValueSetQueryKey()
```
Birkbjo, line 59: Please take a look how this is done here: https://github.com/dhis2/data-entry-app/blob/development/src/data-workspace/data-value-mutations/mutation-functions.js#L31

We should not need to get the context-params manually again here. Please get them from the `mutationKey`. Also a `mutation-key-factory.js` would be great, to follow the same pattern as `dataValueMutation`. But might not be needed for now.
By getting the parameters from the `mutationKey`, we are guaranteed that the queryKey contains everything that is needed for the mutation, which can prevent subtle bugs. Think of the `mutationKey` as a `dependency`-array to the mutation. 
If you don't think a separate mutationKey is needed here, still try to get the params from the `useDataValueSetQueryKey`. 

```js

//outside hook

function mutateFn({ completed} ) {
        // get params for the mutation from the `mutationKey`
       const { mutationKey } = this
       const { params } = mutationKey[1]
       const variables = {
            dataSetId: params.ds,
            periodId: params.pe,
            orgUnitId: params.ou
            categoryComboId: params.cc
            categoryOptionIds: params.cp
       }
      const {
          ds: dataSetId,
          pe: periodId,
          ou: orgUnitId,
          cc: categoryComboId,
          cp: categoryOptionIds,
       } = mutationKey[1]

        return engine.mutate(MUTATION_SET_FORM_COMPLETION, { variables })
    }

// in hook
   // could create a new hook that returns in non-abbreviated format
   // instead of having to translate this in mutateFn above
const dataValueSetQueryKey = useDataValueSetQueryKey()

return useMutation(mutationFn, {
      retry: 0,
      mutationKey: dataValueSetQueryKey,
      onMutate: ...
```

Is this functionality available offline? If so, we should extract the `mutationFn` (like above) to a separate function and setup a `defaultMutationFn` in `use-query-client`. Similar to how it's done for `dataValues`.

src/shared/completion/use-set-form-completion-mutation.js
Birkbjo: Yeah, ideally the backend would return the updated values in the response to the POST. But I guess we could create our own object in the meantime? 

If not, it would also be great to fetch this from the backend without the `dataValues`, so we could invalidate just this information, by doing a simple fetch to eg. `dataEntry/dataSetCompletion` (with params for the selection of course).

Birkbjo (changes requested):

Birkbjo:
> I also wanted to point out that the isLoading is being set to false in finally() statements, which means that this is just staying in loading state when offline? Could we just put the loader in the button or something, so it doesn't block the rest of the app? (completion is technically available offline in the old app but seems not to work)

I think it should be possible to support this offline. We would maybe need a confirmation to send the completion once that user comes online again, though?

Right now, clicking `incomplete` while offline creates a forever-loading spinner and makes the app unusable.

src/shared/completion/use-set-form-completion-mutation.js
Mohammer5: This has been addressed by both @Birkbjo as well as @kabaros

src/shared/completion/use-set-form-completion-mutation.js
Mohammer5: Would you mind if we address this in a separate issue / PR?

.github/workflows/dhis2-verify-app.yml
```diff
@@ -114,7 +114,7 @@ jobs:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
                   CYPRESS_dhis2BaseUrl: https://debug.dhis2.org/dev
-                  CYPRESS_dhis2ApiVersion: 39
+                  CYPRESS_dhis2ApiVersion: 40
```
Birkbjo, line 117: Do we actually want this, should we not run the fixtures against `2.39/dev`?

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -0,0 +1,166 @@
...
+                            )
+                        ) {
+                            setRightHandPanel(validationResultsSidebarId)
+                        }
+                    }
+                )
+                .catch((e) => {
+                    console.error(e)
+                    // For now this can fail "silently".
+                    // ES6 Promises don't have a mechanism for reporting
```
Birkbjo, line 102:
```js
Promise.allSettled() 
```
Would do this, but I'm not sure if it's more appropriate, this solution  looks fine to me.

src/shared/validation/use-imperative-validate.js
```diff
@@ -0,0 +1,52 @@
...
+ **/
+export default function useImperativeValidate() {
+    const client = useQueryClient()
+    const [dataSetId] = useDataSetId()
+    const [periodId] = usePeriodId()
+    const [orgUnitId] = useOrgUnitId()
+    const {
+        attributeCombo: categoryComboId,
+        attributeOptions: categoryOptionIds,
+    } = useApiAttributeParams()
```
Birkbjo, line 27:
```js
    const [{ dataSetId, orgUnitId, periodId }] = useContextSelection()
    const {
        attributeCombo: categoryComboId,
        // Note these are sorted:
        attributeOptions: categoryOptionIds,
    } = useApiAttributeParams()

```

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -0,0 +1,166 @@
...
+                        throw new Error(completingFormFailedMessage)
+                    })
+                }
+            })
+}
+
+function useOnCompleteWhenValidNotRequiredClick() {
+    const { mutateAsync: setFormCompletion } = useSetFormCompletionMutation()
+    const setRightHandPanel = useSetRightHandPanel()
+    const validate = useImperativeValidate()
```
Birkbjo, line 76: Is there a reason this needs to be imperative?

Why can't we do a regular `useQuery` here, with a `onError` ? The error is ignored below anyway, and the `setFormCompletion` does not seem to be dependant on the return-value of this?

There's some benefits to it I think, like react-query handling when the component is unmounted, so it doesn't call `onSuccess` etc.

Birkbjo: We might want to just keep it like this for now though...

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -0,0 +1,166 @@
...
+        } else if (validCompleteOnly && !offline) {
+            promise = onCompleteWhenValidRequiredClick()
+        } else {
+            promise = onCompleteWhenValidNotRequiredClick()
+        }
+
+        return (
+            promise
+                // this will eventually catch any error thrown and display the error message
+                .catch((e) => showErrorAlert(e.message))
```
Birkbjo, line 163: I think this should be handled on `onError` to `useMutation`. 

We could also look into having a global handler for this.

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -0,0 +1,166 @@
...
+            cancelCompletionMutation()
+        }
+
+        let promise
+        if (isLoading && offline) {
+            // No need to complete when the completion request
+            // hasn't been sent yet due to being offline.
+            // It's important to still perform the request online as we don't
+            // know if the mutation actually reached the server already
+            return Promise.resolve()
```
Birkbjo, line 150: Not sure I fully understand this. Is the mutation actually sent here? The request was just cancelled above, we should prepare a mutation in this case, no?

src/shared/completion/use-imperative-cancel-completion-mutation.js
```diff
@@ -0,0 +1,38 @@
...
+        // undo the optimistic update
+        // @TODO: Why does this cause a request to the data values that will
+        //   end up with a 404 due to being offline while the `onMutate`
+        //   callback in the `useSetFormCompletionMutation` does not?
+        const { completed } = foundMutation.options.variables
+
+        const previousDataValueSet =
+            queryClient.getQueryData(dataValueSetQueryKey)
+
+        queryClient.setQueryData(dataValueSetQueryKey, () => ({
```
Birkbjo, line 30: https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetquerydata
This gets passed the old-data.

```js
queryClient.setQueryData(dataValueSetQueryKey, ({ completed: prevCompleted} ) => ({
```

src/bottom-bar/complete-button.js
```diff
@@ -0,0 +1,75 @@
...
+            cancelCompletionMutation()
+        }
+
+        if (isLoading && offline) {
+            // There's no need to incomplete the form when it hasn't been
+            // completed yet. It's important to still perform the request
+            // online as we don't know if the mutation actually reached the server
+            // already
+            return
+        }
```
Birkbjo, line 45: I don't think there should be any difference between "incomplete" and "complete". This is just passed in as a variable to the mutation-function right?

You shouldn't handle these different, the previous state shouldn't really matter other than sending `complete: false` or `complete: true` to the mutate-fn.

src/bottom-bar/complete-button.js
```diff
@@ -0,0 +1,75 @@
...
+
+export default function CompleteButton({ disabled }) {
+    const { offline } = useConnectionStatus()
+    const setFormCompletionMutationKey = useSetFormCompletionMutationKey()
+    const isLoading = useIsMutating(setFormCompletionMutationKey)
+    const { show: showErrorAlert } = useAlert((message) => message, {
+        critical: true,
+    })
+
+    const dataValueSet = useDataValueSet()
```
Birkbjo, line 28: Let's use `useValueStore`. It's more efficient. 

```js
const hasData = useValueStore(state => !!state.getDataValues())
// maybe create a selector for `getDataValueSet()` in data-value-store.js
const isCompleted = useValueStore(state => state.dataValueSet.completeStatus)
```

Birkbjo (changes requested):

.github/workflows/dhis2-verify-app.yml
Mohammer5: Hmm I think 2.40 should be fine. It should have the same endpoints (- it does afaik) and once we release these should point to 2.40 anyway. I wouldn't mind changing it, do you see an advantage in setting them back to 39?

src/bottom-bar/use-on-complete-callback.js
Mohammer5: No mutation is being sent. When the user is offline and we have a pending mutation, instead of adding a second pending request, we can simply cancel the pending request:

pending-complete-request -> pending-incomplete-request ---> status: incomplete
pending-complete-request -> cancel request ---> status: incomplete

src/bottom-bar/complete-button.js
Mohammer5: Yeah, the `useSetFormCompletionMutation` hook is re-used by both the `onCompleteClick` and the `onIncompleteClick` handlers.

The difference between them is which requests should be fired alongside. When incompleting, there's never another request being sent. When completing, then in two cases we also send the validation request, hence two different click handlers

src/bottom-bar/use-on-complete-callback.js
Birkbjo: Yeah, then I think the `cancelCompletionMutation()` above should be move into this if-test.

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -0,0 +1,166 @@
...
+    const onCompleteWhenValidRequiredClick =
+        useOnCompleteWhenValidRequiredClick()
+    const onCompleteWhenValidNotRequiredClick =
+        useOnCompleteWhenValidNotRequiredClick()
+    const onCompleteWithoutValidationClick =
+        useOnCompleteWithoutValidationClick()
+
+    return () => {
+        if (isLoading) {
+            cancelCompletionMutation()
```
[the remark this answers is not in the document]
Mohammer5: That's a good question.. I'm not really sure tbh, I didn't even find docs for `MutationCache.remove` method, just copied from existing code..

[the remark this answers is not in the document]
Mohammer5: Seems to work.. At least, when I click on the button twice (to run this part of the code) and then go online again, nothing is being sent to the server

src/bottom-bar/use-on-complete-callback.js
Mohammer5: Let's do this in another PR so we can get this merged today

src/bottom-bar/use-on-complete-callback.js
Mohammer5: :ok_hand:, done

src/bottom-bar/use-on-complete-callback.js
Mohammer5: https://dhis2.atlassian.net/browse/TECH-1434

.github/workflows/dhis2-verify-app.yml
Birkbjo: Should be fine, but it can be unstable. `2.39dev` should really be what we run these tests towards, since that's the version we are targeting currently.

Birkbjo (approved):
Thanks a lot for this! And for tolerating all of my comments.
Looks very good!

PR #151, opened 2022-09-08
revert: enable validation for min/max limits on generic-input

DNR800:
Reverts changes on https://github.com/dhis2/data-entry-app/pull/144

This arouse out of the [discussion on slack](https://dhis2.slack.com/archives/C02NHB92S2H/p1662655152593399)

There was a performance issue introduced as part of https://github.com/dhis2/data-entry-app/pull/144, this pr gave the option to revert that change

This reverts commit ffc94ae3afee0c466931b55d457f654faf5169f5.

If https://github.com/dhis2/data-entry-app/pull/153 is accepted and merged this PR can be closed

PR #153, opened 2022-09-11
fix: minMaxValues performance

Birkbjo:
Fixes re-renders from `useMinMaxLimits` by using a [zustand](https://github.com/pmndrs/zustand) store.

Note that we are not cloning state here. We are using the same object-references as `tanstack/query`, so the memory footprint should be minimal.  

- (+) Way less boilerplate than react contexts
        - Don't have to use confusing "setter"-context to split value from setter
- **(+)Selectors to subscribe to just the state you want**
- (+)It's just a hook, API is minimal and ergonomic
- (+) Same object-reference in store, no duplicated data
- (+) No need to manually sync fields
- (-) It's a new library

### Why did I end up using a new library?

The main goal was to make it easy for components across the app to subscribe to a single data value, without re-rendering when another value is updated. It should also be possible for all cells to subscribe without causing re-renders or other performance issues. 

I've used parts of the weekend to research this, and this was the best solution I've come up with. Selectors in useQuery (https://github.com/dhis2/data-entry-app/pull/154) seemed very promising, but the stutters made me abandon that solution after some testing. 

After some brief discussion with @HendrikThePendric , my initial approach was to try to use the react-final-form [mutators](https://final-form.org/docs/final-form/types/Mutator). That way, we could imperatively update the `meta.data`-field for a single field. I was a bit skeptical to that solution, but it had some benefits, like not needing to subscribe to anything in each field. But it also had some major drawbacks; we would have to manually keep the `data`-field up to date when changing a comment or minMax-value. We would also need to have a separate context to store the `form`-reference, or somehow make the sidebar be able to update the fields in the data-value-form. However, what made me seek other solution was that there was no slick way to integrate it with `createInititalValues`, since this happens "before" the form has a reference to the instance. There's probably way around this (eg. using [createForm](https://final-form.org/docs/final-form/api#createform), keeping a ref, or moving that logic to another component down the tree. 

However, I found that it would be hard to follow that approach, and sought other solutions. I've briefly looked into `zustand` before, and I've mentioned it at least once in some early frustrations with `react-query`. So I did a quick test, and it worked really well out of the box! I actually tried this before learning about the react-query-selector, and I was really optimistic about that solution before learning the performance issues with it. Since I know there might be some resistance of introducing a new "state manager", but this is so minimal and beneficial that I feel like it's worth it. 

I think this can help us clean up some confusing parts with the app and the many contexts, and at least help with some of the performance issues we are having. 

This library is awesome, and I think we should use this in-place of most contexts. That might be controversial, since it's a new library for us, and "another" state manager. So It might be something we could discuss in a frontend meeting. But it solves so many problems, and it's so easy to use! It also has a great reputation.


DNR800:
I actually really like this. And I like zustand store provides without having to write code that would feel like boiler plate.

I know that the comments were a similar problem, could the comments be potentially folded into the same solution? (I'm unfamiliar with the comments implementatin at this point) I would happily look at getting that introduced as well.

In comparing behaviour with whats currently on development I think I'm seeing that we are loosing the the initial form validation that we do on page load. Might have to consider how that is reinstated too

HendrikThePendric (approved):
Even at 6x CPU throttle the app is behaving well after this PR. Awesome. And as @Mohammer5 also hinted, there is much more to optimise. Great work @Birkbjo, thank you.

HendrikThePendric:
For the record, by now the addition of the zustand lib has been discussed with most/all team members, and there were no objections. The consensus was that this app would really benefit from having a state manager that allows for subscribing to slices of the state and derived state. It will allow us to address some performance issues and clean up the codebase too.

DNR800 (approved):
I think this looks good. As we are going with this I'll close https://github.com/dhis2/data-entry-app/pull/151

PR #154, opened 2022-09-11
fix(use-data-value-set): add a hook to get dv-data with selector

Birkbjo:
A solution for `minMaxValue`-rerenders, and a way to only subscribe to a single value from the `useQuery`. Was the solution for subscribing to a single value just infront of us all this time, and we just didn't know that `select` would only subscribe to what was returned? Kind of, but it has big performance issues by itself. 

It's interesting to me that this functionality is barely mentioned in a secondary sentence in [this blog post](https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option). I've researched a lot, and never seen a mention of that functionality in select. 

This seemed like it would be the perfect solution, but the app "freezes" for 1-2seconds when posting a value. Seems like the algorithm they're using to check if the value changes is not very efficient. I've also tried setting `structuralSharing: false`, as I've read this can cause performance issues, but didn't make much of a difference. This prevent re-rendering of all fields, but the app still stutters when a value is synced.

Anyway, I would prefer the [zustand solution](https://github.com/dhis2/data-entry-app/pull/153), since the performance using that is a lot better. I just wanted to show that this is possible, and we might want to investigate if we could get this to work instead - and it could potentially be useful for a few components, but seems like rendering this in every field in a big form (Child Health) causes a lot of stuttering when changing a value (after value is successfully synced).

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2022-09-07T10:30:31.473Z\n"
-"PO-Revision-Date: 2022-09-07T10:30:31.473Z\n"
+"POT-Creation-Date: 2022-09-11T14:24:30.880Z\n"
+"PO-Revision-Date: 2022-09-11T14:24:30.880Z\n"
 
 msgid "Not authorized"
 msgstr "Not authorized"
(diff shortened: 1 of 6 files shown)


HendrikThePendric (commented):
I like this solution. Have tested it and as you said, the app still stutters a bit when a value is updated. Furthermore, this solution would open up a way to give field-level access to most, but not all information we need. Specifically, we should still find a way to optimise the highlighted field feature, but we would be able to access follow-up, comment, and min/max.

You mentioned you would prefer the zustand solution, and since I also prefer to add some global state management library to this app, I am inclined to agree.

Birkbjo:
Closed in favor of #153

PR #155, opened 2022-09-12
fix: highlighted fields performance

Birkbjo:
Refactors the HighlightedField-feature to use zustand store instead of context. This prevents re-rendering of all fields when focusing a cell, and only the fields that are focused/defocused are re-rendered. 
I've also refactored some code since I had to touch it anyway (like the `mapDataValuesToFormInitialValues`-function.


src/shared/highlighted-field/use-highlighted-field.js
```diff
@@ -40,26 +46,27 @@ function gatherHighlightedFieldData({ de, coc, dataValueSet }) {
...
+        state.getDataValue({
+            dataElementId: item?.dataElementId,
+            categoryOptionComboId: item?.categoryOptionComboId,
+        })
+    )
+    const { data: metadata } = useMetadata()
+    const dataElement = selectors.getDataElementById(
+        metadata,
+        item?.dataElementId
+    )
```
Birkbjo, line 60: I thought the way the `datElement` was attached to the `item` in the context was quite confusing. So I find it's cleaner to just store the ID, and then we grab the dataElement here.

src/shared/use-data-value-set/use-data-value-set.js
```diff
@@ -11,38 +11,18 @@ function mapDataValuesToFormInitialValues(dataValues) {
...
-            }
-
-            if (!acc[dataElement]) {
-                acc[dataElement] = { [categoryOptionCombo]: dataValueData }
-            } else {
-                acc[dataElement][categoryOptionCombo] = dataValueData
+    const formInitialValues = dataValues.reduce((acc, dataValueData) => {
+        if (!acc[dataValueData.dataElement]) {
+            acc[dataValueData.dataElement] = {
+                [dataValueData.categoryOptionCombo]: dataValueData,
```
Birkbjo, line 17: I didn't really understand why some of the fields were filtered out. I think it's much clearer when this contains the exact data we have in the cache/get from the API.

src/data-workspace/data-entry-cell/entry-field-input.js
```diff
@@ -91,8 +90,21 @@ export function EntryFieldInput({
...
+
+    useEffect(() => {
+        return () => {
+            if (highlighted) {
+                setHighlightedFieldId(null)
+            }
+        }
+        // only used to cleanup on unmount - cannot include deps here
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [])
```
Birkbjo, line 107: Want some input on this.
I've noticed `setHighlightedField(null)` was called in different components around the app. But isn't it easier to just set this null when the component unmounts? I think we would want to reset this whenever the field that is highlighted disappears from the screen, right?

src/shared/use-context-selection/use-manage-inter-param-dependencies.js
```diff
@@ -42,7 +42,9 @@ function useHandleDataSetIdChange() {
         organisationUnits: assignedOrgUnits,
         periodType: dataSetPeriodType,
     } = selectors.getDataSetById(metadata, dataSetId) || {}
-    const setHighlightedFieldId = useSetHighlightedFieldIdContext()
+    const setHighlightedFieldId = useHighlightedFieldStore(
+        (state) => state.setHighlightedField
+    )
```
Birkbjo, line 47: I think we can remove this, if you agree about my comment above (regarding resetting on umount).

src/data-workspace/data-entry-cell/entry-field-input.js
HendrikThePendric: As far as I can tell, there are 3 places where we set the highlighted field to null. In 1 case we are setting it to null when the dataset changes, and this could be replaced by setting it to null when the components unmounts. However, in the other two cases, we are setting the highlighted field to null because a filter input is getting focus (see discussion [here](https://dhis2.slack.com/archives/C02NHB92S2H/p1662636916927349?thread_ts=1662630289.253189&cid=C02NHB92S2H)). In these cases the entry field won't unmount.

To conclude: I actually think we can keep this part as it is. It makes sense to have it there.

src/shared/use-context-selection/use-manage-inter-param-dependencies.js
HendrikThePendric: I don't mind removing it from here, if you insist. But on the other hand, I also think that doing this here actually makes quite a lot of sense too. If we set it to null when the entry field unmounts we'd have to do so conditionally, if `highlighted` is true.

src/shared/use-data-value-set/use-data-value-set.js
HendrikThePendric: Agreed.

src/data-workspace/data-entry-cell/entry-field-input.js
Birkbjo: Yeah, it depends on what kind of functionality we want. I could see that it would be beneficial to keep the `highlighted`-field when filtering fields, and if you filter that field away - it would loose focus this way.

I also think it's "safer", in the sense that we might implement something later, or haven't thought of a navigation that should reset the highlighted-field, but doesn't now.

src/data-workspace/data-details-sidebar/basic-information-follow-up-button.js
```diff
@@ -1,17 +1,15 @@
...
 import { Button, colors, IconFlag24, Tooltip } from '@dhis2/ui'
 import React from 'react'
-import { useHighlightedField } from '../../shared/index.js'
 import { useSetDataValueMutation } from '../data-value-mutations/data-value-mutations.js'
 import ItemPropType from './item-prop-type.js'
 
 const FollowUpButton = ({ item }) => {
-    const { value } = useHighlightedField()
-    const isEmptyField = !value
+    const isEmptyField = !item?.value
```
DNR800, line 8: I changed this as `item` and `value` were the same object. `useHighlightedField()` is used in \<DataDetailsSidebar \/\> and it passes its item down through props

https://github.com/dhis2/data-entry-app/blob/d2745e7dfd83a62d04ec7a0399a3c2c3a7619a69/src/data-workspace/data-details-sidebar/data-details-sidebar.js#L16-L34

src/shared/highlighted-field/use-highlighted-field.js
```diff
@@ -15,51 +22,53 @@ function gatherHighlightedFieldData({ de, coc, dataValueSet }) {
...
-    const { data: dataValueSet } = useDataValueSet()
-    const { item } = useHighlightedFieldIdContext()
-    const [currentItem, setHighlightedFieldId] = useState(() => {
-        const { de, coc } = item
-        return gatherHighlightedFieldData({ de, coc, dataValueSet })
-    })
+    const item = useHighlightedFieldStore((state) =>
+        state.getHighlightedField()
+    )
+    const dataValue = useValueStore((state) =>
```
DNR800, line 52: This is more a general point rather than just for this PP, but I find myself getting confused around the word `item` its used very interchangable and can mean very different things in different parts of the code. Example here

https://github.com/dhis2/data-entry-app/blob/d2745e7dfd83a62d04ec7a0399a3c2c3a7619a69/src/data-workspace/data-details-sidebar/data-details-sidebar.js#L16-L34

Made we should try not to use the work item in the code in future? its such a generic term its hard to understand what it s mean to be sometimes

src/shared/stores/highlighted-field-store.js
```diff
@@ -0,0 +1,25 @@
...
+     * @returns
+     */
+    setHighlightedField: (item) => set({ item }),
+    getHighlightedField: () => get().item,
+    isFieldHighlighted: ({ dataElementId, categoryOptionComboId }) => {
+        const item = get().item
+        if (!item) {
+            return false
+        }
+        return (
```
DNR800, line 20: I think if you flip the check for `if(!item)` that you might just be able to do 
```
return (
     !!item &&
     item.dataElementId === dataElementId &&
     item.categoryOptionComboId === categoryOptionComboId
   )
```

src/shared/highlighted-field/use-highlighted-field.js
Birkbjo: Yeah, I agree. I renamed it to `highlightedFieldId`.

src/data-workspace/data-entry-cell/entry-field-input.js
Mohammer5: @HendrikThePendric I've read through the thread you've linked. Right now I don't understand how you came to the following conclusion:

> Unsetting the active cell when a filter input gets focus should be fairly trivial.

I can't find anything related to this in the messages above. To me, it's quite weird that we hide the sidebar / unset the highlighted item when we focus the highlighted field. What if I just opened the sidebar and then want to filter the results in such a way that the previously highlighted field is still highlighted? Generally it's feels odd that something is happening in the UI just by focussing a search input (in contrast to "changing the filter" value; if we unset the highlighted field when changing the value from an empty string to a non-empty string, that'd be something else)

Birkbjo: > To me, it's quite weird that we hide the sidebar / unset the highlighted item when we focus the highlighted field.

I guess you mean the `filter field`-here?

I kind of agree, but I do think it's "safer" to clear this when the field unmounts. In the current implementation, we are using `display: none` - so the use-case you highlight (no pun intended) would actually work. I'm not sure if it's actually something we want though - since it could be kind of confusing to have a highlighted-field that is not visible in the form?

Mohammer5 (approved):

src/data-workspace/data-entry-cell/entry-field-input.js
Mohammer5: From slack:

> **Hendrik de Graaf**
> Does this help: https://dhis2.slack.com/archives/C02NHB92S2H/p1662636916927349?thread_ts=1662630289.253189&cid=C02NHB92S2H
>
> > Joe Cooper
> > I think this is almost there. I agree with the premise, and think it's a nice solution. Just testing it though, it's a bit odd when I'm tabbing through a form with sections. The first input in a section is a filter input, so the last input in the previous section gets the grey800 border while the filter input in the next section is active. 

I see, it's about tabbing through the form. IMO we shouldn't unset the active field when focusing the input field because of this:

> To me, it's quite weird that we hide the sidebar / unset the highlighted item when we focus the highlighted field. What if I just opened the sidebar and then want to filter the results in such a way that the previously highlighted field is still highlighted?

It would get tricky to differentiate different reasons why the filter input got focused. And for that reason my preference would be to simply live with the fact that two fields can be highlighted at the same time. If @Birkbjo's current approach keeps the sidebar open & the highlighted item in the state, then I'd prefer to resolve this discussion

cc @cooper-joe I think we need input from you, too

HendrikThePendric: If unsetting the highlighted field when focussing on a filter input makes the whole sidebar showing/hiding unnecessarily complicated, then I would also be in favour of just skipping that feature. It was more of a "nice to have" feature to begin with.

@cooper-joe I was thinking we could lighten the outline color of the highlighted-cell-without-focus a bit to make it a bit easier for the user to see that he/she currently has focus on the filter-input, and not the highlighted cell?

cooper-joe: Yeah, I see the dilemma here. Each option has a downside, so we need to choose which is less-bad. Each option presented as user feedback:
1.  "Why does the last cell in a section have a black border when I tab to the next section? It's confusing."
2.  "Why does the details sidebar close when I tab to the next section? It slows down my workflow." (This PR)

It would be good to implement this in a way that we avoid both of those downsides, but I understand that introduces complexity. So, for now, I'd say living with downside no.1 is less-bad than downside no.2, because downside no.2 could really slow down workflows. So, I suggest dropping the "unset highlighted field when focussing filter" functionality, for now.

There are also other cases that @Birkbjo highlights above, like having a highlighted/focussed field that gets filtered out of the table, so we should ideally land on a solution that solves for all those difficulties too.

src/shared/use-context-selection/use-manage-inter-param-dependencies.js
tomzemp: This seems no longer necessary as you are setting the highlighted field to null on EntryField unmount, conditional on `highlighted` being true

src/bottom-bar/bottom-bar.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react'
-import { useHighlightedFieldIdContext } from '../shared/index.js'
+import { useHighlightedFieldStore } from '../shared/index.js'
 import styles from './bottom-bar.module.css'
 import DataItemBar from './data-item-bar.js'
 import MainToolBar from './main-tool-bar.js'
 
 export default function BottomBar() {
-    const { item } = useHighlightedFieldIdContext()
+    const item = useHighlightedFieldStore((state) =>
```
tomzemp, line 8: I feel like the choice of the word `item` here is now particularly confusing because you are only getting the highlighted field id at this point, which is less clear because `useHighlightedFieldStore` as a name makes it less clear than `useHighlightedFieldIdContext` that the store contains IDs and not the full item information (i.e. from use-highlighted-field). I'd suggest renaming here, or renaming the store

tomzemp (approved):
looks good to me, though I left two small suggestions

src/bottom-bar/bottom-bar.js
Birkbjo: Yeah I found naming this was particularly difficult. We refer to `${dataElementId}.${cocId}` as `fieldId`, and this store stores those ids in an object. I did initially just want to call this `field` or `highlightedField`- but it's not really descriptive either. Then I renamed the item in the store to `highlightedFieldId`, due to a comment above. But I'm not sure I like that either, since `id` have a strong connotation to the value being a string. So I think I would prefer `highlightedField` now, even if `field` doesn't really describe what it is.

tomzemp: Yeah. I think `highlightedField` would be better 👍

PR #156, opened 2022-09-12
fix: prevent refetch after post

HendrikThePendric:
Adding the `staleTime` is the real fix. Adding `useMemo` to the `useDataValueSetQueryKey` doesn't really fix anything, but could prevent issues in the future since having a stable query key for a single selection makes sense.


HendrikThePendric:
Closing this one in favour of #159

PR #157, opened 2022-09-12
fix(min-max): fix delete parameters

tomzemp:


tomzemp:
closed in favour of https://github.com/dhis2/data-entry-app/pull/160

PR #159, opened 2022-09-12
fix: prevent the form from being re-fetched on every data value change

kabaros:
fixes [TECH-1410](https://dhis2.atlassian.net/browse/TECH-1410)

### DIFF
--- a/src/shared/use-data-value-set/use-data-value-set.js
+++ b/src/shared/use-data-value-set/use-data-value-set.js
@@ -1,4 +1,4 @@
-import { useQuery, useIsMutating } from '@tanstack/react-query'
+import { useQuery } from '@tanstack/react-query'
 import { createSelector } from 'reselect'
 import { useIsValidSelection } from '../use-context-selection/index.js'
 import useDataValueSetQueryKey from './use-data-value-set-query-key.js'
@@ -73,12 +73,10 @@ const select = createSelector(
 export const useDataValueSet = ({ onSuccess } = {}) => {
     const isValidSelection = useIsValidSelection()
     const queryKey = useDataValueSetQueryKey()
-    const activeMutations = useIsMutating({ mutationKey: ['dataValues'] })
 
     const result = useQuery(queryKey, {
-        // Only enable this query if there are no ongoing mutations
         // TODO: Disable if disconnected from DHIS2 server?
-        enabled: activeMutations === 0 && isValidSelection,
+        enabled: isValidSelection,
         select: select,
         // Fetch once, no matter the network connectivity;
         // will be 'paused' if offline and the request fails.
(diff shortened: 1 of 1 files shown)


HendrikThePendric (approved):
I've tested and discussed this with @kabaros on Slack. This solution delivers the desired results and is less invasive and seems more idiomatic than https://github.com/dhis2/data-entry-app/pull/156/files I produced earlier. So I will close mine in favour of this one.

tomzemp:
I think the overall approach is good, but wonder if it should just be `refetchOnReconnect: false`? It seems like the query will be blocked from refreshing if the context is invalid by `enabled: isValidSelection` and then I don't know if there's an advantage to only refreshing when we come online and there are no outstanding mutations (that is, if we do not refresh when there are outstanding mutations, why do we refresh when there are no outstanding mutations)?

HendrikThePendric:
> I think the overall approach is good, but wonder if it should just be `refetchOnReconnect: false`? It seems like the query will be blocked from refreshing if the context is invalid by `enabled: isValidSelection` and then I don't know if there's an advantage to only refreshing when we come online and there are no outstanding mutations (that is, if we do not refresh when there are outstanding mutations, why do we refresh when there are no outstanding mutations)?

It could well be that using a static `false` value would be just as good. At least the for the form values. The condition has been put there by @KaiVandivier for a specific reason:
>  when offline a lot of request are queued and by default they’d all be fired in parallel when coming online. And for this particular request that can cause issues, because the GET won’t include the POSted values yet. So then the local cache will be overridden by the stale values

But even if there are no pending mutations and a GET does happen, I don't think the form values are actually updated based on the response of this request, because these now live in the RFF form state. However, we also have to deal with other types of data like comments, follow-up and min/max values. So in theory, if comment is added by another user and the app comes back online, without having any pending mutations, the new comment should be usable in the app.

While writing the above, I started to realise this is quite complex behaviour, and quite unpredictable, because there are various scenarios in which we have slightly different behaviour. Perhaps to keep things fairly easy to reason about it would make more sense to just use `false`.

Using `false` or using `() => activeMutations === 0 && isValidSelection` for `refetchOnReconnect` are both fine with me.

Birkbjo:
This does **not** guarantee that the update to the cache will be sent after mutations. In fact, it always sends a GET-request to `/dataValues` before mutations, when going offline -> online. 

I'm pretty sure this will cause data-anomalies. Yes, the "mutations" will not be affected, and update the data on the server - but now you're in a sitatuins where your client cache and server data are conflicting. You will could for instance add a comment offline -> go online -> data is refetched, comment is reset to server -> mutation is sent and updated -> comment is still showing old value since no /dataValues request will update this after mutations. 

I think we should prevent fetching of data as long as there are outstanding mutations.

PR #162, opened 2022-09-13
fix(org unit selector preloading): respect default setting for offline org unit level

Mohammer5:
* If the org unit hierarchy's offline levels are set to `Default`, then the selector will now preload all levels from the user's root down to the level specified in the system settings app (which is `level: 3` in our SL db)
* If there's an error while preloading the org units in any way, we show a warning alert bar to the user

NOTE: When using the admin user, this change will cause the app to do >2k requests! I don't know any other way around this other than modifying the org unit tree component (which is somewhat time consuming as well)

### DIFF
--- a/src/context-selection/org-unit-selector-bar-item/get-offline-levels-to-load.js
+++ b/src/context-selection/org-unit-selector-bar-item/get-offline-levels-to-load.js
@@ -1,11 +1,11 @@
 /**
- * @param {Object} data
- * @param {Object} data.organisationUnitLevels
- * @param {Integer} data.organisationUnitLevels.level
- * @param {Integer} data.organisationUnitLevels.offlineLevels
- * @param {Object} data.userOrganisationUnits
- * @param {String} data.userOrganisationUnits.id
- * @param {Integer} data.userOrganisationUnits.level
+ * @param {Object} options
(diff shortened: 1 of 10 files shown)


HendrikThePendric:
I looked at the old app, and that is "only" pre-fetching 166 orgunits. I understand that due to the nature of how our org-unit-tree works, this would take a lot of requests (perhaps even 166). But I *don't* get why we would be doing 2k requests....

Mohammer5:
> I looked at the old app, and that is "only" pre-fetching 166 orgunits. I understand that due to the nature of how our org-unit-tree works, this would take a lot of requests (perhaps even 166). But I don't get why we would be doing 2k requests....

@HendrikThePendric This is because the org unit tree does a single request per org unit. It'll start with the root org unit, fetch its children and load the data of each of these and repeat the process for every org unit loaded until the offline level has been reached.

> I can imagine that introducing this feature is more harmful than omitting it from the app

I suppose most users using this app don't have access to the root org unit so it wouldn't be as bad as when using an admin user. But ultimately that's something only @larshelge can answer

Mohammer5:
@HendrikThePendric has implemented this in a different, better way

PR #163, opened 2022-09-13
refactor: comments

DNR800:
Looking to apply the zustand state management to comments handling which means we can remove the Provider/Context code


src/shared/stores/data-value-store.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+    getHighlightedField: () => get().highlightedField,
+    isFieldHighlighted: ({ dataElementId, categoryOptionComboId }) =>
+        get().highlightedField?.de?.id === dataElementId &&
+        get().highlightedField?.coc?.id === categoryOptionComboId,
+    setHighlightedField: (value) => {
+        set({ ...get(), highlightedField: value ?? null })
+    },
+    getHighlightedFieldData: () => {
+        const highlightedField = get().getHighlightedField()
```
DNR800, line 32: Having a centralised store means that you can have more complex selectors - I think there are a couple of places where we "munge" the highlighted item data and data value so potentially you could do this in the selector

Birkbjo: I think we should keep them separate, and let the `dataValue`-store be an exact reference to the same object as RQ-return data.

src/shared/stores/data-value-store.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+    hasComment: ({ dataElementId, categoryOptionComboId }) => {
+        const dataValue = get().getDataValue({
+            dataElementId,
+            categoryOptionComboId,
+        })
+
+        return !!dataValue?.comment
+    },
+    setComment: ({ dataElementId, categoryOptionComboId, comment }) => {
```
Birkbjo, line 79: I don't think we should update this store - then we don't share the object-reference anymore.
Also this is done in a mutation, and optimistically updated - so you should be able to just get this from the cache - without having to set it yourself.

src/shared/stores/data-value-store.js
DNR800: Ok I'll drop the commits for HighlightedField

src/shared/stores/data-value-store.js
DNR800: Have remove the use of `setComments`, the old functionality for this (in the provider that has been deleted on this PR) where originally updating a lookup object. checking on the at the dataValue object directly means we don't have to keep anything else up to date.

Birkbjo (approved):
So much cleaner! LGTM!

PR #165, opened 2022-09-13
fix: adjust client time to server timezone when computing date ranges

HendrikThePendric:
This PR ensures that we are using dates that are corrected for the server time zone when needed. Please note that a large chunk of this code comes from the approval app. Probably we will need to move this into its own repo, but for now I have just copied things over. So when reviewing, please ignore these files:
- `src/shared/use-server-time-as-local/use-server-date-time-as-local.js`
- `src/shared/use-server-time-as-local/use-server-date-time-as-local.test.js`


src/shared/use-server-time-as-local/use-server-date-time-as-local.js
```diff
@@ -0,0 +1,21 @@
...
+    const { systemInfo } = useConfig()
+
+    if (!dateTime) {
+        return undefined
+    }
+
+    const localNow = new Date()
+    localNow.setMilliseconds(0)
+    const nowAtServerTimeZone = new Date(
+        localNow.toLocaleString('en-US', {
```
tomzemp, line 13: For practical purposes, I think this is okay, but the date parsing is implemented by browser and is not guaranteed to work on non-ISO formats (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse). I think we can just live with this as it seems major browsers parse dates well, but we could also select a locale that is closer to ISO format (e.g. 'sv-SE') if we wanted

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
```diff
@@ -0,0 +1,21 @@
...
+            timeZone: systemInfo.serverTimeZoneId,
+        })
+    )
+    const timestamp = new Date(dateTime).getTime()
+    const timeOffset = localNow.getTime() - nowAtServerTimeZone.getTime()
+
+    return new Date(timestamp + timeOffset)
```
tomzemp, line 20: I think this should be `return new Date(timestamp - timeOffset)`

above you are calculating `offset = local - server`
then in this line you are creating the new date as `local + offset` or `local + (local - server)`
if you make this subtraction, it will be `local - (local - server) = server`

(Alternately, you could change the offset calculation to be `offset = server - local`)

tomzemp (changes requested):
Looks good on the overall approach @HendrikThePendric!

I left some comments (the only one that I this is particularly important is the one on line 20 of use-server-date-time-as-local.js)

There are also a couple of other places in the code where getCurrentDate should be swapped with useNowAtServerTimezone() [these affect the year boundary, e.g. if I log in 1 January 2023 in Addis, but server time is 31 December 2022, the year should be 2022] [I don't know if this is an issue that would in reality ever cause problems because I don't imagine people are doing much data entry on 1 January or 31 December]:
https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/context-selection/period-selector-bar-item/period-selector-bar-item.js#L38
https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/context-selection/period-selector-bar-item/period-selector-bar-item.js#L64
https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/context-selection/period-selector-bar-item/use-periods.js#L24
https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/shared/fixed-periods/fixed-periods.js#L168
https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/shared/fixed-periods/fixed-periods.js#L886

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
```diff
@@ -0,0 +1,21 @@
...
+
+    const localNow = new Date()
+    localNow.setMilliseconds(0)
+    const nowAtServerTimeZone = new Date(
+        localNow.toLocaleString('en-US', {
+            timeZone: systemInfo.serverTimeZoneId,
+        })
+    )
+    const timestamp = new Date(dateTime).getTime()
+    const timeOffset = localNow.getTime() - nowAtServerTimeZone.getTime()
```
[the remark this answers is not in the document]
HendrikThePendric: Memoization also comes at a cost and this is not an expensive computation at all.

src/shared/use-server-time-as-local/use-now-at-server-timezone.js
```diff
@@ -0,0 +1,9 @@
+import { getCurrentDate } from '../fixed-periods/index.js'
+import { useServerDateTimeAsLocal } from './use-server-date-time-as-local.js'
+
+export const useNowAtServerTimezone = () => {
+    const localNow = getCurrentDate()
+    localNow.setMilliseconds(0)
+
+    return useServerDateTimeAsLocal(localNow.toUTCString())
```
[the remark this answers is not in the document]
HendrikThePendric: That's a good point actually. When I initially built this hook, it was intended to accept date/time strings coming from API responses. But it makes sense to also accept actual date instances.

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
HendrikThePendric: I think you may be overly cautious here. The MDN docs for `Date.prototype.toLocaleString()` mention in the [parameters section](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#parameters) that a _BCP 47 language tag_ may be provided as a locale string.

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
```diff
@@ -0,0 +1,21 @@
...
+export const useServerDateTimeAsLocal = (dateTime) => {
+    const { systemInfo } = useConfig()
+
+    if (!dateTime) {
+        return undefined
+    }
+
+    const localNow = new Date()
+    localNow.setMilliseconds(0)
+    const nowAtServerTimeZone = new Date(
```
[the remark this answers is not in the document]
HendrikThePendric: This has some history to it. This was added to resolve a bug. But that was in another app, where we were always passing date/time strings comping from the server. Now these strings are only accurate to the second, and what was happening, especially on a local instance, was the following:
- We were doing a mutation and when the mutation was completed we would fetch the same data from the server.
- Sometimes this would happen in under a second
- And because the local now was accurate to the millisecond but the server date string was not (I guess they were also rounding things server side), we could get mismatches.
- This resulted in us sometimes displaying weird "time ago" messages, such as "This was approved a few seconds from now"

But, I will still keep this in mind when I rework this PR, since you suggested we should be able to pass in a date instance, and also it would just look more consistent as you suggested already.

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
HendrikThePendric: God, I just realised I got all of this completely the wrong way around. The calculation I was doing in the other app was the inverse of what we're trying to do here. Thanks for pointing that out.

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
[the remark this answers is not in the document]
HendrikThePendric: This whole spiel turned out to no longer be relevant, because I removed this hook.

HendrikThePendric:
Thanks @tomzemp for catching quite a few important things wrong with the current implementation. 🙇 I've now simplified it a lot and done things the right way around 🙃 . 

> There are also a couple of other places in the code where getCurrentDate should be swapped with useNowAtServerTimezone() [these affect the year boundary, e.g. if I log in 1 January 2023 in Addis, but server time is 31 December 2022, the year should be 2022] [I don't know if this is an issue that would in reality ever cause problems because I don't imagine people are doing much data entry on 1 January or 31 December]:
> 
> * https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/context-selection/period-selector-bar-item/period-selector-bar-item.js#L38
> * https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/context-selection/period-selector-bar-item/period-selector-bar-item.js#L64
> * https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/context-selection/period-selector-bar-item/use-periods.js#L24
> * https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/shared/fixed-periods/fixed-periods.js#L168
> * https://github.com/dhis2/data-entry-app/blob/TECH-1348/src/shared/fixed-periods/fixed-periods.js#L886

Reg. the above I tacked the first three points but decided to leave the last two. The reason is that the period generator is not a hook, so we'd have to pass the current date from the outside, or compute it in the module itself, which is also problematic because that also doesn't have access to server's timezone. We should still address these things at some point though, probably when we revisit the period implementation as a whole.

I've resolved most of your comments because I have updated the code accordingly, but left one open. Please resolve it if you agree.

src/context-selection/period-selector-bar-item/use-periods.js
```diff
@@ -15,13 +15,16 @@ export default function usePeriods({
     year,
     dateLimit,
 }) {
+    const nowAtServerTimezone = useNowAtServerTimezone()
+    const dateString = nowAtServerTimezone.toISOString().split('T')[0]
```
tomzemp, line 19: I'm not sure I get the use of the `.toISOString()` here?

For example, I'm in Auckland and it's 4:00 AM on 2024-01-01
My server is in Melbourne and it's 1:00 AM on 2024-01-01

So I send my nowAtServerTimeZone (which I've tricked to be 1:00 AM Auckland time on 2024-01-01) and now I get 11:00 AM on 2023-12-31 UTC from the conversion in toISOString(), which then causes my year limit to be 2023, even though both I and my server are in 2024?

I might also be misunderstanding. Either way, given this is only an issue on year boundaries, I don't think anyone will use the app on 31 December or 1 January, so maybe not worth it thinking it through anymore 🤪

tomzemp (approved):
Looks good to me now @HendrikThePendric 🚀! I left one other small question, but I think it doesn't matter either way, to be honest.

src/shared/use-server-time-as-local/use-server-date-time-as-local.js
tomzemp: Yeah...I think that it doesn't matter honestly 👍. I think MDN also just notes that the Date() instantiation can be implementation-specific, but I think all major browsers in reality parse correctly from localeDateStrings, so I don't think it's worth it to code against the possibility that browsers start making their parse implementation strange.

src/context-selection/period-selector-bar-item/use-periods.js
HendrikThePendric: Very good point. I misused the `.toISOString()` method here, because I had a misconception of what it does. My assumption was that it would return a string representation of a local date. I didn't consult [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString), but if I had, I would have read:
> The timezone is always zero UTC offset, as denoted by the suffix `Z`

Instead of keeping this flawed logic in the codebase, I created a helper in 223d3ec which will create a date string from a local date.

tomzemp: looks good, but I wonder if it might just be easier in the end to set the time portion to 0 and not convert back and forth to a string so as to avoid the exposure to parsing issues?
It seems like we're doing this back and forth string to date conversion (`const currentDate = new Date(dateString)`) in order to get a date with 00:00:00 time stamp, but we could also achieve that with:
```
nowAtServerTimezone.setHours(0)
nowAtServerTimezone.setMinutes(0)
nowAtServerTimezone.setSeconds(0)
nowAtServerTimezone.setMilliseconds(0)
```

HendrikThePendric: The conversion to a string is intentional, because we need to use it in the dependency array of the `useEffect` hook. `nowAtServerTimezone` is producing a new object reference on each render, so if we were to use it directly as a dependency here, the effect would run very often. The `dateString` on the other hand is a primitive, which will be recomputed on each render, but produce the same value unless the date actually changes. As such adding this variable to the `useEffect` dependency array will only trigger a re-render when the date actually changes, which is what we want.

PR #167, opened 2022-09-14
feat(bottom-bar): count number of form-errors

Birkbjo:
Fixes [TECH-1405](https://dhis2.atlassian.net/browse/TECH-1405)

Pretty simple implementation, with a entryForm zustand-store. Another approach would be to share the `form`-ref in a context, like @HendrikThePendric proposed earlier. Or we could just put it in the same zustand-store. Not sure what you think is best. 
Sharing form-ref could be beneficial if we eg. want to focus the first errored-field when the user clicks on the "Invalid"-icon or something?

### DIFF
--- a/src/bottom-bar/main-tool-bar.js
+++ b/src/bottom-bar/main-tool-bar.js
@@ -5,7 +5,11 @@ import cx from 'classnames'
 import React from 'react'
 import { validationResultsSidebarId } from '../data-workspace/constants.js'
 import useRightHandPanelContext from '../right-hand-panel/use-right-hand-panel-context.js'
-import { useDataValueSetQueryKey, useLockedContext } from '../shared/index.js'
+import {
+    useDataValueSetQueryKey,
+    useLockedContext,
+    useEntryFormStore,
+} from '../shared/index.js'
 import styles from './main-tool-bar.module.css'
 
 export default function MainToolBar() {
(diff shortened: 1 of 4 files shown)


HendrikThePendric (commented):
Thanks for realising we still needed to implement this. Having the button UI implemented correctly is already a great improvement. I think the functionality might actually be hard to implement completely correctly....
> Sharing form-ref could be beneficial if we eg. want to focus the first errored-field when the user clicks on the "Invalid"-icon or something?

I agree that this would probably be the way to go, but fear that possibly out UI inputs will need some work to support this as well. I think it will only be possible to programmatically focus an input once we wrap the input components in a `forwardRef` and then actually attach that ref parameter to the underlying `input` tag. But I could be wrong. It might be worth doing a small experiment to see what happens if we were to do sth like this:
```js
const form = useForm()
form.focus('fieldname')
```

My expectation is that we will not see anything change in the UI, for the following reasons:
- The field will become active
- But since our UI inputs don't support a ref, they will not get focussed
- And since they don't get focussed. the onFocus callback will not fire
- And without this callback the highlighted field will not be set
- So the cursor won't be in the input, and the input won't have the highlighted outline.

🤔  Perhaps what we could do as a not completely correct, but fairly accurate solution, is to set the highlighted field? That should be possible already, and will produce a dark-grey outline around the input. The input is not going to be focussed, but at least the user can see where to go.

Birkbjo:
> My expectation is that we will not see anything change in the UI, for the following reasons: 

You are right, I just tested this and nothing happens. `form.focus` is there to tell RFF that the field received focus, not to actually focus it. For that you would need the ref, as you suggest. 

Setting the highlighted field would be easy. But I think one benefit of focusing the field is that it would scroll that field into view. Maybe we could do both? And do eg `document.getElementByName(fieldName)[0]?.scrollIntoView()`?

Mohammer5 (approved):

PR #170, opened 2022-09-14
docs: add link to development demo

Birkbjo:
https://dhis2-data-entry.netlify.app


HendrikThePendric (approved):

Birkbjo:
> Though I assume netlify will also add the status to the deployed commit, so maybe a bit unnecessary to add a badge.

True, and we will probably remove this after/before merging to main/master.

PR #172, opened 2022-09-14
fix(aoc): persist selection on OrgUnit and period update

tomzemp:
https://dhis2.atlassian.net/browse/TECH-1379

Fix to ensure that aoc selection persists (when appropriate) when period or orgUnit is changed.

Add cypress tests for this


tomzemp:
Fixes the underlying issue that caused the bug and adds a cypress test for checking these are persisted. With the SL database, adding a cypress test that should not persist when updating period/orgUnit requires selecting a different OrgUnit (I think, but I have not 100% reviewed categoryOption metadata). Clicking the OrgUnit tree in cypress in CI proved buggy, so I have not implemented a test for this

cypress/integration/context-selection/categories/index.js
```diff
@@ -106,6 +106,23 @@ Given(
...
+Given(
+    'a data set, org unit, period and some but not all options have been selected and there are some categories connected to the data set',
+    () => {
+        cy.visit(
+            '/#/?dataSetId=lyLU2wR22tC&orgUnitId=ImspTQPwCqd&periodId=202212&attributeOptionComboSelection=LFsZ8v5v7rq-CW81uF03hvV'
+        )
+        cy.get('[data-test="data-set-selector"]').should('exist')
+    }
+)
+
```
Mohammer5, line 118: Duplicate of the same step definition right above this one

cypress/integration/context-selection/categories/index.js
```diff
@@ -159,3 +184,14 @@ Then('the selector should show that some items have been selected', () => {
 Then('a message is being displayed in the data workspace', () => {
     cy.contains('The current selection does not have a form').should('exist')
 })
+
+Then('a message is being displayed in the data workspace', () => {
+    cy.contains('The current selection does not have a form').should('exist')
+})
```
Mohammer5, line 190: Duplicate of the same step definition right above this one

Mohammer5 (changes requested):

Mohammer5:
I'm surprised that the PR checks pass. I would've expected that we'll have to generate a new set of fixtures for the network shim

cypress/integration/context-selection/categories/index.js
tomzemp: fixed. sorry

cypress/integration/context-selection/categories/index.js
tomzemp: fixed. sorry

tomzemp:
> I'm surprised that the PR checks pass. I would've expected that we'll have to generate a new set of fixtures for the network shim

@Mohammer5: I see information on the dataset is already in the fixtures we have saved and the categorycombo (etc.) is being used in other tests within that feature, so I assumed it was okay. It also passes for me in stub mode locally (as well as in live mode running against debug), but I could update the fixtures if you think that would be useful.

Mohammer5 (approved):
Alright, then! No need to do anything other than merging this PR I guess

PR #174, opened 2022-09-15
fix: manage unsaved comments and limits when moving cells

kabaros:
Before this change, if a user opens the details bar, adds or edits a comment (or min/max limits), then navigates on the next cell, the state of the comment  and min/max limits belonging to the previous are not reset and are kept while navigating through cells.

We discussed several solutions with @cooper-joe and chose a solution where we do not discard the comment / limits on focus change, but persist in local and/or session storage. We discussed having an indicator somewhere to show the unsaved items, but decided to not implement one at this stage (similar to GitHub issues comments). Full discussion [here](https://dhis2.slack.com/archives/C02NHB92S2H/p1663745881774069?thread_ts=1663142517.716799&cid=C02NHB92S2H).

## Changes in this PR
- implement a store with zustand to manage the unsaved state of comments and limits
- use [zustand persist](https://github.com/pmndrs/zustand#persist-middleware) middleware to ensure that the state is saved even when refreshing
- update the UI components to use the unsaved comment or limit for a cell if there is one
- tests for both comments and limits

### Screenshot / recording
https://user-images.githubusercontent.com/1014725/192112372-92a2c9e9-630a-47ad-b7a2-0ec88ccb9840.mov

### DIFF
--- a/src/data-workspace/data-details-sidebar/comment.js
+++ b/src/data-workspace/data-details-sidebar/comment.js
@@ -1,6 +1,5 @@
 import i18n from '@dhis2/d2-i18n'
 import {
-    CircularLoader,
     Button,
     ButtonStrip,
     SingleSelect,
@@ -25,7 +24,7 @@ import LoadingError from './loading-error.js'
 
 const title = i18n.t('Comment')
 const errorMessage = i18n.t(
-    'There was a problem loading the comment for this data item'
+    'There was a problem updating the comment for this data item'
 )
 
 const CommentOptionSelector = ({ commentOptionSetId, inputOnChange }) => {
@@ -62,13 +61,13 @@ CommentOptionSelector.propTypes = {
 }
 
 const CommentEditField = ({
+    fieldName,
     comment,
     commentOptionSetId,
     onBlur,
     unsavedComment,
-    commentId,
 }) => {
-    const { input, meta } = useField(`comment_${commentId}`, {
+    const { input, meta } = useField(fieldName, {
         subscription: { value: true },
         initialValue: unsavedComment || comment || '',
     })
@@ -93,8 +92,8 @@ const CommentEditField = ({
 
 CommentEditField.propTypes = {
     comment: PropTypes.string,
-    commentId: PropTypes.string,
     commentOptionSetId: PropTypes.string,
+    fieldName: PropTypes.string,
     unsavedComment: PropTypes.string,
     onBlur: PropTypes.func,
 }
@@ -107,6 +106,7 @@ function CommentEditForm({
     closeEditor,
     unsavedComment,
     commentId,
+    isError,
 }) {
     const contextSelectionId = useContextSelectionId()
     const cellId = getCellId({ contextSelectionId, item })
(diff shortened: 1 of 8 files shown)


src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -153,6 +195,15 @@ export default function Comment({ item }) {
         deId: item.dataElement,
         cocId: item.categoryOptionCombo,
     })
+    const contextSelectionId = useContextSelectionId()
+    const commentId = getCellId({ contextSelectionId, item })
+    const unsavedComment = useUnsavedDataStore((state) =>
```
Birkbjo, line 200: `CommentEditForm` is already importing the store (for the setter), so I would argue this can be moved down to that component. 
Or move the setter to this component.

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -83,17 +93,44 @@ const CommentEditField = ({ comment, commentOptionSetId }) => {
...
+    const setUnsavedComment = useUnsavedDataStore(
+        (state) => state.setUnsavedComment
+    )
+
     const onSubmit = (values) => {
         // Don't send `undefined` (or 'undefined' will be stored as the comment)
-        const comment = values.comment || ''
+        const comment = values[`comment_${commentId}`] || ''
         syncComment({ comment })
+        setUnsavedComment(cellId, null)
```
Birkbjo, line 121: Should we have a `deleteUnsavedComment` instead? That way we are not "persisting" the ID at all. I'm pretty sure this would still linger in the store and not be garbage collected.

```
deleteUnsavedComment: (cellId) =>
    set((state) => {
        const unsavedComments = { ...state.unsavedComments }
        delete unsavedComments[cellId]
        return { unsavedComments }
    }),```

File: src/data-workspace/data-details-sidebar/comment.js [hunk 3]
```diff
@@ -83,17 +93,44 @@ const CommentEditField = ({ comment, commentOptionSetId }) => {
...
+    const cellId = getCellId({ contextSelectionId, item })
+    const setUnsavedComment = useUnsavedDataStore(
+        (state) => state.setUnsavedComment
+    )
+
     const onSubmit = (values) => {
         // Don't send `undefined` (or 'undefined' will be stored as the comment)
-        const comment = values.comment || ''
+        const comment = values[`comment_${commentId}`] || ''
         syncComment({ comment })
```
Birkbjo, line 120: We're not really guaranteed this succeeds right?
Due to the shared-data-value-mutation there needs some refactor to pass an `onSuccess`. But we could use the `mutateAsync`-instead. Se comment below

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -172,14 +223,16 @@ export default function Comment({ item }) {
...
     }
 
-    if (editing) {
+    if (editing || unsavedComment) {
         return (
             <CommentEditForm
                 item={item}
                 open={open}
                 setOpen={setOpen}
                 syncComment={setDataValueComment.mutate}
```
Birkbjo, line 232: Could use `syncComment={setDataValueComment.mutateAsync}`. Then `syncComment.then(() => deleteUnsavedComment` above.

src/shared/stores/unsaved-data-store.js
```diff
@@ -0,0 +1,36 @@
+import create from 'zustand'
+import { persist } from 'zustand/middleware'
+
+const inititalState = {
+    unsavedComments: {},
+    unsavedLimits: {},
```
Birkbjo, line 6: Minor and just my opinion, but I would argue that `unsaved`  is implied by the name of the store. So `comments` and `limits` would suffice.

Birkbjo (changes requested):
Most comments are minor, but would like to see a `delete`-method in the store instead of setting `unsaved` to `null`.

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -58,10 +61,16 @@ CommentOptionSelector.propTypes = {
...
-const CommentEditField = ({ comment, commentOptionSetId }) => {
-    const { input, meta } = useField('comment', {
+const CommentEditField = ({
+    comment,
+    commentOptionSetId,
+    onBlur,
+    unsavedComment,
+    commentId,
+}) => {
+    const { input, meta } = useField(`comment_${commentId}`, {
```
Birkbjo, line 71: Think it could be useful to have a helper-function to "generate" these, as they are hard-coded in two different components.

src/data-workspace/data-details-sidebar/comment.js
Birkbjo: I, see thanks for that!

While testing I noticed that the entire form is replaced by a spinner when saving. I know that this was not introduced in this PR, but it's an unfortunate cumulative layout shift. I think we should move this "saving"-logic down to the form, and show a spinner on the button instead.

Birkbjo (approved):
Good job! Thanks a lot for resolving everything so quick :)

PR #178, opened 2022-09-16
fix(validation button): disabled validation run when offline [TECH-1377]

tomzemp:
https://dhis2.atlassian.net/browse/TECH-1377:

Disables bottom bar 'Run validation' button when offline. 

I have added the tooltip (which is non-functional because ui library does not render tooltips on disabled buttons) (this is consistent with some of implementations elsewhere in the app where we have preemptively implemented the tooltip in anticipation of ui library update).

https://dhis2.atlassian.net/browse/TECH-1387:

Currently, hitting the "Run validation" button a second time hides the validation sidebar. Updating the button to rerun the validation seemed to me to require a non-trivial refactor which would not provide a lot of value; I think the question of how the button should behave should probably be run by Joe first given the work effort to change it. For the time being, I've just updated the button say "Hide validation" when the validation sidebar is open, so that the user knows that clicking the button will in fact hide the validation sidebar and not (re)run validation.

hide_validation


tomzemp:
Thanks @kabaros : I didn't honestly spend a great deal of time evaluating the point about TECH-1387 beyond deciding it probably needed more evaluation/refactoring than I wanted to do prior to the code freeze (so maybe my comment was too dramatic and great if you've found out there was an easy-to-implement solution 😀!) I believe I thought it would be hard to cancel the current request (which was in Joe's specs)?

I can wait on https://github.com/dhis2/aggregate-data-entry-app/pull/200 and the update this PR to remove the "Run validation/Hide validation" toggle once that is merged

Mohammer5:
> I can wait on https://github.com/dhis2/aggregate-data-entry-app/pull/200 and the update this PR to remove the "Run validation/Hide validation" toggle once that is merged

Could you either turn this PR into draft mode or prepend the title with "BLOCKED" or something similar?

tomzemp:
Marked as draft pending resolution of PR200 / TECH-1387

PR #181, opened 2022-09-18
fix: highlighted cell synced background

Birkbjo:
Fixes some styling issues with highlighted field. Previously, the `!important` keyword on highlighted-field prevented `invalid` and `synced` styles from being applied to the highlighted field when this was unfocused.
I think we want the `white`-background when editing the field - and this is now done by using the `.active`-selector, which is in my opinion more correct than using the `highlighted-field`-selector.

Highlighted-synced unfocused:

image

Before: 
image

Highlighted-synced focused:
(same as before)
image

Highlighted-error unfocused:

image

Before:
image

### DIFF
--- a/src/data-workspace/data-entry-cell/data-entry-cell.module.css
+++ b/src/data-workspace/data-entry-cell/data-entry-cell.module.css
@@ -24,6 +24,10 @@
     outline: 1px solid #a0adba;
 }
 
+.active {
+    background-color: #fff !important;
+}
+
 .highlighted {
     outline: 3px solid var(--colors-grey800) !important;
     border: none !important;
@@ -31,6 +35,7 @@
     /* Fix to prevent bottom outline to be clipped by next cell*/
     z-index: 1;
 }
+
 
 .highlighted.active {
     outline-color: var(--theme-focus) !important;
@@ -40,10 +45,11 @@
     background: var(--colors-grey300);
 }
 
-.invalid:not(.highlighted) {
+.invalid:not(.active) {
     background: var(--colors-red200);
-    outline: 1px solid var(--colors-red600) !important;
+    outline: 1px solid var(--colors-red600);
 }
+
 .invalid:hover {
     background: #ffb3bc;
 }
--- a/src/data-workspace/data-entry-cell/data-entry-cell.module.css
+++ b/src/data-workspace/data-entry-cell/data-entry-cell.module.css
@@ -31,8 +31,7 @@
 .highlighted {
     outline: 3px solid var(--colors-grey800) !important;
     border: none !important;
-    background: #fff;
-    /* Fix to prevent bottom outline to be clipped by next cell*/
+    /* Fix to prevent bottom outline to be clipped by next cell */
     z-index: 1;
 }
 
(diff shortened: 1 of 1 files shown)


src/data-workspace/data-entry-cell/data-entry-cell.module.css
```diff
@@ -38,10 +45,11 @@
     background: var(--colors-grey300);
 }
 
-.invalid:not(.highlighted) {
+.invalid:not(.active) {
```
Birkbjo, line 48: Using the `highlighted`-selector prevented the highlighted-but-unfocused cell to not show the red-background.

HendrikThePendric (approved):
I think the changes you have made here and the rationale behind them are correct. So we can go ahead and merge this in. This visual bug was a regression from https://github.com/dhis2/data-entry-app/pull/142 where I introduced the highlighted cell state for cells without focus.

Whilst testing the branch via the PR link, I noticed a few odd things in this cell UI. I'm pretty sure they weren't introduced in this PR, but might be something for @cooper-joe to consider....

Below is a cell into which I first entered a valid value, which was then synced. I then changed the value to a string which is invalid. Now the cell is "synced and errored". Technically correct, but users will get confused....
Screenshot 2022-09-19 at 09 25 56
Perhaps here we should hide the sync indicator? (Not sure...)

And below is that same "synced and errored" cell again, but now I have focussed the input. In this case we have a cell with an invalid value, and we have a tooltip above which indicates _why_ the value is invalid. However, since there is no red color, and we're even seeing some green from the sync indicator, I don't think the user can see very easily that this cell is invalid right now...
Screenshot 2022-09-19 at 09 29 10
Perhaps it would make sense to show a red outline in this case?

Birkbjo:
@HendrikThePendric 

I've a PR up to fix some of the conflicting cell-states you are seeing here https://github.com/dhis2/data-entry-app/pull/182 . This simply resets the "synced"-status whenever the value changes. I think that makes sense, since now the value does not match the last synced value.

tomzemp (approved):
Looks good!

It does make me wonder if we should also reset the highlightedFieldID if they click outside the cell? (separate issue)

HendrikThePendric:
> It does make me wonder if we should also reset the highlightedFieldID if they click outside the cell? (separate issue)

No because then we don't see the highlighted cell anymore when interacting with the sidebar.

cooper-joe:
> Below is a cell into which I first entered a valid value, which was then synced. I then changed the value to a string which is invalid. Now the cell is "synced and errored". Technically correct, but users will get confused....
>
> Perhaps here we should hide the sync indicator? (Not sure...)

Yeah, I think hiding the sync indicator makes sense here. The invalid value is not synced, so a sync indicator is false signal.

> However, since there is no red color, and we're even seeing some green from the sync indicator, I don't think the user can see very easily that this cell is invalid right now...
>
> Perhaps it would make sense to show a red outline in this case?

The cell did have a red border/outline before the user focused it. I think showing the user that they have focused the cell is important, as they will hopefully already be aware of the error state from the tooltip and previous red background/outline.  In isolation, it looks confusing, but I'm thinking it'll be clear in the context of using the app for real data entry.

PR #183, opened 2022-09-19
fix(comment-indicator): use correct parameter

Birkbjo:
Fixes [Tech-1424](https://dhis2.atlassian.net/browse/TECH-1424)

Might've been introduced in a bad merge-conflict, but we were using the wrong parameter-name.


tomzemp (approved):
lgtm 👍

PR #184, opened 2022-09-19
fix(update-min-max): only show error message after submit failed

Birkbjo:
~~Adds a `touched`-check for validation so that the error-message does not show up before they have the user has a chance to prevent the error.~~

Errors are now hidden until the user has clicked "Save limits", and only shown if this fails. This is to prevent validation message showing up before user has a chance to prevent the error. After the first time submission fails, the validation will work like before. This is to make the error-message response once an error has occurred. 

I've enabled the `Save Limits`-button to prevent confusion, where the `Save limits`-button was disabled without an error message. Error message will show and stop submitting (updating minmax) if no valid limits.

I was also a bit confused about the bi-directional `'A min is required when providing a max` for both min and max-values. This basically just means that both values are required, right? Anyway I've updated that, since the disabled field is removed, it was possible to post empty-values - which results in 409 from the server.

### DIFF
--- a/src/data-workspace/data-details-sidebar/update-limits.js
+++ b/src/data-workspace/data-details-sidebar/update-limits.js
@@ -19,7 +19,7 @@ function UpdateLimits({
     onCancel,
 }) {
     const form = useForm()
-    const { submitting, errors, touched } = form.getState()
+    const { submitting, submitFailed, errors } = form.getState()
 
     const minField = useField('min', {
         initialValue: limits.min,
@@ -46,8 +46,7 @@ function UpdateLimits({
                     <LimitsInput
                         {...minField.input}
                         label={limitInputLabelsByName.min}
-                        error={minField.meta.touched && !!minField.meta.error}
-                        meta={minField.meta}
+                        error={submitFailed && !!minField.meta.error}
                     />
 
                     <div className={styles.spaceBetween}></div>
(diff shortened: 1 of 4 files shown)


src/data-workspace/data-details-sidebar/update-limits.js
```diff
@@ -46,31 +46,25 @@ function UpdateLimits({
                     <LimitsInput
                         {...minField.input}
                         label={limitInputLabelsByName.min}
-                        error={!!minField.meta.error}
+                        error={minField.meta.touched && !!minField.meta.error}
+                        meta={minField.meta}
```
tomzemp, line 50: do you want to pass this? It seems like it's not beeing used in LimitsInput component?

tomzemp (commented):
I'm okay with the updates (and am also okay with the previous implementation), but it still has the "issue" where you can get error messages before you are finished typing, e.g. enter min of 7 and then go to max and start typing 1:
min_max_limits

Do we want to just make it display only once save is clicked like Joe said he was okay with?

src/data-workspace/data-details-sidebar/update-limits.js
Birkbjo: Nop, this is remnants from another way to do this... Thanks!

Birkbjo:
@tomzemp thanks for your input! I've now simplified this, and just use `submitFailed` to prevent error-message from showing until the first failed submittion.

PR #185, opened 2022-09-19
ci: fix package.json name to ensure correct d2-ci deployment

amcgee:
The GitHub Actions workflow uses the name in `package.json` to determine the repository at https://github.com/d2-ci to push to, so this needs to match the package name.

The name/slug can be specified correctly as `data-entry` in `d2.config.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "data-entry",
+    "name": "data-entry-app",
     "version": "1.0.0",
     "description": "",
     "license": "BSD-3-Clause",


Birkbjo:
So this mean we won't push to https://github.com/d2-ci/data-entry/tree/development anymore?

amcgee:
> So this mean we won't push to https://github.com/d2-ci/data-entry/tree/development anymore?

Correct, it will push to https://github.com/d2-ci/data-entry-app

Birkbjo (approved):

PR #187, opened 2022-09-20
feat: reduce org-unit-tree network traffic and support prefetching

HendrikThePendric:
This PR dramatically reduces the amount of network traffic the org-unit-tree produces in general, and also adds support for reading from a list of pre-loaded items before issuing a network request (i.e. the network request will only be triggered when the units are not prefetched).

To achieve this I have copied the org-unit-tree component from @dhis2/ui and tweaked the implementation. This is obviously not ideal, but I think in this case it is justified, because the improvements are very significant. I'm not quite sure if these changes will ever make it back to @dhis2/ui because a more holistic refactor might be more appropriate for this component.

The changes I made to the component to add prefetching support and reduce network requests are all in these two commits https://github.com/dhis2/data-entry-app/pull/187/commits/7d41e5f5f910d06be134953b61a2c180313e7b0c and https://github.com/dhis2/data-entry-app/pull/187/commits/07d30d5128b69421f6c54285b0399dcac0d0d860. I  suggest that when you review this PR, you review this particular commit for changes in the org-unit-tree component, but you don't review the `src/context-selection/org-unit-selector-bar-item/organisation-unit-tree` folder as a whole.

One annoying consequence of moving this component, which uses styled-jsx, was that it broke our test suite. So I have to find a way around that (WIP). I'll probably also have to update some of the tests in the component as well. For these reasons, the PR is still in draft.

----

Below are some screenshots to illustrate the improved network traffic. When reviewing them, please take the following into account:
* I applied a filter on `organisationUnits` to the network-tab, so the visible requests are all related to the org-unit-tree. As such I don’t thing that the request count in the bottom left is completely accurate. It does give a good indication though.
* I first let the app load, then cleared the network tab, and then opened the org-unit-tree. As such the screenshot that demonstrates things *with* prefetching has a bit of an unfair advantage, since the requests that are needed to get the pre-fetched org-units in place are not visible. I believe that in total that’s 4 requests you’re not seeing (1 for the actual units and 3 to establish which levels we need to fetch). Having said that, for the prefetched version there are obviously many cases we can imagine where a user would interact with the org-unit-tree without ever producing a network request.

#### Original implementation (as on development)
opening_4_levels_on_development

#### New implementation *without* offline-levels/preloading
opening_4_levels_new_Implementation_no_preload

#### New implementation *with* offline-levels/preloading
opening_4_levels_new_Implementation_with_preload


HendrikThePendric:
I've updated quite a few tests, but one test suite was extremely flaky. This flakiness seemed to be completely unrelated to changes I made. Locally they seem to pass once and then fail. In any case, since this sin't super relevant for now, I simply skipped them in https://github.com/dhis2/data-entry-app/pull/187/commits/f2a84283e4b9276588452357d98d665e5741bc0c

src/context-selection/org-unit-selector-bar-item/use-offline-levels.js
```diff
@@ -0,0 +1,102 @@
...
+                        id,
+                        level,
+                        offlineLevels,
+                    })),
+            },
+            {
+                queryKey: ['configuration/offlineOrganisationUnitLevel'],
+                select: ({ level }) => level,
+            },
+        ],
```
Mohammer5, line 77: This "violates" the app's standard how to construct query keys.
See [here](https://github.com/dhis2/aggregate-data-entry-app/blob/development/src/data-workspace/query-key-factory.js) or [here](https://github.com/dhis2/aggregate-data-entry-app/blob/development/src/shared/use-data-value-set/query-key-factory.js) for examples.

Mohammer5 (approved):

PR #189, opened 2022-09-21
fix: restrict access to min/max edits based on authorities [TECH-1426]

tomzemp:
https://dhis2.atlassian.net/browse/TECH-1426

This restricts min/max edits to users with appropriate authorities `F_MINMAX_DATAELEMENT_ADD` `F_MINMAX_DATAELEMENT_DELETE` or `ALL`

With this change, buttons are hidden from user if they do not have relevant authorities.

_For example without any edit authorities, user will see:_
image
image

_With only add authority (and existing values)_
image

_With only delete authority (and existing values)_
image


src/app/load-app.js
```diff
@@ -1,16 +1,37 @@
...
 import i18n from '@dhis2/d2-i18n'
 import { CircularLoader, CenteredContent, NoticeBox } from '@dhis2/ui'
 import { node } from 'prop-types'
 import React from 'react'
 import { useCustomFormsPrefetch } from '../custom-forms/index.js'
-import { useMetadata } from '../shared/index.js'
+import { useMetadata, useUserInfoStore } from '../shared/index.js'
 import css from './load-app.module.css'
 
+const query = {
```
tomzemp, line 10: This query is the same as the one in the headerbar (to reuse cache): https://github.com/dhis2/ui/blob/222e21f93a8b0482f94ff6e499ce8b7cdd380824/components/header-bar/src/header-bar.js#L21-L25

src/app/load-app.js
```diff
@@ -31,7 +52,9 @@ const LoadApp = ({ children }) => {
         )
     }
 
-    return children
+    if (data && (userData || userError)) {
```
tomzemp, line 55: I set it up this way as it seems we don't want to block the user from using the app if something fails with api/me request, but at the same time it's better to wait for the response to complete...could also refactor this to await query with useDataEngine?

src/data-workspace/data-details-sidebar/update-limits.js
```diff
@@ -47,6 +49,7 @@ function UpdateLimits({
                         {...minField.input}
                         label={limitInputLabelsByName.min}
                         error={submitFailed && !!minField.meta.error}
+                        disabled={!canAdd}
```
tomzemp, line 52: technically you can't get to the edit state if you don't have add authority, but I kept this check in case we wanted to change the logic/let user with delete authority open the "edit" window to perform the delete alone

src/data-workspace/data-details-sidebar/limits-display.js
```diff
@@ -76,13 +78,20 @@ export default function LimitsDisplay({
...
                 )}
             </div>
-
             <ButtonStrip>
-                <EditButton onClick={onEditClick} />
-                <LimitsDeleteButton
-                    dataElementId={dataElementId}
-                    categoryOptionComboId={categoryOptionComboId}
-                />
+                {canAdd && <EditButton onClick={onEditClick} />}
```
tomzemp, line 82: the UpdateLimits component opened by going to edit mode has a delete button, but it seemed strange to let users with delete authorities open the edit component just in order to have access to the delete button there (they already have access to delete button in this component). Anyway, I've hidden EditButton if people don't have add authority.

src/app/load-app.js
Birkbjo: That's great! I think a comment in the code would be beneficial?

src/app/load-app.js
```diff
@@ -1,16 +1,37 @@
...
+    const {
+        loading: userLoading,
+        error: userError,
+        data: userData,
+    } = useDataQuery(query, {
+        onComplete: (userData) => {
+            updateStore(userData.user?.authorities)
+        },
+    })
+
```
Birkbjo, line 31: I think we should try to use `tanstack/query` here, like everywhere else in the app.
It uses `app-runtime` `engine` under the hood for the actual request, and I think we should manage to make them the same? Or does that apply headers or something we cannot reuse the cached-response?

Birkbjo (changes requested):

src/app/load-app.js
tomzemp: Sounds good. I thought maybe we wanted to only use tanstack query for the requests we wanted to more actively handle caching strategy for, but if we're okay with using it for this, that sounds good, and it will eliminate the need for the zustand store.

Commit here to refactor with tanstack query: https://github.com/dhis2/data-entry-app/commit/1dc468048105ef3136aa04b33278f83c4c8fe8b3

Birkbjo (approved):
Nice, LGTM!

PR #190, opened 2022-09-21
fix(data-value-set): refetch only when changing context-selection

Birkbjo:
Sorry for these back-and-forth fixes for refetching. I thought I had fixed this by not "subscribing" with `useIsMutating`-hook. But apparently if the subscribed field re-renders in the meantime between fetches - it will refetch all dataValues again. This could happen if you changed a field, then highlighted another field quickly. 

I've set the `staleTime: infinity` to have more control over refetches. Now we manually refetch whenever the form selection changes. 
Also simplified `refetchOnMount`, because this is not needed anymore due to our manual `invalidateQueries`.

### DIFF
--- a/src/data-workspace/data-workspace.js
+++ b/src/data-workspace/data-workspace.js
@@ -46,9 +46,11 @@ export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
     // used to reset form-state when context-selection is changed
     const formKey = useContextSelectionId()
 
+    // to keep one stable dependency for effect below
+    const validFormKey = isValidSelection ? formKey : false
     // force refetch when context-selection changes
     useEffect(() => {
-        if (isValidSelection) {
+        if (validFormKey) {
             // note this will only refetch "active"/mounted queries,
             // so it's safe to not provide a queryKey
             queryClient.invalidateQueries(null, {
@@ -57,8 +59,7 @@ export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
                 cancelRefetch: false,
             })
         }
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [formKey])
+    }, [validFormKey, queryClient])
 
     if (selectionHasNoFormMessage) {
         const title = i18n.t('The current selection does not have a form')
--- a/src/data-workspace/data-workspace.js
+++ b/src/data-workspace/data-workspace.js
@@ -20,7 +20,7 @@ import {
 import styles from './data-workspace.module.css'
 import { EntryForm } from './entry-form.js'
 import { FinalFormWrapper } from './final-form-wrapper.js'
-
+import { dataValueSets as dataValueSetQueryKey } from './query-key-factory.js'
 export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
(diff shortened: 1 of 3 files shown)


src/data-workspace/data-workspace.js
```diff
@@ -46,6 +47,20 @@ export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
...
+            // so it's safe to not provide a queryKey
+            queryClient.invalidateQueries(null, {
+                // if new selection is not in cache, a fetch will already be in-flight
+                // so we do not need to send another one.
+                cancelRefetch: false,
+            })
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [formKey])
+
```
HendrikThePendric, line 63: I don't like this rule to be disabled, since I have actually seen bugs with stale values in the `useEffect` callback due to this. I do see the problem though. As a solution, you could do the following:

Add this below line 48:
```js
const validFormKey = isValidSelection ? formKey : ''
```

On line 52, change change the check on `if (isValidSelection)` to a check for `if (validFormKey)`

On line 62, add the `validFormKey` to the dependency array. And also add `queryClient`, this should be a stable reference, so that shouldn't matter.

HendrikThePendric (approved):
Added one comment, I think it would be nice to address it, but approving for time's sake.

tomzemp (commented):
I'm theoretically okay with this...

but from the documentation, it sounds like calling invalidateQueries without specifying the queryKey will reset staleTime on all queries...which seems like it then makes our metadata query stale, and then causes it to refetch? I am at least seeing that now refetch on the netlify deploy whenever I change context selection

HendrikThePendric (changes requested):
> but from the documentation, it sounds like calling invalidateQueries without specifying the queryKey will reset staleTime on all queries...which seems like it then makes our metadata query stale, and then causes it to refetch? I am at least seeing that now refetch on the netlify deploy whenever I change context selection

Very well spotted by @tomzemp 🙏 I had only looked at the code.

Now I have compared the behaviour of the development branch to that on this branch, and when make the same selection in both and then change the period, this branch is producing way more requests (16) than development (2).

So I don't think this should be merged in in its current state.

Birkbjo:
> > but from the documentation, it sounds like calling invalidateQueries without specifying the queryKey will reset staleTime on all queries...which seems like it then makes our metadata query stale, and then causes it to refetch? I am at least seeing that now refetch on the netlify deploy whenever I change context selection
> 
> Very well spotted by @tomzemp 🙏 I had only looked at the code.
> 
> Now I have compared the behaviour of the development branch to that on this branch, and when make the same selection in both and then change the period, this branch is producing way more requests (16) than development (2).
> 
> So I don't think this should be merged in in its current state.

Thanks, I had filtered my network-tab and tunnel-visioned on `dataValueSet`-queries...
I've updated this now to pass the query-key.

I also had to include `{ type: 'active' }` , even though the docs say this is the default. If I didn't do that, it would send 2 requests when changing context, for some reason...

HendrikThePendric (approved):
Perfect @Birkbjo thanks for that.

tomzemp (approved):
looks good to me 🚢

PR #192, opened 2022-09-22
fix: open file links in new tab

kabaros:


src/data-workspace/inputs/file-inputs.js
```diff
@@ -108,7 +108,12 @@ export const FileResourceInput = ({
             {file ? (
                 <>
                     <IconAttachment16 color={colors.grey700} />
-                    <a href={fileLink} className={styles.fileLink}>
+                    <a
+                        target="_blank"
+                        rel="noreferrer"
```
tomzemp, line 113: I think we should include `noopener` as well

tomzemp (approved):

PR #197, opened 2022-09-26
fix: allow for unit paths starting at the real root instead of user root

HendrikThePendric:
This fixes [TECH-1445](https://dhis2.atlassian.net/browse/TECH-1445), which includes a full description of the problem.

To address the problem we now shorten all prefetched unit `path`s so that they start from the user-root, not the real root.


HendrikThePendric:
I've returned this PR into draft status. The current implementation doesn't work reliably. I'm investigating why at the moment.

Mohammer5 (changes requested):
This is quite a small change, I think we could add a test for this one to ensure it's working as expected without too much hassle

Mohammer5 (approved):

PR #198, opened 2022-09-26
fix(data-value-set): disable when mutating

Birkbjo:
Revisiting this, again... Hopefully last time now!

In [PR-190](https://github.com/dhis2/aggregate-data-entry-app/pull/190), we changed to `staleTime: Infinity` and manually `invalidating` the query to have more control when to refetch `dataValues`. However, we still need to disable the query when it is mutating, since this would refetch and overwrite the cache if you reloaded the app, with pending mutations. This should now be safe and not refetch on every request (like we had a problem with earlier), due to `staleTime: Infinity`.

Also cleaned up usage of `dataValueSetQueryKey` - since these were exported both from `data-workspace` and `shared`.


Mohammer5 (approved):
LGTM

PR #201, opened 2022-09-27
fix: use a stable date string instead of an instable date instance

HendrikThePendric:


tomzemp (approved):
Looks good to me! 

Since I don't fully trust myself on these dates, I reviewed the code (looked good) and then tested with Samoa time zone (currently 1 day behind Oslo), looking at a data set with daily period type. I also tested at 30 Sep/1 Oct date boundaries and again things seemed to work as expected.

Note that if a user manipulates the date on their computer, they will be able to see some additional periods, but editing the data will still be locked by the backend, so I don't think we need to do anything special client-side to police against this (i.e. user would have to "cheat" before they get any extended periods they shouldn't).

PR #203, opened 2022-09-30
feat: respect F_DATAVALUE_ADD user authority

Mohammer5:
Closes TECH-1439 ([link](https://dhis2.atlassian.net/browse/TECH-1439))

* [x] Shows a "locked" info notice box when authorities are missing
* [x] Disabled inputs when authorities are missing
* [x] Disabled (un-)marking for followup when authorities are missing
* [x] Disabled adding/editing comments when authorities are missing
* [x] Disabled settings/deleting min- & max-limits when authorities are missing
* [x] Ensures that `<LoadApp />` always returns something (no implicit `undefined` return)
* [x] Removes `console.log` statement from `use-current-date-string-at-server-timezone.js`

NOTE: @HendrikThePendric commented on the tooltip issue. We've agreed to tackle that separately as it's already tracked in [TECH-1341](https://dhis2.atlassian.net/browse/TECH-1341) and it's not a new issue of this PR.

### DIFF
--- a/src/app/load-app.js
+++ b/src/app/load-app.js
@@ -32,14 +32,16 @@ const LoadApp = ({ children }) => {
         )
     }
 
-    if (!metadata.data) {
+    if (userInfo.isError) {
         return (
             <NoticeBox
                 error
                 className={css.noticeBoxWrapper}
-                title={i18n.t('There was a problem loading metadata')}
+                title={i18n.t(
+                    "There was a problem loading the user's information"
+                )}
             >
-                {i18n.t('The metadata response is invalid')}
+                {userInfo.error}
             </NoticeBox>
         )
     }
--- a/src/app/load-app.js
+++ b/src/app/load-app.js
@@ -9,14 +9,10 @@ import css from './load-app.module.css'
 const LoadApp = ({ children }) => {
     useCustomFormsPrefetch()
 
-    const { isLoading, isError, data, error } = useMetadata()
-    const {
-        isLoading: userLoading,
-        isError: userIsError,
-        data: userData,
-    } = useUserInfo()
+    const metadata = useMetadata()
+    const userInfo = useUserInfo()
 
-    if (isLoading || userLoading) {
+    if (metadata.isLoading || userInfo.isLoading) {
         return (
             <CenteredContent className={css.center} position={'top'}>
                 <CircularLoader />
@@ -24,22 +20,31 @@ const LoadApp = ({ children }) => {
         )
     }
 
-    if (isError) {
-        console.error(data)
+    if (metadata.isError) {
         return (
             <NoticeBox
-                className={css.noticeBoxWrapper}
                 error
+                className={css.noticeBoxWrapper}
                 title={i18n.t('There was a problem loading metadata')}
             >
-                {error?.message}
+                {metadata.error}
             </NoticeBox>
         )
     }
 
-    if (data && (userData || userIsError)) {
-        return children
+    if (!metadata.data) {
+        return (
+            <NoticeBox
+                error
+                className={css.noticeBoxWrapper}
+                title={i18n.t('There was a problem loading metadata')}
+            >
+                {i18n.t('The metadata response is invalid')}
+            </NoticeBox>
+        )
     }
+
+    return children
 }
 
 LoadApp.propTypes = {
--- a/src/data-workspace/data-details-sidebar/basic-information-follow-up-button.js
+++ b/src/data-workspace/data-details-sidebar/basic-information-follow-up-button.js
@@ -4,10 +4,12 @@ import React from 'react'
 import {
     useHighlightedField,
     useSetDataValueMutation,
+    useCanUserEditFields,
 } from '../../shared/index.js'
 import ItemPropType from './item-prop-type.js'
 
 const FollowUpButton = ({ item }) => {
+    const canUserEditFields = useCanUserEditFields()
     const { value } = useHighlightedField()
     const isEmptyField = !value
 
@@ -31,12 +33,13 @@ const FollowUpButton = ({ item }) => {
         )
     }
 
+    const disabled = isEmptyField || !canUserEditFields
     const markForFollowUpButton = (
         <Button
             secondary
             icon={<IconFlag24 color={colors.grey600} />}
             onClick={onMarkForFollowUp}
-            disabled={isEmptyField}
+            disabled={disabled}
         >
             {i18n.t('Mark for follow-up')}
         </Button>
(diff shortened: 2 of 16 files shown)


src/app/load-app.js
```diff
@@ -9,37 +9,42 @@ import css from './load-app.module.css'
...
+    if (metadata.isError) {
         return (
             <NoticeBox
-                className={css.noticeBoxWrapper}
                 error
+                className={css.noticeBoxWrapper}
                 title={i18n.t('There was a problem loading metadata')}
             >
-                {error?.message}
+                {metadata.error}
```
HendrikThePendric, line 30: Isn't error and object? And wouldn't we want to render a string?

src/app/load-app.js
```diff
@@ -9,37 +9,42 @@ import css from './load-app.module.css'
...
+    if (!metadata.data) {
+        return (
+            <NoticeBox
+                error
+                className={css.noticeBoxWrapper}
+                title={i18n.t('There was a problem loading metadata')}
+            >
+                {i18n.t('The metadata response is invalid')}
+            </NoticeBox>
+        )
```
HendrikThePendric, line 44: Why is this here? Does this actually happen? I would have expected that if `metadata.isLoading` and `metadata.isError` are both `false`, that `metadata.data` would always be populated.

src/app/load-app.js
HendrikThePendric: Also, what about the `userInfo.isError` case? Don't we need to show an error notification for that too? I expect that other parts of the app are relying on the user info to be there.

src/data-workspace/data-details-sidebar/basic-information-follow-up-button.js
```diff
@@ -56,8 +59,21 @@ const FollowUpButton = ({ item }) => {
...
 
+    if (!canUserEditFields) {
+        return (
+            <Tooltip
+                content={i18n.t(
+                    'You do not have the authority to mark this value for follow -up'
+                )}
+            >
+                {markForFollowUpButton}
+            </Tooltip>
```
HendrikThePendric, line 70: I believe this won't work. The issue is that once a form element (incl. the `button` tag) is `disabled` it stops emitting events. Since the `button` covers the entire tooltip wrapper, the `mousenter` and `mouseleave` events will not fire and the tooltip won't show.

To deal with this type of situation the `Tooltip` from @dhis2/ui supports render-props and you can do something like the last example here: https://codesandbox.io/s/disabled-component-with-tooltip-ubtyk?file=/src/App.js

src/data-workspace/data-details-sidebar/comment.js
```diff
@@ -258,14 +272,17 @@ export default function Comment({ item }) {
...
-                {item.comment ? i18n.t('Edit comment') : i18n.t('Add comment')}
-            </Button>
+            {!canUserEditFields && (
+                <Tooltip
+                    content={i18n.t(
+                        'You do not have the authority to add or edit comments'
+                    )}
+                >
+                    {addEditButton}
+                </Tooltip>
```
HendrikThePendric, line 282: See my previous comment. Perhaps it makes sense to have a dedicated component for this. I.e. a `ButtonWithTooltip` or something.

src/data-workspace/data-details-sidebar/limits-delete-button.js
```diff
@@ -12,10 +16,25 @@ export default function LimitsDeleteButton({
...
+                content={i18n.t(
+                    'You do not have the authority to delete limits'
+                )}
+            >
+                <Button small secondary disabled>
+                    {label}
+                </Button>
+            </Tooltip>
+        )
+    }
```
HendrikThePendric, line 36: Again, see previous comment

src/data-workspace/data-details-sidebar/limits-display.js
```diff
@@ -12,8 +15,21 @@ import styles from './limits.module.css'
...
 
+    if (!canEditFields) {
+        return (
+            <Tooltip
+                content={i18n.t('You do not have the authority to add limits')}
+            >
+                <Button small primary disabled>
+                    {editButtonLabel}
+                </Button>
+            </Tooltip>
```
HendrikThePendric, line 29: Again. See previous comment.

src/data-workspace/data-details-sidebar/no-limits.js
```diff
@@ -2,15 +2,31 @@ import i18n from '@dhis2/d2-i18n'
...
+        return (
+            <Tooltip
+                content={i18n.t('You do not have the authority to add limits')}
+            >
+                <Button small disabled className={noLimitsStyles.addButton}>
+                    {buttonLabel}
+                </Button>
+            </Tooltip>
+        )
+    }
```
HendrikThePendric, line 28: Repeat

HendrikThePendric (changes requested):
the overall approach is looking good. But I think there is a structural problem with tooltips which wrap a disabled button, and think we might want to handle things slightly differently in `LoadApp`.

HendrikThePendric:
Hadn't realised this, but the issue with the tooltip not working on disabled elements is browser specific. On FireFox things are already working as expected, but not on Chrome.

src/data-workspace/data-details-sidebar/basic-information-follow-up-button.js
Mohammer5: See note in the PR's description

src/data-workspace/data-details-sidebar/limits-delete-button.js
Mohammer5: See note in the PR's description

src/data-workspace/data-details-sidebar/comment.js
Mohammer5: See note in the PR's description

src/data-workspace/data-details-sidebar/limits-display.js
Mohammer5: See note in the PR's description

src/data-workspace/data-details-sidebar/no-limits.js
Mohammer5: See note in the PR's description

src/app/load-app.js
Mohammer5: What I observed in the code prior to my changes is:

* There is no explicit condition for when this is the case, but there's a check for the existence of `data` (`metadata.data` after I changed the code)
* Children won't be rendered if there's no user data or an error.

I suppose in regards to the user info we could add a case similar to the one I added for missing metadata data for the userData as well, because right now there would be just a blank workspace if that happens

src/app/load-app.js
Mohammer5: See comment below

src/app/load-app.js
Mohammer5: @HendrikThePendric and I have agreed that it doesn't make much sense to check for the presence of `.data`, so now all loading and error states have been covered instead

HendrikThePendric (approved):
I agree we can tackle the tooltip issue in a separate PR, but we do need to make sure it gets implemented during the hard-freeze. To ensure that happens I've moved [TECH-1341](https://dhis2.atlassian.net/browse/TECH-1341) into the [TECH-1356](https://dhis2.atlassian.net/browse/TECH-1356) epic.

PR #206, opened 2022-10-03
fix: handle temporal mutation errors

Birkbjo:
- Handles pausing mutations when a network or "access" error occurs
- Handles resuming mutations when a request succeeds
    - We may want to ping the server regularly when offline?
    - This will probably be handled in app-runtime / header-bar in the future.
- Cleaned up mutation-configurations, using defaults as much as possible. 
- Handles session expiration - user get's alerted with an Alertbar when a request fails due to no session.
    - This is temporary, since Austin is working getting this into platform/runtime. 
- This does **not** implement error-handling for other errors (like 409s, 500-errors). Will implement that in a separate PR. 

Here's a few tips for testing this (using Google Chrome devtools):

* Try to block requests to /dataValues
    *  Right click dataValue request and "Block request url"
    * Try to change values
    * Disable the blocking
    * Change another value
    * Observe all request are sent
    * Repeat above, but reload app before enabling request
 * Test "session expiration"
     * Go to application-tab in chrome devtools
     * Cookies -> Delete JSESSIONID
     * Try to change a value
* Test "session expiration", simulating logging in in another tab
    * Go to application-tab in chrome devtools
    *  Cookies -> Right-click JSESSIONID -> Edit "Name"
    * Rename to sommething else
    * Try to edit value.
    * Rename cookie back to "JSESSIONID"
    * Send another value
    *


src/shared/data-value-mutations/use-api-error.js
```diff
@@ -1,27 +1,75 @@
...
+ */
+const statusCodeErrorAlertProps = {
+    // this is handled in `retry`
+    // 401: {
+    //     message: i18n.t(
+    //         'Session Expired. Please reload to log in again. You will not lose any data.'
+    //     ),
+    //     actions: [{ label: 'Reload now', onClick: () => location.reload() }],
+    // },
+}
```
Birkbjo, line 21: I refactored the structure of this before I realised the session-expiration handling had to be done in `retry` instead of `onError`. 

Should I delete this refactor? It does make the overrides more flexible for the future, but also introduces a bit more complexity.

src/app/query-client/use-query-client.js
```diff
@@ -1,34 +1,61 @@
...
+// error-types that should trigger offline-mode
+const triggerOfflineErrorTypes = new Set(['network', 'access'])
 const queryClient = new QueryClient({ logger })
 
 const useQueryClient = () => {
     const engine = useDataEngine()
     const queryFn = createQueryFn(engine)
+    const { onError } = useApiError()
+
+    const { show: showSessionExpiredAlert } = useSessionExpiredAlert()
```
Birkbjo, line 27: Can remove this once this is implementid in platform/runtime.

src/app/query-client/use-session-expired-alert.js
```diff
@@ -0,0 +1,21 @@
...
+        () =>
+            i18n.t(
+                'Session Expired. Please reload to log in again. You will not lose any data.'
+            ),
+        () => ({
+            critical: true,
+
+            message: i18n.t(
+                'Session Expired. Please reload to log in again. You will not lose any data.'
+            ),
```
Mohammer5, line 15: Why is the message passed to the options object again? Seems superfluous to me

src/shared/default-on-success.js
```diff
@@ -0,0 +1,22 @@
...
+
+export function defaultOnSuccess(callback) {
+    if (typeof callback !== 'function') {
+        return handleDefaultOnSuccess
+    }
+    return function composeSuccessHandlers(...args) {
+        handleDefaultOnSuccess.apply(this, args)
+        callback.apply(this, args)
+    }
+}
```
Mohammer5, line 11: We can address this in post-release work (if we decide that it needs to be addressed at all):

This is a weird pattern to me (passing the callback and having it being called after `handleDefaultOnSuccess`) as this is synchronous code. The current implementation hides the fact when what is called in what order, which leads to code that's harder to reason about. I'd prefer to simply call the function after calling `handleDefaultOnSuccess` as that makes core more understandable:

```js
{
  onSuccess: () => {
    // doesn't accept any parameters, so can call it like this
    handleDefaultOnSuccess()

    // if this needs parameters, they can be defined explicitly in the `onSuccess` callback.
    // Otherwise, if we need to call something before `handleDefaultOnSuccess`,
    // we'd have to either handle that inside `handleDefaultOnSuccess`, too.
    // Or we end up with an inconsistent pattern (on callback before `defaultOnSuccess` and one inside)
    callback() 
  },
}
```

I suppose the idea is that, by using `defaultOnSuccess` on success as a convention, it's guaranteed that the default behavior will be executed, but it's just a convention. It could simply be expected to use `handleDefaultOnSuccess` in the `onSuccess` callback.

src/shared/default-on-success.js
```diff
@@ -0,0 +1,22 @@
...
+
+export function handleDefaultOnSuccess() {
+    if (!onlineManager.isOnline()) {
+        // reset to default online-implementation
+        // this is done to move from 'offline' to 'online' whenever a request succeeds
+        onlineManager.setOnline(undefined)
+        // notify listeners
+        // this sends paused mutations
+        onlineManager.onOnline()
```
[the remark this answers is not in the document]
Mohammer5: Could you add your comment above as a code comment to the code?

src/shared/data-value-mutations/use-api-error.js
```diff
@@ -1,27 +1,75 @@
...
+    }
+    // find alertProps for errorCode, using detail-prop according to priority above
+    for (const [key, errorMap] of errorDetailsMessagePriority) {
+        const errorCode = error?.details[key]
+        if (errorCode) {
+            errorMessage = errorMap[errorCode]
+            break
+        }
+    }
+    return errorMessage
```
Mohammer5, line 45:
```suggestion
    for (const [key, errorMap] of errorDetailsMessagePriority) {
        const errorCode = error?.details[key]
        if (errorCode) {
            return errorMap[errorCode]
        }
    }
    return { message: error.message }
```

Mohammer5 (commented):
Just a code review and some comments, will test the functionality now

Mohammer5 (commented):
Overall this looks good to me and works fine! Let me know what you think about my code comments and once they're resolved, I can approve this PR

> Test "session expiration"

Not an issue with this PR, but something we might want to think about: When changing values, then reload the app, login again, after the pending values have been synced, then their field is not green

Birkbjo:
> Not an issue with this PR, but something we might want to think about: When changing values, then reload the app, login again, after the pending values have been synced, then their field is not green

I've noticed this as well, but I do think it's a pretty complex case to solve, with the way we are using "optimistic updates". I think it's somewhat minor. If you are still offline after reloading, the syncing indicators will still be correct. It's only when they are sent, and in that case the user didn't really do any changes to the form yet - so showing green fields might even be confusing. From a user's perspective, the form should be the same as visiting a form with values already in. Green fields should indicate that changes I did in the current session are synced and valid. We are guaranteed that all the mutations are sent, and the form is refetched from the server before showing the form - so it is initialised with the correct values.

src/shared/default-on-success.js
Mohammer5: As I said, this is something we can discuss post-release when looking at refactoring the app. I'll create an issue in the post-release epic with my "concerns", stating that I'd like to discuss it, but wouldn't mind keeping things as they are if that's the consensus

Mohammer5 (approved):
From my side only the code comment is missing, other than that this PR looks very good to me! :muscle:

src/data-workspace/data-workspace.js
```diff
@@ -79,11 +79,9 @@ export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
...
     }
 
-    // Currently this can cause the form to reload when going back online
-    if (
-        initialDataValuesFetch.isFetching &&
-        initialDataValuesFetch.data === undefined
-    ) {
+    // We want to block initialization of form if in-flight
+    // or else we might use stale data that won't be updated once request completes
+    if (initialDataValuesFetch.isFetching) {
```
Birkbjo, line 84: Note that this is fine for now, since we are not refetching form after mount. Might cause issues we do that down the line though.

src/shared/data-value-mutations/use-api-error.js
Birkbjo: Note that this part will probably be changed a bit in a new PR handling "permanent" submit errors.

src/shared/default-on-success.js
Birkbjo: I don't really understand the pushback on this though, from an API-perspective (`defaultOnSuccess()`), it's just a regular callback? It's perfectly valid to use callbacks in synchronous code. I understand it might be a bit "weird", since it looks like we are just calling the callback after the default-functionality, but there's a reason - we keep the `this`-context intact. The `this`-context can be valuable in `onSuccess`-handlers to get info about the `query/mutation-key` and other info.
There's also less code than having to manually call the `handleDefaultOnSuccess` in every place we have this.

PR #208, opened 2022-10-04
fix: handle invalid parameters [TECH-1382] [TECH-1384]

tomzemp:
This PR addresses two tickets regarding invalid parameters.

**General note:** I have updated the code to silently handle errors (e.g. if you enter an invalid data set, or a data set you do not have access to, I have not updated to provide an explicit warning but rather to clear out selection, so you can choose another data set). I assume that issues with invalid parameters in links would arise mostly if users share links to the data entry app but they have different access privileges, so I am not sure we would want to draw attention to this lack of access? That might be a question to Joe for future refinement?

Also note that I have updated a number of cypress tests as part of this ticket. The tests were previously working, but stopped working because we were using future periods / orgUnits not assigned to dataset, so they stopped working. I've just updated them to use current periods / valid orgUnits to make them run after these code updates.

### [TECH-1382](https://dhis2.atlassian.net/browse/TECH-1382) 
App errors out when invalid parameters are provided. 

_Current:_
Entering an invalid parameter (e.g. `dataSetId=fakeDataSe` thttps://play.dhis2.org/dev/dhis-web-aggregate-data-entry/index.html#/?dataSetId=fakeDataSet&orgUnitId=ImspTQPwCqd&periodId=202211) causes app to error out:
image

_After:_
fake_data_set_after
(Invalid data set id causes data set to be reset, clearing out the other provided parameters)

Note: code contains fixes to handle cases where values are undefined (to prevent erroring out) and code to reset invalid values.

### [TECH-1384](https://dhis2.atlassian.net/browse/TECH-1384) 
App allows you to open forms when directly entering orgUnitId not associated with dataset or when entering an invalid period for the dataset

_Current:_
image

_After_:
invalid_org_unit

OrgUnit is cleared because it is not associated with data set.


Mohammer5:
How difficult would it be to keep the parameters but instead show the user a message that he doesn't have access to "this form"?
I suppose that'd include putting some work into the context selector items as well?

tomzemp:
> How difficult would it be to keep the parameters but instead show the user a message that he doesn't have access to "this form"? I suppose that'd include putting some work into the context selector items as well?

@Mohammer5 : that's possible, but 
- requires a bit of refactoring to set the "no form" message within the context selector items
- would we want the message to be specific? For example, if the user tries  `dataSetId=restrictedDE` and then should we display a no form message that says something like `The dataset (restrictedDE) does not exist or is not shared with you`? (Capture has `Program doesn't exist`) (I think most likely you end up with invalid links if a user with different sharing privileges sends you a link to look at something...which means I guess we would want to say something about sharing?)
- I assume if we show a no form message, we'd also want to keep the invalid item in the url, like `dataSetId=BfMAe6Itzgt&orgUnitId=fdc6uOvgoji&periodId=invalidPeriod`, in which case we also need to modify the code to handle invalid selections (what I've implemented is easier because it just clears out invalid parameters, but if we keep them around, we need to also update logic some places, e.g. do not pass valid orgUnitId for Bombali if dataSet is not assigned to orgUnit, and also the logic for assessing the selection validitiy
- If we do this, is it sufficient to have a greedy approach to determining failure/displaying no form, like if user posts a link like `dataSetId=restrictedDE&orgUnitId=restrictedOrgUnit&periodId=invalidPeriod`, can we treat it as a failure at dataSetId, set a message for 'No form (restrictedDE invalid)` and then ditch the other parameters, so we end up with `dataSetId=restrictedDE` in the URL? Otherwise, I think we can run into issues keeping parameters that do not align?

I'm not sure if it makes sense to refactor as part of these tickets, or to open another ticket to discuss the ideal final handling with feedback?

Mohammer5:
I'd like to get some input from @cooper-joe. If he thinks that the currently implemented solution (in this PR) is good enough for the first iteration, then we can improve this later.

I think (regardless of whether we do it now or later), we should show a message to the user and keep the invalid parameters in the url. It's more clear why the link isn't working and consistent with what capture app does

cooper-joe:
I think this PR right now is OK for the first iteration. It would be even clearer if we could have an `AlertBar` type `warning` triggered by the failure and URL parameter clearance. Elsewhere in DHIS2 we display a generic message when there's multiple reasons why something could be failing and we can't (or don't want to, for security reasons) expose whether that's because it doesn't exist or because of access.

Example in this case could be:
`There was a problem loading some of the Data set, Org. unit, or Period selections. You might not have access, or the selections might be invalid.`
image

Regarding keeping invalid parameters in the URL, I think this warrants a bigger discussion. My first reaction is "keep them if possible", but @tomzemp you raise some valid points about the problems, so I'm not sure it's worth it. Would be good to have a discussion around this if/when we want to add it.

tomzemp:
Okay. Sounds good. I think adding the alert should not be too much additional work. I've put the PR back into draft and will update.

tomzemp:
@Mohammer5 @cooper-joe: I've added alerts. 

Since the logic for clearing out invalid parameters occurs on the level for each parameter, we can deliver a more specific message by metadata object and we can include the offending id in the message (so even though we clear it from the URL, the user can still see what value is the problem):
image
 
- **For dataSet, organisationUnit, and attributeOptionCombo**: I used the message from Joe (slightly modified to incorporate id / reflect reference to only one object): `There was a problem loading the [Data Set || Organisation Unit || Attribute Option Combo] selection ([id]). You might not have access, or the selection might be invalid.`
- **For period:** this text seemed a bit imprecise (it's not necessarily a problem with the period, but rather the period might be the wrong type or out-of-bounds for the data set), so I used this language: `The Period ({{id}}) is not open or is invalid.`

_Note:_ I cannot get a line break display in the AlertBar, so I'm not sure if that's possible or if I'm just doing something wrong (I just tried passing the message with a line break, like `First sentence.\nSecond sentence.` )

Mohammer5 (approved):

PR #210, opened 2022-10-07
fix: handle inconsistent formats from api for orgUnits [DHIS2-13888]

tomzemp:
See https://dhis2.atlassian.net/browse/DHIS2-13888

This fix:

- returns organisationUnits property  (what the app is currently doing) if `organisationUnits` exists on the response
- returns the response wrapped in an array [response] if `organisationUnits` is not on the response (which will happen when there is only one organisationUnit returned)

I think it would ultimately be better to get the api to be consistent in its responses as coding this logic in the app seems like it becomes hard to catch if the format changes again in the future, but I gather that would be difficult at this point of time?

Note: we have opened a [Ticket 13901](https://dhis2.atlassian.net/browse/DHIS2-13901) to address this on the backend (once fixed, we can revert this change).


Birkbjo:
> I think it would ultimately be better to get the api to be consistent in its responses as coding this logic in the app seems like it becomes hard to catch if the format changes again in the future, but I gather that would be difficult at this point of time? 

I think it's fine to fix in the app. I agree, ideally it would be better if the backend returned a consistent response, but you never know if apps depend on this response-format. In this case, that seems highly unlikely. But it's still a breaking change, and I don't think we should do that this late in the cycle.

HendrikThePendric (approved):
Yeah, I agree this is the fix we need to implement right now. Once the backend is fixed we can remove it. Please include a link to the backend JIRA issue in the comment before merging: https://dhis2.atlassian.net/browse/DHIS2-13901

PR #216, opened 2022-10-17
chore: add entry info [DHIS2-13940]

tomzemp:
This PR adds text/icon when app is in empty state (selections to load form are not complete)
See [DHIS2-13940](https://dhis2.atlassian.net/browse/DHIS2-13940). Note that text/link may still be modified.

image


Mohammer5 (approved):

cooper-joe (approved):
Approving the design / component. Think we can merge and update the text if/when settling on a revised text, but would like this to go in!

PR #219, opened 2022-10-18
fix: remove custom styles for data details button

cooper-joe:
This PR removes the custom style overrides used on the `View details`/`Hide details` button in the bottom bar. These styles were to spec, but weren't working well during testing. 
The button now uses a standard `ui` button, type `small secondary`.

Before:
image

After:
image


Birkbjo (approved):

src/bottom-bar/data-item-bar.js
```diff
@@ -18,6 +18,7 @@ export default function DataItemBar() {
             </span>
 
             <Button
+                secondary
                 small
                 className={styles.dataDetailsButton}
```
tomzemp, line 23: can we remove the className, now that there is no `dataDetailsButton` class? Otherwise, looks good to me

cooper-joe: Yes, definitely. Done. Nice spot 👍

Mohammer5:
@cooper-joe this PR has been approved by @Birkbjo and the only requested change by @tomzemp has been implemented. I think you can go ahead and merge this PR

PR #229, opened 2022-10-19
fix: make useDataValueParams and useApiAttributeParams stable

Birkbjo:
These were not stable, and thus the result of these hooks were unfit for use in other hooks, as that would cause them to run on every-render.

### DIFF
--- a/src/shared/use-api-attribute-params.js
+++ b/src/shared/use-api-attribute-params.js
@@ -1,6 +1,6 @@
+import { useMemo } from 'react'
 import { useMetadata, selectors } from './metadata/index.js'
 import { useContextSelection } from './use-context-selection/index.js'
-
 /**
  * Finds the attributeComboId for the current attribute-option selection.
  * Returns { attributeCombo: undefined, attributeOptions: undefined }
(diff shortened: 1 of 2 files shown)


src/shared/use-api-attribute-params.js
```diff
@@ -11,30 +11,34 @@ export const useApiAttributeParams = () => {
...
-
-    return {
-        attributeCombo: undefined,
-        attributeOptions: undefined,
-    }
+        return {
+            attributeCombo: undefined,
+            attributeOptions: undefined,
+        }
+    }, [data, dataSetId, attributeOptionComboSelection])
```
tomzemp, line 43: I think `attributeOptionComboSelection` itself won't be stable (see @Mohammer5's note here: https://github.com/dhis2/aggregate-data-entry-app/blob/554025ccca2063af53d20c77b65e27b8f9ada478/src/shared/use-context-selection/use-manage-inter-param-dependencies.js#L168)?

Birkbjo: From my testing this is not the case. 

```
    const { attributeOptionComboSelection } = useContextSelection()
    useEffect(() => {
        console.log('run useffect')
    }, [attributeOptionComboSelection])
```

Only runs when the option is changed.

tomzemp: Cool. Well that makes things easier 🙂

tomzemp (approved):
lgtm

KaiVandivier (approved):

PR #237, opened 2022-10-28
fix: audits styling [TECH-1469] [TECH-1470]

tomzemp:
[TECH-1469](https://dhis2.atlassian.net/browse/TECH-1469) Remove default overflow-x scroll
[TECH-1470](https://dhis2.atlassian.net/browse/TECH-1470) Right align audits

_before_
image

_after_
image


tomzemp:
@cooper-joe : requesting review from me in case you want to weigh in (see [slack discussion](https://dhis2.slack.com/archives/C02NHB92S2H/p1666951136597659))

Birkbjo (approved):

Mohammer5:
There are still some "issues" when the values have a different amount of decimals:
image

There are alternative ways of implementing this (e.g. with flex) so that it could look like this:
image

Right now I think that the correct implementation would be using sub grids which are not supported yet. So might not be worth looking into this any further than the PR already does. That's a decision that @cooper-joe has to make.

cooper-joe (approved):
I think `right-aligned` is OK. It does get weird with many different value lengths, but that's a downside generally when showing in a table column.
We can move to sub-grids when supported.

tomzemp:
We should also right-align deletes, right @cooper-joe?
image

I didn't include those initially, but have updated to change.

If you're okay with right aligning deletes, I'll merge/close

cooper-joe:
> We should also right-align deletes, right @cooper-joe?

Yes, let's do that 👍

PR #238, opened 2022-10-31
fix: handle remove periods calculation for end-of-month [DHIS2-14007]

tomzemp:
Adds fix for [DHIS2-14007](https://dhis2.atlassian.net/browse/DHIS2-14007):

The underlying issue is that we find the current period when we have openFuturePeriods, and to find the current period, we look for the period that falls in the range (now - 1 period) to now. 

However, for monthly periods, the JavaScript date calculation logic (`date.setMonth(date.getMonth()-1)`) will not work properly on some monthly boundaries (e.g 31 October subtract 1 month with this approach, yields 1 October because 31 September does not exist).

I've implemented a fix for this, but if we have a standard approach for getting periodId for the current date, it might be worth it instead to change our logic for getting the current period?

I noticed some other small issues on periods when reviewing this (e.g. if you open a monthly dataset on 15 January 2023, January 2023 will not be open, but the period selector will open in 2023; there appear to be some discrepancies in openFuturePeriods calculations because of variability in endDates for months, etc.,). I'm not sure if it also makes sense to address these, or just to wait until some point when we might standardize period calculation logic.


src/shared/fixed-periods/fixed-periods.js
```diff
@@ -76,8 +96,42 @@ const createUpdateFullPeriodTimeToDate = (multiplier) => {
...
+}
+
+const monthlyPeriodTypes = Object.keys(monthlyCounts)
+
+const getPreviousMonthNumber = (startMonth, periodType) => {
+    const monthsToRemove = monthlyCounts[periodType]
+    if (startMonth - monthsToRemove < 0) {
+        return startMonth - monthsToRemove + 12
+    }
+    return startMonth - monthsToRemove
```
Mohammer5, line 115: calculation could be put in a variable

Mohammer5 (approved):
I don't know of any standard way of dealing with this. This must've been encountered by core frontend developer already though. Could you put a question into the #frontend channel? If there is no response or no solution, then we can go ahead with this implementation and merge it

src/shared/fixed-periods/fixed-periods.js
Mohammer5: After having read my comment again, I myself have no idea what I was trying to say here.. I'll simply resolve this

tomzemp: I think you were suggesting to put `startMonth - monthsToRemove` calculation in a variable as it's used three times? I'm happy to update that, I just hadn't done any work/updates on data entry app for a while 🙂

PR #241, opened 2022-11-07
fix(indicators): handle expressions that cannot be evaluated

HendrikThePendric:


Birkbjo (approved):

cooper-joe:
Let's keep it feature-parity with the old app, don't have enough info to make any kind of change here. Thanks for the ping 👍

PR #271, opened 2022-12-06
test(data value store): cover with test & establish testing pattern

Mohammer5:
Closes DHIS2-14235 ([link](https://dhis2.atlassian.net/browse/DHIS2-14235))

# Issues with the "standard" approach

There is no standard appraoch, only a suggestion how to approach resetting the store with a mock in the official docs. Testing a zustand store was a lot more complicated than I thought for several reasons:

### Zustand doesn't have a built-in way to set an initial state when testing

There's no `setState` method on stores unless explicitly added. This way we can't set a state explicitly at the beginning of each test without using the hook (as the stores use the `create` method which return a react hook) and hook testing utilities. That seems to complicate tests which is far from ideal.

### Zustand doesn't have a built-in way to set / reset a store's state to the initial state

Zustand provides a way how to implement a mock that automagically resets the state between tests:
https://docs.pmnd.rs/zustand/guides/testing

There are a few issues with this approach due to the app-platform, so this was not viable.
One issue, which I was able to find a fix for, is that the proposed jest import doesn't work.
For some reason we have to use `jest.requireActual('zustand').default` (which is not the case in a new CRA app, I tested it)

### App platform uses the mock in every test

The reason why I couldn't follow that approach is that the mock was run in every test, even though I only used `jest.mock('zustand')` in a single test so far. This caused every test that implicitly imported a file that creates a store (e.g. by importing `src/shared/index.js`) to throw an error. I have now idea why exactly that happens or what exactly causes the error, but it didn't happen in a fresh CRA app, so I don't even want to try to fix this within the app

# The current solution

So.. I'm not really sure how to test this within the app platform. The current solution that seems to work is: Use a combination of `createStore` as well as `useStore` instead of `create` (allows us to synchronously call `.setState` at the beginning of a test)

We'd have to export the default state as well as the store from the module, too. And we'd have to refactor all stores if we go for this approach.

One upside of this approach is that we're explicitly clearing the store between tests in each test. That's repetitive, but it makes tests much easier to reason about. That's an upside because tests as a diagnosis tool when something goes wrong should be as understandable as possible, which is not the case when you have to ask yourself why the store's state has been reset without **knowing** that a mock automagically does that

---

I think the way suggested in this PR is the better alternative compared against the mock and implicit clearing of state


src/shared/stores/data-value-store.js
```diff
@@ -35,3 +38,7 @@ export const useValueStore = create((set, get) => ({
         return !!dataValue?.comment
     },
 }))
+
+export function useValueStore(selector) {
+    return useStore(valueStore, selector)
+}
```
Birkbjo, line 44: You don't actually have to do this, even though it's fine, it's not necessary - you can call `setState` and `getState` on the returned hook as well.

```js
export const useValueStore = create(....)

useValueStore.setState()
```

You could also use probably do something like this to not have to import `initialState` in tests:
```js

describe('data value store', () => {
    const initialState = useValueStore.getState()
    beforeEach(() => {
        useValueStore.setState(initialState)
    })
```

Birkbjo (approved):
Looks good to me. Minor comment that may simplify it a little bit.

src/shared/stores/data-value-store.js
Mohammer5: Ah, that's great! How did you figure that out? I couldn't find an api reference in the docs

Birkbjo: > Ah, that's great! How did you figure that out? I couldn't find an api reference in the docs

A combination of looking at the implementation of the "automagic" testing-link:
import actualCreate from 'zustand' 
```js
const create = (createState) => {
  const store = actualCreate(createState) // this is using default `create` from zustand which is the same for hooks
  const initialState = store.getState()
  storeResetFns.add(() => store.setState(initialState, true))
  return store
}
```

That made me just check the type returned:
image

PR #285, opened 2023-01-09
test(use update limits): cover hook with tests

Mohammer5:
Closes DHIS2-14379 ([link](https://dhis2.atlassian.net/browse/DHIS2-14379))

This PR includes the first test that will test mutations with optimistic updates and therefore will define the standard of how to test optimistic updates. **Please look at this test extra carefully so we can hash out how we want to do this and whether my proposed solution is actually an OK one or not**.

### DIFF
--- a/src/data-workspace/min-max-limits-mutations/use-update-limits.test.js
+++ b/src/data-workspace/min-max-limits-mutations/use-update-limits.test.js
@@ -5,26 +5,33 @@ import { dataValueSets } from '../../shared/use-data-value-set/query-key-factory
 import { Wrapper, useTestQueryClient } from '../../test-utils/index.js'
 import useUpdateLimits from './use-update-limits.js'
 
-jest.mock('../../shared/use-data-value-set/use-data-value-set-query-key.js', () => ({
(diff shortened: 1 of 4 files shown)


src/data-workspace/min-max-limits-mutations/use-update-limits.test.js
```diff
@@ -0,0 +1,157 @@
...
+                dataElement: 'data-element-id',
+                orgUnit: 'orgUnitId',
+                minValue: 2,
+                maxValue: 3,
+            })
+        })
+
+        await waitFor(() => result.current.isSuccess)
+
+        const { queries } = queryClient.queryCache
```
Birkbjo, line 58: The `QueryCache`-class has a find method where you can provide the queryKey directly. See https://tanstack.com/query/v4/docs/react/reference/QueryCache#querycachefind

Mohammer5: Thanks @Birkbjo, updated the code!

Mohammer5:
Superseeded by #313

PR #287, opened 2023-01-09
fix: adjust disabled org unit style

cooper-joe:
Fixes [UX-110](https://dhis2.atlassian.net/browse/UX-110).

This PR changes the style of disabled (not available for selection) org units in the org unit selector:
- Icons are removed,
- Tooltip is shown on hover,
- Icon "legend" at bottom of menu removed,

A small change is made to the menu itself:
- Menu design is changed to remove margin around all content.

Note: the tooltip reuses the available i18n label `Dataset is not assigned to this organisation unit`, even though the label _should_ use `Data set` in two words, but it didn't seem worth the i18n change.

[UX-110]: https://dhis2.atlassian.net/browse/UX-110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomzemp (approved):
lgtm

PR #290, opened 2023-01-16
feat(option-set): add multi text support

Birkbjo:


src/data-workspace/inputs/option-set.js
```diff
@@ -81,22 +118,22 @@ export const OptionSet = ({
...
                     ))}
-                </SingleSelect>
+                </SelectComponent>
             </div>
             {input.value && (
                 <Button
                     small
                     secondary
                     className={styles.whiteButton}
-                    {...input}
```
Birkbjo, line 99: Not sure why this was needed?

src/data-workspace/inputs/option-set.js
```diff
@@ -21,21 +27,43 @@ export const OptionSet = ({
     onFocus,
     disabled,
     locked,
+    multi,
 }) => {
+    const parse = (value) => {
+        console.log('parse', value)
```
tomzemp, line 33: delete?

src/data-workspace/inputs/option-set.js
```diff
@@ -21,21 +27,43 @@ export const OptionSet = ({
...
     const {
         input,
         meta: { data },
     } = useField(fieldname, {
         subscription: { value: true, data: true },
+        format:
+            // format to an array when multi, since component needs an array
+            multi
+                ? (value) => {
+                      console.log({ value })
```
tomzemp, line 51: delete?

src/data-workspace/inputs/option-set.js
```diff
@@ -21,21 +27,43 @@ export const OptionSet = ({
...
+    }
+
     const {
         input,
         meta: { data },
     } = useField(fieldname, {
         subscription: { value: true, data: true },
+        format:
+            // format to an array when multi, since component needs an array
+            multi
```
tomzemp, line 49: this could be a bit more concise, e.g.
`multi ? (value) => ((value && value.split(',')) || []) : undefined`
or similar, but obviously just a stylistic choice up to you

src/data-workspace/inputs/inputs.module.css
```diff
@@ -61,6 +61,10 @@
     flex-grow: 1;
 }
 
+.select {
+    max-width: 150px;
```
tomzemp, line 65: Hmm...we do want this. I mean I get that it's a bit distracting for the width to grow, but it does help show more of the selections?

src/data-workspace/inputs/option-set.js
tomzemp: Was it to pass `name` to the Button? Or it might have just been there because we were generally passing all the input props down in the other input types?

src/data-workspace/inputs/option-set.js
Birkbjo: Yup, originally did it like that, but changed to explicit return for the log statement while debugging.

src/data-workspace/inputs/inputs.module.css
Birkbjo: Yeah I'm a bit unsure what to do here. This was just an initial pass on the styling. Should probably consult @cooper-joe , but we definitely cannot show all options. And I do think it could be a bit distracting that it grows so much. Can probably be a bit wider though.

Another option would be to enable "wrapping" and have it grow vertically instead of horizontally. Also probably a better approach for printing the form, as we would need to include all options in print-format.

src/data-workspace/inputs/option-set.js
Birkbjo: Will keep it like this, since we now sort the result of format. And I think something like 
```js
((value && value.split(MULTI_TEXT_SEPERATOR)) || []).sort(sortByOptionsOrder)
```

is harder to read.

src/data-workspace/inputs/inputs.module.css
```diff
@@ -78,3 +83,23 @@
...
+@media print {
+    .hideForPrint {
+        composes: hideForPrint from '../../app/app.css'
+    }
+    /* hide icons when printing  */
+    .select svg {
+        visibility: hidden;
+    }
+
+    .select [data-test="dhis2-uicore-select-input"] {
```
Birkbjo, line 95: Not sure if it's "hacky" or bad to use the data-test as selectors... But I think the alternative of selecting nested divs are worse.

HendrikThePendric: Yeah agree that this is better than relying on the DOM structure. 👍

src/data-workspace/inputs/option-set.js
```diff
@@ -11,6 +18,15 @@ import {
...
 import { InputPropTypes } from './utils.js'
 
+const MULTI_TEXT_SEPERATOR = ','
+
+// This is used to preserve the order of the optionSet for the selected options
+const createSortByOptionsOrder = (options) => (a, b) => {
+    const aIndex = options.findIndex((option) => option.code === a)
+    const bIndex = options.findIndex((option) => option.code === b)
+    return aIndex - bIndex
+}
```
HendrikThePendric, line 28: Pretty sure this is OK because in the end it's just to parse a few selected values. But this function is super expensive.

src/data-workspace/inputs/option-set.js
```diff
@@ -21,21 +37,54 @@ export const OptionSet = ({
...
+                    value
+                        .sort(sortByOptionsOrder)
+                        .join(MULTI_TEXT_SEPERATOR)) ||
+                ''
+            )
+        } else {
+            // Empty values need an empty string
+            return value || ''
+        }
+    }
```
HendrikThePendric, line 61: You could consider using an implicit return, conditional chaining and `??` instead of `||`.
```suggestion
    const parse = (value) =>
        multi
            ? value?.sort(sortByOptionsOrder).join(MULTI_TEXT_SEPERATOR) ?? ''
            : value ?? ''
```
I find it a bit more readable.

HendrikThePendric (approved):
I haven't tested this, but the code looks good to me.

src/data-workspace/inputs/option-set.js
Birkbjo: Yeah, you're right. But we currently don't have any searching functionality in options, so hopefully the optionSets shouldn't be that huge. I did consider converting to a map-object, but decided it's probably not worth it. 
Also, this only happens when actually changing the selection. So there would need to be a lot of options for this to actually matter.

Birkbjo: We could of course find both indexes in one pass, by doing something like this.

```js
const createSortByOptionsOrderOnePass = (options) => (a, b) => {
    let aIndex, bIndex
    for (let i = 0; i < options.length; i++) {
        if (aIndex && bIndex) {
            break
        }
        const option = options[i]
        if (option.code === a) {
            aIndex = i
        }
        if (option.code === b) {
            bIndex = i
        }
    }
    return aIndex - bIndex
}
```

src/data-workspace/inputs/option-set.js
Birkbjo: I can add the name. `focus` styling are working. Not sure what functionality we would want for `onFocus` and `onBlur` on this button?

src/data-workspace/inputs/option-set.js
Birkbjo: The reason for `value && value.sort` is that `value &&` will not be true in case of empty string. `value?.sort...` will throw on empty string.

PR #294, opened 2023-02-13
fix: metadata and user info error handling

KaiVandivier:
Previously, the app would crash without a helpful error message

Before:
Screenshot 2023-02-13 at 5 02 31 PM

After:
Screenshot 2023-02-13 at 5 02 58 PM


KaiVandivier:
Yes you're right, the same applies for the user info -- fixed too, thanks!

PR #300, opened 2023-02-25
feat: form expiry info [DHIS2-14600]

KaiVandivier:
Implements [DHIS2-14600](https://dhis2.atlassian.net/browse/DHIS2-14600)
also [DHIS2-14218](https://dhis2.atlassian.net/browse/DHIS2-14218)

**_(Resolved, but kept for discussion history)_**
@cooper-joe I kinda forgot the UI spec while I was working on this and put it in the bottom bar, but I think it's nice because it's a bit less invasive down below -- what do you think about that? Should I move it to the top of the form, where the "Data set locked" message might show? I also just used a localized date string instead of a relative time to avoid some challenges, e.g. wondering if "Closes in 1 day at 1:00am" means 1am tonight or 1am the day after -- I'm wondering what you think of that too

**_Response:_**
> For the timestamp, I would think a more precise relative date may help here... we do need to a way to show the timestamp if needed. We could do so when hovering the "Closes in..." element, with text "This data set closes and will be locked at $fullTimestamp".

Result:
Screenshot 2023-05-09 at 3 28 55 PM

**An edge case**
Since we're letting the backend do the work of checking for lock exceptions or super user exceptions normally, there is a case where a form has `expiryDays` but won't _actually_ be locked for the user after the expiry days, yet the front end will still show the lock date. Once that date has passed though, the lock date won't be shown. I'm wondering if that's an acceptable edge case, or if we should check for super user and lock exceptions on the frontend to avoid that case.

**Design**
I decided to integrate the `lockDate` checks in with `checkLockStatus` to avoid repeating much of that logic and duplicating calculations -- an effect of this was refactoring the `lockStatus` value in the `LockContext` from a string into an object with both the `state` string and the `lockDate`

[DHIS2-14600]: https://dhis2.atlassian.net/browse/DHIS2-14600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-14218]: https://dhis2.atlassian.net/browse/DHIS2-14218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/bottom-bar/main-tool-bar.js
```diff
@@ -81,23 +82,23 @@ export default function MainToolBar() {
...
             <ErrorSummary />
+            <FormExpiryInfo />
             {data?.completeStatus?.lastUpdatedBy && (
                 <span
                     className={cx(styles.completionSummary, styles.toolbarItem)}
                 >
-                    <span
-                        className={cx(styles.icon, styles.goToInvalidValueIcon)}
-                    >
+                    <span className={cx(styles.icon)}>
```
KaiVandivier, line 90: This style seemed unused

src/bottom-bar/main-tool-bar.js
```diff
@@ -81,23 +82,23 @@ export default function MainToolBar() {
...
-                                : i18n.t('Last incompleted by')}
-                        </span>
-                        {data?.completeStatus.lastUpdatedBy}
+                        {data.completeStatus.complete
+                            ? i18n.t('Last completed by {{-username}}', {
+                                  username: data?.completeStatus.lastUpdatedBy,
+                              })
+                            : i18n.t('Last incompleted by {{-username}}', {
+                                  username: data?.completeStatus.lastUpdatedBy,
+                              })}
```
KaiVandivier, line 101: I added the username to the string here for i18n because I think the username might not end up in the last position in every translation -- this freed up an extra span and CSS class

src/data-workspace/data-workspace.js
```diff
@@ -29,17 +27,8 @@ export const DataWorkspace = ({ selectionHasNoFormMessage }) => {
...
-    const initialDataValuesFetch = useDataValueSet({
-        onSuccess: (data) => {
-            updateLockStatusFromBackend(
-                frontEndLockStatus,
-                data?.lockStatus,
-                setLockStatus
-            )
-        },
-    })
+    const initialDataValuesFetch = useDataValueSet()
```
KaiVandivier, line 31: Since `useCheckLockStatus` uses a `useEffect` dependency on `dataValueSet.data?.lockStatus`, the update from the backend resulted in two updates -- see also `use-check-lock-status.js`

src/shared/locked-status/use-check-lock-status.js
```diff
@@ -10,96 +12,138 @@ import { useDataValueSet } from '../use-data-value-set/use-data-value-set.js'
...
-        return
-    }
-
-    // a lock status of 'OPEN' from the backend could mean either that the form is open OR
-    // that the form should be locked due to data input period, SO
-    // set to OPEN unless frontend check has identified that data input period as out-of-bounds
-    if (frontEndLockStatus !== LockedStates.LOCKED_DATA_INPUT_PERIOD) {
-        setLockStatus(LockedStates.OPEN)
-    }
-}
```
KaiVandivier, line 105: This was mostly integrated in `useCheckLockStatus` due to the `useEffect` dependency on `dataValueSet.data?.lockStatus`, so tried to let that handle the backend check

cooper-joe:
> @cooper-joe I kinda forgot the UI spec while I was working on this and put it in the bottom bar, but I think it's nice because it's a bit less invasive down below -- what do you think about that? Should I move it to the top of the form, where the "Data set locked" message might show? I also just used a localized date string instead of a relative time to avoid some challenges, e.g. wondering if "Closes in 1 day at 1:00am" means 1am tonight or 1am the day after -- I'm wondering what you think of that too

- I think the placement in the bottom bar is OK. 
- For the timestamp, I would think a more precise relative date may help here. In your example, if the time was 23:00 then a more precise relative date would be "Closes in 2 hours" vs. "Closes in 1 day".
    -  If we do this relative display, we do need to a way to show the timestamp if needed. We could do so when hovering the "Closes in..." element, with text "This data set closes and will be locked at $fullTimestamp".

cooper-joe:
> Since we're letting the backend do the work of checking for lock exceptions or super user exceptions normally, there is a case where a form has expiryDays but won't actually be locked for the user after the expiry days, yet the front end will still show the lock date. Once that date has passed though, the lock date won't be shown. I'm wondering if that's an acceptable edge case, or if we should check for super user and lock exceptions on the frontend to avoid that case.

This is a bit confusing, but I think it might be OK. Technically, the form will be locked at that date, just not for this user that has an exception. So it seems useful to show that the form will be effected by `expiryDays`. I'm really unsure if users with lock exceptions are aware of that fact though, which is hard to know. The lock date could end up causing confusing if they aren't aware.

tomzemp:
I haven't gone through and done a code review @KaiVandivier , but wanted to add some context here:

- the frontend checks exist independently of the backend checks, so that a user could be offline and still know if the form is locked. I believe that is why some of the code is kind of weird here and was not centralized in the use-check-lock-status hook. 
- we never decided if we were going to also check expiryDays on the front end. There's an open ticket for that: [TECH-1428](https://dhis2.atlassian.net/browse/TECH-1428). It was mostly that we were missing lockExceptions when I started the work, so I didn't include them and then we decided it wasn't worth the time at the end to modify the check when we already had a backend check. Doing this client-side, we could prevent people from opening a closed form when offline (though this seems like a relatively minor concern)
- the logic for data input periods is actually incorrect in this app and in the old app. It needs to be aligned with the backend: [DHIS2-14218](https://dhis2.atlassian.net/browse/DHIS2-14218). If you are making changes, it might be useful to address that one at the same time?

src/bottom-bar/main-tool-bar.js
tomzemp: Agree on the translation point, but I'm not sure the i18next scanner handles optional chaining well?

I get this error when starting:
```
Generating internationalization strings... 
i18next-scanner: Unable to parse code "{
                                  username: data?.completeStatus.lastUpdatedBy,
                              }"
```

I think I remember the same issue when working on the data-exchange-app, and I just defined the interpolation value outside of the i18n translation. I think this might be resolved in more recent versions of i18next-scanner, so it could also be an issue of updating cli-app-scripts?

HendrikThePendric: To build on what @tomzemp said, I think it's important to consider whether or not we actually needed conditional chaining here in the first place. And in this case we didn't because if we consider the surrounding code we find that this part is only going to be rendered when `data?.completeStatus?.lastUpdatedBy` is _truthy_: 
https://github.com/dhis2/aggregate-data-entry-app/blob/8a7c0633adcf1c10ab2239769cfd4c0a578ff4e7/src/bottom-bar/main-tool-bar.js#L86-L104

So in this case we can simply remove the conditional chaining operator `?`.

BTW this problem was introduced prior to the current PR and in my opinion is a great illustration of why, in general, we should only apply conditional chaining (or conditional log in general really) when we know it is needed. Adding `?.`, `||` and `&&` without a specific case in mind might make the code look very robust, but it is a recipe for disaster in the long run. See this Wikipedia page about [fail-fast systems](https://en.wikipedia.org/wiki/Fail-fast#:~:text=In%20systems%20design%2C%20a%20fail,continue%20a%20possibly%20flawed%20process.) for a more detailed explanation.

HendrikThePendric (approved):
Looks good to me. There is one `i18n` issue to address, but after this I feel this is good to go.

src/bottom-bar/main-tool-bar.js
KaiVandivier: Yeah, that's right 🙂

src/bottom-bar/main-tool-bar.js
KaiVandivier: Ah yeah, thanks to both for pointing this out -- I think removing the conditional chaining for the reason Hendrik described is good. Good point about fail-fast systems too 👍

KaiVandivier:
I believe I addressed the comments above and wouldn't mind a last look over 🙂

I ended up including [DHIS2-14218](https://dhis2.atlassian.net/browse/DHIS2-14218) because it seemed to make sense, but also chose not to include [TECH-1428](https://dhis2.atlassian.net/browse/TECH-1428) because it is best supported by a change on the backend to include `lockExceptions` in the `dataEntry/metadata` response -- I made a Jira issue for that

src/shared/locked-status/use-check-lock-status.js
```diff
@@ -10,96 +12,161 @@ import { useDataValueSet } from '../use-data-value-set/use-data-value-set.js'
...
 
-    return (
-        openingDate > currentDateAtServerTimeZone ||
-        closingDate < currentDateAtServerTimeZone
-    )
+        if (currentDate.serverDate < expiryDate.serverDate) {
+            // Take the sooner of the two possible lock dates
+            clientLockDate = clientLockDate
+                ? new Date(Math.min(clientLockDate, expiryDate.clientDate))
+                : expiryDate.clientDate
```
tomzemp, line 105: I guess you could have logic here about the user so that if they are superuser (have ALL authorities), you would not record a clientLockDate that is associated with expiryDays? (e.g. so you don't display the info about it closing if it's because of expiryDays and they're a superuser) (we're already getting the authorities, so I guess adding a check for ALL authorities would be pretty minor).

src/shared/locked-status/use-check-lock-status.js
```diff
@@ -10,96 +12,161 @@ import { useDataValueSet } from '../use-data-value-set/use-data-value-set.js'
...
+        // selectedPeriod.endDate is a string like '2023-02-20' --
+        // convert that to a date and get time to get ms to add expiry days.
+        // Also add one day more because selectedPeriod.endDate is the START
+        // of the period's last day (00:00), and we want the end of that day
+        // (confirmed with backend behavior).
+        // Converting the selectedPeriod.endDate string to an object creates
+        // it in the client's timezone; parse it from there.
+        const expiryDate = fromClientDate(
+            new Date(
+                new Date(selectedPeriod.endDate).getTime() +
```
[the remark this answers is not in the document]
KaiVandivier: Aha, thanks for pointing this out -- I'll look into this 🤔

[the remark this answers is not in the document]
KaiVandivier: Ah I think I get it -- I guess this was accidentally working because server was in UTC time, so the `new Date('2023-02-23')` happened to give the right client date. I pushed a commit to parse the date from the correct server date instead, like in other places in the code here

src/shared/locked-status/use-check-lock-status.js
KaiVandivier: Yeah that's pretty easy -- added it 🙂

KaiVandivier:
> Re: checking lockExceptions: these actually are available at api/dataEntry/lockExceptions (...) It's a point of consideration though whether it's worth it to improve the client-side/offline checks with lockExceptions info (if they're going to be bulky to retrieve)

That's great -- thanks for pointing that out! I think adding those will increase the scope of this PR enough that it would warrant its own PR, so I'll leave that to the one for [TECH-1428](https://dhis2.atlassian.net/browse/TECH-1428).

I did add the user authority check for the "will lock" date, but still omitted any actual locks from the frontend while we don't have lockExceptions implemented, so that we don't falsely block a user while offline

tomzemp (approved):
lgtm! 🚀

PR #311, opened 2023-04-05
fix: include attribute options in completion post [DHIS2-15032]

tomzemp:
Fixes the completion for data sets with attribute combos.

Note the backend requires attribute options in an array of options (I'd think it might be better for the backend to accept a concatenated string of these options, i.e. `cp:someUID0001;someUID0002;someUID0003` so that it matches the request format for dataValues post, but I don't think it's worth it to request a change as part of this fix)

*before*
image

### DIFF
--- a/src/shared/use-data-value-set/query-key-factory.js
+++ b/src/shared/use-data-value-set/query-key-factory.js
@@ -5,14 +5,14 @@ export const dataValueSets = {
         periodId,
         orgUnitId,
         categoryComboId,
-        categoryOptionIds = [],
+        categoryOptionIds,
     }) => {
         const params = {
             ds: dataSetId,
             pe: periodId,
             ou: orgUnitId,
             cc: categoryComboId,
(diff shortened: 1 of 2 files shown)


src/shared/use-data-value-set/query-key-factory.js
```diff
@@ -5,14 +5,14 @@ export const dataValueSets = {
...
-        categoryOptionIds = [],
+        categoryOptionIds,
     }) => {
         const params = {
             ds: dataSetId,
             pe: periodId,
             ou: orgUnitId,
             cc: categoryComboId,
-            cp: categoryOptionIds.join(';'),
+            cp: categoryOptionIds && categoryOptionIds.join(';'),
```
tomzemp, line 15: this change is to have cp be undefined rather than an empty string; the empty string would be included in the payload and cause errors; it seemed easier to adjust here rather than adding logic in generating the payload in use-set-form-completion-mutation.js

PR #315, opened 2023-04-12
fix: cover useSetFormCompletionMutation with tests

Mohammer5:
Fixes [DHIS2-14378](https://dhis2.atlassian.net/browse/DHIS2-14378)
PR title has `fix` as there's a fix-commit

[DHIS2-14378]: https://dhis2.atlassian.net/browse/DHIS2-14378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/shared/completion/use-set-form-completion-mutation.test.js
+++ b/src/shared/completion/use-set-form-completion-mutation.test.js
@@ -71,7 +71,7 @@ describe('useSetFormCompletionMutation', () => {
 
         expect(dataValueSet.current.isLoading).toBe(true)
 
-        // Mutate a current data value
+        // Complete a form
         act(() => setFormCompletion.current.mutate({ completed: true }))
 
         await waitForSetFormCompletion(() => {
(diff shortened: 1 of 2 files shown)


src/shared/completion/use-set-form-completion-mutation.test.js
```diff
@@ -0,0 +1,249 @@
...
+                        }}
+                    >
+                        {children}
+                    </Wrapper>
+                ),
+            })
+
+        expect(dataValueSet.current.isLoading).toBe(true)
+
+        // Mutate a current data value
```
tomzemp, line 74: correct comment?

src/shared/completion/use-set-form-completion-mutation.test.js
```diff
@@ -0,0 +1,249 @@
...
+            expect(dataValuesResolver).toHaveBeenCalledTimes(1)
+
+            // useDataValueSet is not loading but never has finished, which
+            // verifies that the active query has been cancelled
+            expect(dataValueSet.current.isLoading).toBe(false)
+            expect(onDataValueSuccess).toHaveBeenCalledTimes(0)
+        })
+    })
+
+    it('should optimistically update the completion state', async () => {
```
tomzemp, line 96: I don't get some things with this test. Why mock the a datavalueset request? I would assume that we do not need to test that this mutation is cancelled as it's covered in the previous test? I would think that the test (from the description) would just be testing that the completion state in the queryCache goes from false to true?

src/shared/completion/use-set-form-completion-mutation.test.js
```diff
@@ -0,0 +1,249 @@
...
+        })
+
+        const { result: setFormCompletion, waitFor: waitForSetFormCompletion } =
+            renderHook(() => useSetFormCompletionMutation(), {
+                wrapper: ({ children }) => (
+                    <Wrapper
+                        queryClientOptions={{ queryCache }}
+                        dataForCustomProvider={{
+                            'dataEntry/dataSetCompletion': () =>
+                                new Promise((_, reject) => {
```
tomzemp, line 198: could you just return Promise.reject to not have to deal with the fake timers?

tomzemp (commented):
overall looks good, but I wasn't sure about whether some of the assertions in the tests were necessary (see questions)?

src/shared/completion/use-set-form-completion-mutation.test.js
Mohammer5: I'm doing this so I can inspect the query cache before the error occurs.
This way we know that the cache has actually been updated. If that doesn't happen, then `expect(cachedDataValues.completeStatus).toBeUndefined()` could result in a false positive (as the value would be undefined if something else cause the hook to not work for some other, completely unrelated, reason)

src/shared/completion/use-set-form-completion-mutation.test.js
Mohammer5: True. I've removed the unnecessary code

src/shared/completion/use-set-form-completion-mutation.test.js
tomzemp: Got it

tomzemp (approved):
lgtm 🛸

PR #319, opened 2023-04-20
chore: update cypress to v12 and upgrade related configs & code

Mohammer5:
Blocked by dhis2/cli-utils-cypress#393

* [x] GH workflows need to be adjusted


cypress/e2e/context-selection/period/index.js
```diff
@@ -1,4 +1,10 @@
-import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+import { Given, When, Then } from '@badeball/cypress-cucumber-preprocessor'
+
+// It seems that the new cypress cucumber library splits different steps into
+// different test cases. Cypress does not persist aliases across tests,
+// mutable seem to be working though
+let selectedPeriodLabel = ''
+let selectedPeriodValue = ''
```
ismay, line 7: This should still be possible from what I could find:

- https://github.com/badeball/cypress-cucumber-preprocessor/discussions/959
- https://github.com/badeball/cypress-cucumber-preprocessor/pull/74

I think a mutable value used in this way is an easy way to have tests accidentally affect eachother. Using as should be possible so that seems preferable to me.

ismay: Unless something has changed in the meantime, but I couldn't find any docs confirming that.

ismay: These issues all seem to confirm to me that it should work:

- https://github.com/badeball/cypress-cucumber-preprocessor/issues/246
- https://github.com/badeball/cypress-cucumber-preprocessor/issues/551
- https://github.com/badeball/cypress-cucumber-preprocessor/issues/471
- https://github.com/badeball/cypress-cucumber-preprocessor/issues/413
- https://github.com/badeball/cypress-cucumber-preprocessor/issues/374
- https://github.com/badeball/cypress-cucumber-preprocessor/issues/246

Hopefully there's something in there that helps resolve the aliasing issue

package.json
```diff
@@ -26,23 +28,24 @@
         "jgs:start": "BROWSER=none yarn start --proxy https://debug.dhis2.org/dev"
     },
     "devDependencies": {
+        "@cypress/webpack-preprocessor": "^5.17.0",
```
ismay, line 31: Shouldn't this be provided by the cypress-plugins package?

Mohammer5: Well, it's only required when using cucumber, it's a peer dependency of the cucumber plugin. The cypress-cucumber-preprocessor is marked as a peer dependency as well as the app needs to import and use it directly. So I thought they both should be peer dependencies. Would you then move both deps to the cypress plugins lib?

ismay: Ah I see, I didn't know they were peerDeps of `@dhis2/cypress-plugins`. In that case this makes sense yeah. One thing you could do in the `@dhis2/cypress-plugins` package is mark any peerdependencies that aren't always necessary as optional: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta

> When a user installs your package, npm will emit warnings if packages specified in peerDependencies are not already installed. The peerDependenciesMeta field serves to provide npm more information on how your peer dependencies are to be used. Specifically, it allows peer dependencies to be marked as optional.

Depending on whether you want users to be able to omit installation of `cypress-cucumber-preprocessor` and the webpack processor if they're not using cucumber.

PR #332, opened 2023-05-31
test: cover remaining metadata selectors

Mohammer5:
Closes [DHIS2-15329](https://dhis2.atlassian.net/browse/DHIS2-15329)

[DHIS2-15329]: https://dhis2.atlassian.net/browse/DHIS2-15329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/shared/metadata/selectors.js
+++ b/src/shared/metadata/selectors.js
@@ -349,7 +349,10 @@ export const getNrOfColumnsInCategoryCombo = createCachedSelector(
     getCategoriesByCategoryComboId,
     (categories) => {
         if (!categories) {
-            // @TODO: Is this really correct? Shouldn't it be 0?
+            // It returns the number of columns, not the number of
+            // categoryOptions or categories, and there should always be 1
+            // column to render. In case of missing references, it will just
+            // render a padding-cell
             return 1
         }
 
@@ -406,6 +409,12 @@ export const getComputedCategoryOptionIdsByCatComboId = createCachedSelector(
 )((_, categoryComboId) => categoryComboId)
 
 /**
+ * The `categoryOptionCombos` of a `categoryCombo` is a Set, and therefore
+ * unordered. To be able to render the cocs (eg. an entry field) in the correct
+ * order (same order as the table-header), we need to compute the
+ * categoryOptionCombos client side. This is basically a `cartesian-product` of
+ * all the `categoryOptions` in all `categories` in the `categoryCombo`.
+ *
  * @param {*} metadata
  * @param {*} categoryComboId
  */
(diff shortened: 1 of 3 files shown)


src/shared/metadata/selectors.js
```diff
@@ -340,10 +347,16 @@ export const getGroupedDataElementsByCatCombo = createSelector(
...
 export const getNrOfColumnsInCategoryCombo = createCachedSelector(
     getCategoriesByCategoryComboId,
-    (categories) =>
-        (categories?.map((cat) => cat.categoryOptions.length) || [1]).reduce(
-            (total, curr) => total * curr
-        )
+    (categories) => {
+        if (!categories) {
+            // @TODO: Is this really correct? Shouldn't it be 0?
+            return 1
```
Mohammer5, line 353: @Birkbjo should this be 1? I would've expected that this should be 0

src/shared/metadata/selectors.js
```diff
@@ -398,10 +412,14 @@ export const getComputedCategoryOptionsByCatComboId = createCachedSelector(
 export const getSortedCoCsByCatComboId = createCachedSelector(
     (metadata) => metadata,
```
Birkbjo, line 413: While we're updating this code I think it's nice to have a comment as to why this is necessary. Since it seems like you would have access to the `categoryOptionCombos` directly from the `categoryCombo`.

```
/* The `categoryOptionCombos` of a `categoryCombo` is a Set, and therefore unordered. To be able to render the cocs (eg. an entry field) in the correct order (same order as the table-header), we need to compute the categoryOptionCombos client side. This is basically a `cartesian-product` of all the `categoryOptions` in all `categories` in the `categoryCombo. */
```

src/shared/metadata/selectors.js
Mohammer5: Alright

src/shared/metadata/selectors.js
```diff
@@ -398,10 +412,14 @@ export const getComputedCategoryOptionsByCatComboId = createCachedSelector(
...
     (_, categoryComboId) => categoryComboId,
-    getComputedCategoryOptionsByCatComboId,
-    (metadata, categoryComboId, computedCOCS) =>
-        computedCOCS.map((options) =>
-            getCoCByCategoryOptions(metadata, categoryComboId, options)
+    getComputedCategoryOptionIdsByCatComboId,
+    (metadata, categoryComboId, computedCategoryOptionIds) =>
+        computedCategoryOptionIds.map((categoryOptionIdVector) =>
```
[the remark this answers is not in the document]
Mohammer5: I like `categoryOptionIdsForCoc` and `categoryOptionIds` :ok_hand: Used those

src/shared/metadata/selectors.js
Mohammer5: I added the comment

Birkbjo (approved):

PR #350, opened 2023-08-03
fix: handle indicator decimals [DHIS2-15631]

tomzemp:
This adds functionality to round indicators if they have decimals specified for their output.

**Before**
image

**After**
image

Note: when I looked at the legacy data entry app seems to just round to 1 decimal if decimals were specified (and did not follow the actual number of decimals specified). I assume we don't want to do that. I checked against how the decimals were displaying in visualizer and tried to be consistent with that:
image
visualizer app is also not add trailing 0s if indicator has more specified decimals than necessary (e.g. if the indicator is 2/4 and it has 5 decimals specified, it will only display as .5 and not as .50000, so I have not added any trailing 0s for the data entry app)

### DIFF
--- a/src/data-workspace/indicators-table-body/use-indicator-value.js
+++ b/src/data-workspace/indicators-table-body/use-indicator-value.js
@@ -23,6 +23,7 @@ export const useIndicatorValue = ({
     factor,
     form,
     numerator,
+    decimals,
 }) => {
     const indicatorValueRef = useRef(null)
     const affectedDataElementsLookup = useMemo(
@@ -57,6 +58,7 @@ export const useIndicatorValue = ({
                 numerator,
                 factor,
                 formState: form.getState(),
+                decimals,
             })
         }
 
@@ -68,5 +70,6 @@ export const useIndicatorValue = ({
         factor,
         form,
         numerator,
+        decimals,
     ])
 }
--- a/src/data-workspace/custom-form/replace-td-node.js
+++ b/src/data-workspace/custom-form/replace-td-node.js
@@ -12,6 +12,7 @@ const replaceIndicatorCell = (indicatorId, metadata) => {
         denominator,
         numerator,
         indicatorType: { factor },
+        decimals,
     } = metadata.indicators[indicatorId]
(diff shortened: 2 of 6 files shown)


Birkbjo (approved):
Looks very good to me! 

Thanks for checking how the visualizer app does this, and keeping them consistent!

src/data-workspace/indicators-table-body/use-indicator-value.js
```diff
@@ -57,6 +58,7 @@ export const useIndicatorValue = ({
                 numerator,
                 factor,
                 formState: form.getState(),
+                decimals,
```
Birkbjo, line 61: I know this wasn't added in this PR. However I'm pretty sure this `useMemo`, really should've been an `useEffect`. Because (at least from react18), we shouldn't read or write to the ref during render.

tomzemp: > Because (at least from react18, I don't think it was a "rule" before react18), we shouldn't read or write to the ref during render.

Agree that this could use a rewrite, but I was under the impression that the explicit rules in react18 were due to the concurrent rendering in react 18, which could cause the ref to fall out of sync?

I'm going to leave this code as is for now as I would prefer to work on other things rather than refactoring this. I don't really want to open a ticket either as it will end up being low priority. Maybe we should have a would-be-nice-to-do list for data entry app 😀

Birkbjo: > but I was under the impression that the explicit rules in react18 were due to the concurrent rendering in react 18, which could cause the ref to fall out of sync? 

Yeah, I believe so as well. So it shouldn't _really_ matter for us, but it could potentially be a cause of bugs in the future. But even in react 18, concurrent rendering is opt-in. 

> I'm going to leave this code as is for now as I would prefer to work on other things rather than refactoring this. I don't really want to open a ticket either as it will end up being low priority.

Agree, and it's not trivial to rework this either, especially while trying to follow all react rules and "best practices".

PR #357, opened 2023-09-25
fix(validator): limits as warning instead of error

Birkbjo:
https://dhis2.atlassian.net/browse/DHIS2-16284

Requested by Olav on slack. Previously min-max limits were implemented as "hard" limits, and thus the values would not be saved if the value was outside of the limit.

Context from Olav:
> I think this is problematic and should be changed:
min-max are generally defined as some statistical thresholds, so we have to expect that correct values could be outside the thresholds from time to time
data is typically entered from a paper form. It’s good that suspcious values are flagged, but it’s problematic if the data entry user is not able to enter “official” numbers written on the reporting forms
its inconsistent with how validation rules work: flagging issues, but allowing them to be saved
it looks like if a value has previously been entered in a cell, and it’s overwritten with a value outside the min/max thresholds, the old value is never removed from the database which can be misleading

I've now implemented `warnings` as a concept. `FinalForm` does not have an internal concept of "warnings" - validations are either an error that prevents saving - or no error. I've thus implemented a concept of "warnings". They're run in the `FinalForm`-validate function to keep the same behaviour and lifecycle as normal validations , but they do not influence the result. Since we already have a `zustand` store that stores the errors of the form (for sharing outside of form-context), I've implemented a similar structure for warnings as for errors. Using the exact same way as errors are used internally in FinalForm (with `getIn` and `setIn`). We can thus leverage this the exact same way as we do with errors - but currently they're only used for styling - and maybe later we could aggregate the warnings as we do with errors in the bottom-bar.

### DIFF
--- a/src/data-workspace/data-entry-cell/validation-tooltip.js
+++ b/src/data-workspace/data-entry-cell/validation-tooltip.js
@@ -6,18 +6,17 @@ import styles from './data-entry-cell.module.css'
 /**
  * These components adapt the UI Tooltip component to be used for the
  * validation feedback tooltip in DE cells, which has a few needs:
- * 1. Should only open when the cell state is invalid
- * 2. Should gracefully close when the cell becomes valid again
- * 3. Should not cause cell to lose focus when it becomes invalid
+ * 1. Should only open when enabled prop is true
+ * 2. Should gracefully close when the enabled props becomes false
+ * 3. Should not cause cell to lose focus when tooltip is opened
  * 4. Should open when focused and close when defocused
  */
-
 const TooltipManager = React.forwardRef(function TooltipManager(
-    { active, children, invalid, onMouseOut: close, onMouseOver: open },
+    { active, children, enabled, onMouseOut: close, onMouseOver: open },
     ref
 ) {
     React.useEffect(() => {
-        if (invalid && active) {
+        if (enabled && active) {
             open()
         }
         // When this cell becomes valid or defocused, close the tooltip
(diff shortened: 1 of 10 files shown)


src/shared/stores/entry-form-store.js
```diff
@@ -1,15 +1,26 @@
+import { setIn, getIn } from 'final-form'
 import create from 'zustand'
-
+import { devtools } from 'zustand/middleware'
```
Birkbjo, line 3: Note to self: remove

src/shared/stores/entry-form-store.js
```diff
@@ -1,15 +1,26 @@
...
+export const useEntryFormStore = create(
+    devtools((set, get) => ({
+        ...inititalState,
+        setErrors: (errors) => set({ errors: errors ?? {} }),
+        getErrors: () => get().errors,
+        getNumberOfErrors: () => countLeaves(get().getErrors()),
+        getWarnings: () => get().warnings,
+        getWarning: (fieldname) => getIn(get().getWarnings(), fieldname),
+        setWarning: (fieldname, warning) => {
+            const warnings = get().getWarnings() || {}
```
Birkbjo, line 18: Note to self: `|| {}` should be unecessary.

src/data-workspace/inputs/validators.js
```diff
@@ -87,13 +87,19 @@ export const validatorsByValueType = {
...
 
+    return composeValidators(...validators)
+}
+
+export const warningValidateByValueType = (valueType, limits) => {
+    const validators = []
+
     if (
         CAN_HAVE_LIMITS_TYPES.includes(valueType) &&
         Number.isFinite(limits?.min) &&
```
KaiVandivier, line 105: Finish refactoring ;) `composeValidators` is kinda janky to use on 0 or 1 validator

KaiVandivier (changes requested):
Just one request 🙂 the rest looks good!

src/data-workspace/data-entry-cell/validation-tooltip.js
```diff
@@ -57,16 +56,16 @@ const TooltipManager = React.forwardRef(function TooltipManager(
...
     active: PropTypes.bool,
     children: PropTypes.node,
-    invalid: PropTypes.bool,
+    enabled: PropTypes.bool,
     onMouseOut: PropTypes.func,
     onMouseOver: PropTypes.func,
 }
 
-export const ValidationTooltip = ({ children, content, invalid, active }) => {
+export const ValidationTooltip = ({ children, content, enabled, active }) => {
```
Birkbjo, line 64: Renamed `invalid` to `enabled`. After this PR `invalid`-state of the field does not describe the functionality of what this prop actually does. Actually, previously `syncError` was also used to control this prop. The prop actually controls if the tooltip logic is `enabled` - and thus I think that describes the functionality better.

KaiVandivier (approved):
🚀

PR #358, opened 2023-09-26
fix: prevent unnecessary renders when form has errors

Birkbjo:
While working on https://github.com/dhis2/aggregate-data-entry-app/pull/357 , I noticed the entire form was being re-rendered on each input. We've spent quite a lot of time to optimize this, but I realized it only happened when the form contained an error. The culprit was the logic to keep the zustand-store up to date with the form-errors. I've now moved this to its own component, so the hook won't re-render the entire form. 

Before: 

https://github.com/dhis2/aggregate-data-entry-app/assets/13852438/16cd515e-11a8-4693-b457-c8b36f27ecc0

After:

https://github.com/dhis2/aggregate-data-entry-app/assets/13852438/3e314895-5a0d-4394-8fde-b7e38332ed00

### DIFF
--- a/src/data-workspace/entry-form.js
+++ b/src/data-workspace/entry-form.js
@@ -68,8 +68,12 @@ export const EntryForm = React.memo(function EntryForm({ dataSet }) {
     )
 })
 
-/*  Used to sync store with errors from form
-    In its own component to prevent unecessarily re-renders in the tree */
+/**
+ * Used to sync global store with errors from final-form.
+ * Because of the `errors` subscription in `useFormState`, this re-renders
+ * on every form input change if there is an error in the form. Therefore, this
+ * has its own component to prevent unnecessary re-renders below it in the tree
+ */
 const EntryFormErrorSpy = () => {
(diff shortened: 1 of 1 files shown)


src/data-workspace/entry-form.js
```diff
@@ -72,12 +62,28 @@ export const EntryForm = React.memo(function EntryForm({ dataSet }) {
...
             )}
-
+            <EntryFormErrorSpy />
             <Component dataSet={dataSet} globalFilterText={globalFilterText} />
         </>
     )
 })
 
+/*  Used to sync store with errors from form
+    In its own component to prevent unecessarily re-renders in the tree */
```
KaiVandivier, line 72:
```suggestion
/**
 * Used to sync global store with errors from final-form.
 * Because of the `errors` subscription in `useFormState`, this re-renders
 * on every form input change if there is an error in the form. Therefore, this 
 * has its own component to prevent unnecessary re-renders below it in the tree
 */
```

KaiVandivier (approved):
I still find it weird the component is declared below where it is used 😉 but it works 🙌

PR #363, opened 2023-12-11
chore(report-portal): run reports only when env is setup

Birkbjo:
After report-portal was added, it was run even if the required environment-variables wasn't configured. This causes a lot of verbose logging and error-messages in the console.

Not sure if this is the best way, but I couldn't find any native way to do this in `agent-js-jest`. I'm a bit surprised the tool is actually running even if these variables are not defined.
We might want a different `jest.config` for CI and local testing instead?

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,34 @@
+const reportPortalConfig = [
+    '@reportportal/agent-js-jest',
+    {
+        apiKey: process.env.REPORTPORTAL_API_KEY,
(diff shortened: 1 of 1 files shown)


adeldhis2 (approved):
LGTM 👍
Thanks Birk for the fix!
I found also this link that could help:  [How to avoid local execution reported into project space](https://reportportal.io/docs/issues-troubleshooting/HowToAvoidLocalExecutionReportedIntoProjectSpace)

PR #367, opened 2024-02-20
feat(DHIS2-16132): transpose section forms

kabaros:
This PR implements the ability to display sections transposed, and grouped by category in the data entry app. Implements [DHIS2-16132 ](https://dhis2.atlassian.net/browse/DHIS2-16132).

[transpose-different-options.webm](https://github.com/dhis2/aggregate-data-entry-app/assets/1014725/cfe7ad48-f59a-484d-9129-5e3ccbcc9611)

## Default Section form - without pivoting (existing implementation)
image

### Form pivoted (data elements moved to columns)
image

### Single Category pivoted
image

### Multiple categories (with one pivoted)
image

### Data elements with different category combos
image

## Notes about implementation
As an general approach, I aimed to decouple the display details (table, rows, columns, column and row spans) from our internal concepts (data elements, categories, category options, cocs etc...), to provide a middle layer (a matrix) that represents a form in our context. This middle layer can become the base of a form builder, for implementations that are cross-platform and other improvements in the future.

With the previous implementation, we were more than half-way there, but the data elements were tightly coupled to rows, and in turn the rows/data elements are tightly coupled to functionality like search, filters, highlighting current cell in the table etc... The next step should be to consolidate this implementation with the "normal" section form - there is a followup ticket for this:  https://dhis2.atlassian.net/browse/LIBS-584

The only concept that remains metadata-aware is the `date entry field`  - this was the existing design and it works well, as a self-contained component responsible of its own state, validation etc... regardless of where it appears in a form.

- I avoided changing the original `CategoryComboTable` and added a new `PivotedCategoryComboTable` to handle the new use cases for transposing. I  didn't want to change something that already works and well-tested for what's essentially an experimental new feature (i.e. not part of achieving parity with the existing app). Next step is to consolidate theses two components and the underlying matrix generation.

- In the grouped matrix logic, I added a special case for when there is one category and another for 1+ categories. This is needed because we need a "shadow" row in the case of more categories, this is unnecessary for one category. See the photo below, the first one has a "shadow" row (location fixed/outreach) and the categories are displayed underneath each other - this is necessary for multiple categories. In the case of one category (second table), we can do  with the category being on top of the options.

image

## Background

We are aiming to add more form configuration options as part of an initiative to provide configurations natively to data entry forms to reduce the necessity for custom forms. Users are currently building custom forms as a workaround for shortcomings of the configuration options (ability to transpose, or customise a cell design) or implementation (such to avoid issues with RTL issues). This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to transpose forms and adding free-text descriptions were one of the main reasons people choose to go the custom forms route so we're tackling these first.

## Known limitations
- Transpose (not grouping) doesn't work for a dataset with more than one category combo
- Support multiple transposed categories (right now, we transpose only one)
- Search and filtering doesn't work as these were tightly couple with data elements - we will need to design and rethink that for transposed sections.

## ToDos (future)
- [ ] Consolidate CategoryComboTable and PivotedCategoryComboTable, as well as underlying "matrix" logic:  https://dhis2.atlassian.net/browse/LIBS-584

## Related PRs:
- [Maintenance app](https://github.com/dhis2/maintenance-app/pull/2802)
- [Data entry app (beta)](https://github.com/dhis2/aggregate-data-entry-app/pull/367)
- [Backend](https://github.com/dhis2/dhis2-core/pull/16562)


src/shared/use-context-selection/use-context-selection.js
```diff
@@ -27,6 +27,20 @@ export function usePeriodId() {
     return useQueryParam('periodId', PARAMS_SCHEMA.periodId)
 }
 
+export function useFeature(feature) {
```
kabaros, line 30: I added this "feature toggle" in order to be able to test the functionality without the backend and maintenance app deployed.

src/data-workspace/category-combo-table-body-pivoted/generate-form-matrix/__fixtures/pivot-test-data.js
kabaros, line 1: I needed a test that mimics what we do before (based on the whole metadata) - this file can/should be significantly reduced to only include what's relevant to the test but that wasn't easy to do. I have it in my list for the refactors after (as well as getting rid of snapshot tests)

src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js
```diff
@@ -0,0 +1,174 @@
...
+                                                styles.noWrap,
+                                            ]}
+                                            key={fieldInRow.id}
+                                            colSpan={fieldInRow.colSpan}
+                                            rowSpan={fieldInRow.rowSpan}
+                                        />
+                                    )
+                                }
+
+                                if (fieldInRow.type === 'de') {
```
kabaros, line 118: one thing about my implementation of the matrix generation is that it also generates the data element entry fields. The previous implementation had this as a component concern, and I think that was cleaner - so as part of the consolidating these components, I will also follow the previous approach.

Birkbjo: Sounds good. Yeah I think the data-structure could just return the Ids, and the component would get the dataElement in component-level? But it doesn't matter too much I think.

But I think it makes sense that it generates the structure for them - since now everything is intertwined. It's not as simple as looping over dataElements and categoryOptionsCombos anymore. One row is not always equal to the cocs.length anylonger, so I'm not sure how easy it is to do this on component level?
I guess the structure would have the rows (dataElements and categories), and the headers, and could loop from that info. But I think this is fine for now.

src/data-workspace/category-combo-table-body-pivoted/generate-form-matrix/generate-matrix-grouped-by-category.js
```diff
@@ -0,0 +1,248 @@
+export const generateMatrixGroupedByCategory = (options, groupedBy) => {
+    const columnHeaders = generateColumnHeaders(options, groupedBy)
+
+    // the category options that make part of the coc are defined in the field parentCategoryOptions
+    const columnCategoryOptions = columnHeaders
+        .flat()
```
Birkbjo, line 6: Doesn't really matter, but I guess this could've just been `flat(2)`, right?

src/data-workspace/category-combo-table-body-pivoted/generate-form-matrix/generate-matrix-transposed.js
```diff
@@ -0,0 +1,77 @@
...
+                cat.categoryOptions.includes(fco)
+            )
+
+            // todo(pivot): maybe refactor this logic after confirming this is Functional Design team wants UI-wise
+            const lastCateogry = categoryIndex === categories.length - 1
+            if (!alreadyAdded[categoryOption.id] || lastCateogry) {
+                dataEntryRow.push({
+                    id: categoryOption?.id,
+                    displayFormName: categoryOption?.displayName,
+                    rowSpan:
```
tomzemp, line 47: I'm not sure this is working as intended? When I try it with [Emergency Response data set](http://localhost:3000/#/?attributeOptionComboSelection=eLwL77Z9E7R-JvIqWKLPPkt&dataSetId=Lpw6GcnTrmS&orgUnitId=ImspTQPwCqd&periodId=202402&features=pivot), I get a misformed table (due to miscalculated rowspan): 
image

src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js
```diff
@@ -0,0 +1,174 @@
...
+                                                    [styles.dataElementRowHeader]:
+                                                        fieldInRow.metadataType ===
+                                                        'dataElement',
+                                                }
+                                            )}
+                                            colSpan={fieldInRow.colSpan}
+                                            rowSpan={fieldInRow.rowSpan}
+                                        >
+                                            {fieldInRow.displayFormName !==
+                                                'default' &&
```
tomzemp, line 98: I'm not sure if you want to use displayFormName here? It seems more logical to use formName. If there's a translation that is not default (e.g. into another language), the name will appear:
image

src/shared/use-context-selection/use-context-selection.js
tomzemp: This makes it easy for testing 😅, but I'm not sure we actually would want to expose a query parameter like this? I think people who design data sets would not want their users to arbitrarily transpose things (I assume we're not going to advertise this, but I guess it's possible that people share links if we expose it?)

There aren't really checks for this, so providing invalid categories causes some issues, but maybe that's not of concern if it's just for testing?

tomzemp (approved):
Looks cool as a new feature, and seems like it's working for the most part (with some issues in how the transposition is actually being rendered). 

I'm approving now given the time frame, but I'm not sure if it is really necessary to update this app in time for the soft freeze, or if would be better to resolve the existing issues before merging in?

kabaros:
@tomzemp thanks for testing - I fix the issue with the transposed form with one default combo or one category. Below is how it looks. Also removed the cool 🕶️ feature toggle.

image

src/data-workspace/category-combo-table-body-pivoted/generate-form-matrix/generate-matrix-grouped-by-category.js
kabaros: the second .flat() was unnecessary actually.

src/data-workspace/category-combo-table-body-pivoted/generate-form-matrix/generate-matrix-transposed.js
kabaros: fixed

src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js
kabaros: fixed

PR #373, opened 2024-03-06
feat(DHIS2-17002): display custom text before and after a section

kabaros:
This PR adds the ability to configure custom HTML text before and after a section. This implements [DHIS2-17002](https://dhis2.atlassian.net/browse/DHIS2-17002)

# Implementation notes
- I added a library to sanitise the html content before displaying it. I wanted to just use `react-html-parser` initially as is the case with custom forms, but that doesn't sanitise HTML properly (beyond stripping script tags), and the authors of the library themselves recommended using [DomPurify](https://github.com/peternewnham/react-html-parser?tab=readme-ov-file#what-doesnt-react-protect-me-from)

- Related maintenance PR: https://github.com/dhis2/maintenance-app/pull/2812
- 
[text_before_after_section.webm](https://github.com/dhis2/aggregate-data-entry-app/assets/1014725/fc905226-876c-4006-b201-f4b3ce7450d0)


kabaros:
@michael-markevich this story might have security implications. We're allowing users to add html content to display before and after a data entry form. I've used [a library](https://github.com/cure53/DOMPurify) here that sanitises the HTML and only allows certain HTML tags (including links but it takes care of xss through href).

This sanitisation happens when displaying the form - I am wondering whether we should be actually sanitising  the content when configuring it in the [maintenance app](https://github.com/dhis2/maintenance-app/pull/2812), or is it enough to sanitise when displaying it? I am just wary of implicitly changing something people enter in the maintenance app.

KaiVandivier (approved):
LGTM! 👍 

Edit: actually added some type comments below 🙂

src/data-workspace/section-form/section-description.js
```diff
@@ -0,0 +1,24 @@
...
+    return (
+        <div
+            className={styles.sectionDescription}
+            dangerouslySetInnerHTML={{ __html: html }}
+        ></div>
+    )
+}
+
+SectionDescription.propTypes = {
+    children: PropTypes.node.isRequired,
```
KaiVandivier, line 23:
```suggestion
    children: PropTypes.node,
```

Oh actually this will cause a prop warning if a section doesn't have a description configured, which may be common; maybe this shouldn't be required? (minor)

KaiVandivier:
Also have a look at the display options types: https://github.com/dhis2/aggregate-data-entry-app/blob/72c96bf4e3c144e0d05908868a43a1c7f773e547/src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js#L149-L154

An object seems expected but a string is used, and the names of the before and after text seems different

kabaros:
> (PS why is `displayOptions` a string instead of an object? Just curious)

thanks for the review @KaiVandivier  - `displayOptions` is a string because of a limitation of the backend hibernate mapping, where we'd need to either have it as a string, or map it into Java objects - discussion is [here](https://github.com/dhis2/dhis2-core/pull/16562#issuecomment-1976385763).

src/data-workspace/section-form/section-description.js
kabaros: yup, you're right - updated now

KaiVandivier (approved):

PR #375, opened 2024-04-26
fix(section-form): prevent crash when displayOptions are not present

Birkbjo:
Fixes a bug where the app would crash if used on a server version where `section.displayOptions` are not present. 
We could've just checked for the presence before parsing, but I wrapped `JSON.parse` in a try-catch, because `displayOptions` is just an arbitrary string - so it could fail in various ways.

We also should introduce some error-boundaries to the app, so the entire app doesn't crash on errors like this.

### DIFF
--- a/src/data-workspace/section-form/displayOptions.js
+++ b/src/data-workspace/section-form/displayOptions.js
@@ -1,0 +1,25 @@
+const defaultDisplayOptions = {
+    beforeSectionText: '',
+    pivotMode: 'n/a',
+    afterSectionText: '',
+    pivotCategory: null,
+}
+export const getDisplayOptions = (section) => {
+    if (!section) {
+        return defaultDisplayOptions
+    }
+
+    try {
+        const { displayOptions: displayOptionString } = section
+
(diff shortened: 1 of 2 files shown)


src/data-workspace/section-form/displayOptions.js
```diff
@@ -0,0 +1,25 @@
...
+}
+export const getDisplayOptions = (section) => {
+    if (!section) {
+        return defaultDisplayOptions
+    }
+
+    try {
+        const { displayOptions: displayOptionString } = section
+
+        const displayOptions = JSON.parse(displayOptionString)
```
Birkbjo: I would've liked to validate the actual fields here, since it's an arbitrary string. But that would entail adding a validation lib, and a decision of what to use... Not sure if it's worth it since it seems we're checking for equality before using them. Except for `pivotedCategory` which could be used more defensively.

kabaros (approved):

kabaros:
I would say null. That Was the initial intention at least.

PR #389, opened 2024-06-27
feat: add vertical tab rendering of sections in a data set

flaminic:
Implements [DHIS2-17506](https://dhis2.atlassian.net/browse/DHIS2-17506).
This PR switched the direction of a dataset section tabs if the user has configures the dataset this way (in separate PR). 

### Implementation details

The displayOptions was added as a JSON type in the BE to allow us flexibility to add more options without updating the back-end, similarly to what was done for sections. See discussion in https://github.com/dhis2/dhis2-core/pull/16562 

### Background

We are aiming to add more form configuration options as part of an initiative to provide configurations natively to data entry forms to reduce the necessity for custom forms. Users are currently building custom forms as a workaround for shortcomings of the configuration options (ability to transpose, or customise a cell design) or implementation (such to avoid issues with RTL issues).

This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to show section tabs vertically were one of the main reasons people choose to go the custom forms route so we're tackling these first.

### UI

The ui is still work in progress but it will look like similar to this: 

https://github.com/dhis2/dhis2-core/assets/5418349/06c571eb-916a-4714-8c0e-a0a4ba8c7515

### Related PRs: 

https://github.com/dhis2/dhis2-core/pull/17891

### DIFF
--- a/src/data-workspace/section-form/section-form.js
+++ b/src/data-workspace/section-form/section-form.js
@@ -1,4 +1,5 @@
 import { Tab, TabBar } from '@dhis2/ui'
+import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
 import { useSectionFilter } from '../../shared/index.js'
@@ -11,12 +12,17 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
         ? dataSet.sections.filter((s) => s.id === sectionId)
         : dataSet.sections
 
+    const { displayOptions: displayOptionString } = dataSet
+    const displayOptions =
+        displayOptionString && JSON.parse(displayOptionString)
+
     if (dataSet.renderAsTabs) {
         return (
             <TabbedSectionForm
                 globalFilterText={globalFilterText}
                 sections={dataSet.sections}
                 dataSetId={dataSet.id}
+                direction={displayOptions?.tabsDirection}
             />
         )
     }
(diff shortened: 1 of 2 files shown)


src/data-workspace/section-form/section-form.js
```diff
@@ -11,12 +12,17 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
         ? dataSet.sections.filter((s) => s.id === sectionId)
         : dataSet.sections
 
+    const { displayOptions: displayOptionString } = dataSet
+    const displayOptions =
+        displayOptionString && JSON.parse(displayOptionString)
```
flaminic, line 17: handle the JSON.parse

tomzemp (approved):
looks great to me! I like how you've achieved this with some css updates, so that we don't need to wait on a ui change for now 😎

I assume that you'd going to merge this after the backend change? If we're also updating the maintenance app, it would be nice to update the maintenance app to differentiate between the tab orientation and the "Render vertically" option (which is not relevant to the Data Entry Beta App because it only applies to multi-orgunit forms which are not implemented in the newer app). Otherwise, I guess people could potentially get  confused between these.

image

PR #390, opened 2024-07-08
feat: make data sets sections collapsable

flaminic:
Implements [DHIS2-17507](https://dhis2.atlassian.net/browse/DHIS2-17507).
This PR allows for sections of a data set to be collapsible

### Implementation details

We have decided that all sections should be collapsible by default, and this for now will not be a setting in the maintenance app 

Collapsible is applied only when the  " render as tabs" option is not turned on for the specific dataset

### Background

We are aiming to add more form configuration options as part of an initiative to provide configurations natively to data entry forms to reduce the necessity for custom forms. Users are currently building custom forms as a workaround for shortcomings of the configuration options (ability to transpose, or customise a cell design) or implementation (such to avoid issues with RTL issues).

This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to collapse sections were one of the main reasons people choose to go the custom forms route so we're tackling these first.

### UI

This is a video of the feature: 

https://github.com/user-attachments/assets/0ff743da-9ae0-4ab2-9232-0a670cd62809

[DHIS2-17507]: https://dhis2.atlassian.net/browse/DHIS2-17507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/data-workspace/section-form/section-form.js
+++ b/src/data-workspace/section-form/section-form.js
@@ -3,7 +3,6 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
 import { useSectionFilter } from '../../shared/index.js'
-import { useFeature } from '../use-feature.js'
 import { SectionFormSection } from './section.js'
 import styles from './section.module.css'
 
@@ -24,7 +23,6 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
 
     const { displayOptions: displayOptionString } = dataSet
     const displayOptions = parseDisplayOptions(displayOptionString)
-    const featureToggleCollapsible = useFeature('collapsible')
 
     if (dataSet.renderAsTabs) {
         return (
@@ -45,7 +43,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
                     dataSetId={dataSet.id}
                     key={s.id}
                     globalFilterText={globalFilterText}
-                    collapsible={featureToggleCollapsible}
+                    collapsible
                 />
             ))}
         </>
--- a/src/data-workspace/section-form/section-form.js
+++ b/src/data-workspace/section-form/section-form.js
@@ -43,6 +43,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
                     dataSetId={dataSet.id}
                     key={s.id}
                     globalFilterText={globalFilterText}
+                    collapsible
                 />
             ))}
         </>
--- a/src/data-workspace/section-form/section-form.js
+++ b/src/data-workspace/section-form/section-form.js
@@ -31,6 +31,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
                 sections={dataSet.sections}
                 dataSetId={dataSet.id}
                 direction={displayOptions?.tabsDirection}
+                collapsible={displayOptions?.collapsibleSections}
             />
         )
     }
@@ -43,7 +44,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
                     dataSetId={dataSet.id}
                     key={s.id}
                     globalFilterText={globalFilterText}
-                    collapsible
+                    collapsible={displayOptions?.collapsibleSections}
                 />
             ))}
         </>
@@ -53,6 +54,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
 SectionForm.propTypes = {
     dataSet: PropTypes.shape({
         displayOptions: PropTypes.shape({
+            collapsibleSections: PropTypes.bool,
             tabsDirection: PropTypes.oneOf(['vertical', 'horizontal']),
         }),
         id: PropTypes.string,
--- a/src/data-workspace/section-form/section-form.js
+++ b/src/data-workspace/section-form/section-form.js
@@ -3,6 +3,7 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
 import { useSectionFilter } from '../../shared/index.js'
+import { useFeature } from '../use-feature.js'
 import { SectionFormSection } from './section.js'
 import styles from './section.module.css'
 
@@ -23,6 +24,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
 
     const { displayOptions: displayOptionString } = dataSet
     const displayOptions = parseDisplayOptions(displayOptionString)
+    const featureToggleCollapsible = useFeature('collapsible')
 
     if (dataSet.renderAsTabs) {
         return (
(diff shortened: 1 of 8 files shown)


src/data-workspace/section-form/section.js
```diff
@@ -88,62 +106,80 @@ export function SectionFormSection({ section, dataSetId, globalFilterText }) {
...
-                                            'Placeholder section description'}
+                                <div>
+                                    <div className={styles.title}>
+                                        {section.displayName}
                                     </div>
-                                )}
+                                    {section.description && (
+                                        <div className={styles.description}>
+                                            {section.description ||
+                                                'Placeholder section description'}
```
tomzemp, line 116: this isn't your code, but it seems like we can just get rid of this, given that we already have the conditional check to display only if `section.description` (also, I assume we don't want to display generic text)?

tomzemp (commented):
Looks good! 🪗

I did have a couple of general questions:
- I'm not sure we'd want to disable the collapsing of the section if user has selected a section from context menu:
image
I feel like there's no reason to enable the collapsing of a section if you
- otherwise, I share the general opinion that Joe (or Kim/Karoline/Aarti) should maybe review to approve styling of collapsed sections and such

I'm not really sure about the use of the query parameter for allowing section toggling (I mean, I like it in terms of testing 😄, but not sure how it works for users to actually use in the app). Personally, I feel like this is a fairly non-intrusive UI change, such that maybe we could just it enable it by default (provided that sections are not display as tabs)? I feel like not having the query parameter also helps out in terms of the code organization in the sense that all of our query parameters would remain defined in use-context-selection.js (and related to context selection).

I can approve if you want to merge (the code looks good 🤩), but maybe we could just allow the toggling by default and then potentially update the app again when/if we decide that we want to toggle this feature based on a property of the data set?

src/data-workspace/section-form/section.js
flaminic: good catch

flaminic:
> Looks good! 🪗
> 
> I did have a couple of general questions:
> 
>     * I'm not sure we'd want to disable the collapsing of the section if user has selected a section from context menu:
> 
> 
> image I feel like there's no reason to enable the collapsing of a section if you - otherwise, I share the general opinion that Joe (or Kim/Karoline/Aarti) should maybe review to approve styling of collapsed sections and such
> 
> I'm not really sure about the use of the query parameter for allowing section toggling (I mean, I like it in terms of testing 😄, but not sure how it works for users to actually use in the app). Personally, I feel like this is a fairly non-intrusive UI change, such that maybe we could just it enable it by default (provided that sections are not display as tabs)? I feel like not having the query parameter also helps out in terms of the code organization in the sense that all of our query parameters would remain defined in use-context-selection.js (and related to context selection).
> 
> I can approve if you want to merge (the code looks good 🤩), but maybe we could just allow the toggling by default and then potentially update the app again when/if we decide that we want to toggle this feature based on a property of the data set?

Hey, yes the feature toggle was added in order to hide the feature to users (since there are still unknowns on whether thins should be enabled by default and/or there should be a setting in the maintenance app etc). That decision is pending on a conversation scheduled for September. But i hear you, maybe there is  no point in merging this until that is decided. I would say that we can leave this PR oper, close the JIra ticket and create a follow up ticket to finish this up. What do you think ?

tomzemp:
> . I would say that we can leave this PR oper, close the JIra ticket and create a follow up ticket to finish this up. What do you think ?

I think it's good to put the PR on hold until design team weighs in 👍. I'd probably just let them weigh in on the existing ticket, but sounds good also to close it and create a new one if that's easier for tracking progress.

cooper-joe:
https://github.com/dhis2/aggregate-data-entry-app/pull/392 covers my design/style suggestions.

Functionally, I think it would be ideal if sections retained their width when collapsed. Moving the collapse control to the start of the section heading makes this OK in the UI as the collapse control will not be hidden when the section is overflowing horizontally. Is this possible without too much of a hack? (I assume collapsed content is removed from the DOM, so we no longer have the width?)

flaminic:
> #392 covers my design/style suggestions.
> 
> Functionally, I think it would be ideal if sections retained their width when collapsed. Moving the collapse control to the start of the section heading makes this OK in the UI as the collapse control will not be hidden when the section is overflowing horizontally. Is this possible without too much of a hack? (I assume collapsed content is removed from the DOM, so we no longer have the width?)

Thanks Joe.I have made an attempt
[here](https://www.loom.com/share/11c8933f03ab4816bdafd59d2d3c8af0?sid=9fa9874b-5954-4c7c-ac9e-8f9f46be7518) Is this is what you were thinking?

cooper-joe:
> Thanks Joe.I have made an attempt
[here](https://www.loom.com/share/11c8933f03ab4816bdafd59d2d3c8af0?sid=9fa9874b-5954-4c7c-ac9e-8f9f46be7518) Is this is what you were thinking?

Yes, that's what I had in mind. I think this works OK.

src/data-workspace/section-form/section.js
```diff
@@ -88,44 +106,61 @@ export function SectionFormSection({ section, dataSetId, globalFilterText }) {
...
-                                    {section.displayName}
+                                <div
+                                    className={styles.collapseIcon}
+                                    tabIndex={collapsible ? 0 : -1}
+                                    onClick={onSectionHeadClicked}
+                                    onKeyDown={onSectionHeadEntered}
+                                >
+                                    {collapsible &&
+                                        (showSectionContent ? (
+                                            <IconChevronUp16 color="white" />
```
tomzemp, line 117: we also have a constant value for white, in case we're worried that our design system is ever going to change DHIS2 white 😅 https://github.com/dhis2/ui/blob/master/constants/src/colors.js#L74

src/data-workspace/section-form/section-form.js
```diff
@@ -51,9 +52,7 @@ export const SectionForm = ({ dataSet, globalFilterText }) => {
 
 SectionForm.propTypes = {
     dataSet: PropTypes.shape({
-        displayOptions: PropTypes.shape({
-            tabsDirection: PropTypes.oneOf(['vertical', 'horizontal']),
-        }),
+        displayOptions: PropTypes.string,
```
tomzemp, line 55: not related to this ticket, but it looks like we also have some logic for parsing display options here (https://github.com/dhis2/aggregate-data-entry-app/blob/b48e1ff23e0a608f4d53a725b0ea220ab6227f48/src/data-workspace/section-form/displayOptions.js#L7), so we could probably just have one function for this parsing, and do it at the level. It looks like that getDisplayOptions gets used in the SectionFormSection child component.

tomzemp (approved):
Looks great to me 🚀!

I didn't know of `visibilty: collapse` 😎

PR #395, opened 2024-09-11
fix: handle java language format [DHIS2-18031]

tomzemp:
See https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-18031

### Background 
This ticket converts keyUiLocale from the Java format (e.g. pt_BR) to the ISO format (pt-BR). The app is currently crashing because the functions to generate periods within multi-calendar-dates fails when passed a locale string.

_Note:_ We also have two UI locales that include script (`uz_UZ_Cyrl`, `uz_UZ_Latn`), which in valid ISO format would be (`uz-Cyrl-UZ`, `uz-Latn-UZ`), but these cause problems in multi-calendar-dates even when passed in the proper ISO format, so for not I am just removing the potential script tag and converting to `uz-UZ` (probably need a follow up for multi-calendar-dates to handle languages with scripts) 

**Before**
image

**After** 
image

### Testing
Manual: I have tested with Portuguese (Brazilian), Arabic (Egyptian), English, Portuguese, Arabic, French, and the two varieties of Uzbek

Automated: I have added some tests for the useUserInfo hook to test that the conversion of the language code occurs as expected.

### DIFF
--- a/src/shared/use-user-info/use-user-info.js
+++ b/src/shared/use-user-info/use-user-info.js
@@ -8,8 +8,25 @@ const queryOpts = {
     staleTime: 24 * 60 * 60 * 1000,
 }
 
+const parseLocale = (unparsedLocale) => {
+    if (typeof unparsedLocale !== 'string') {
+        return 'en'
+    }
+
+    // while script is possible in DHIS2 locales, it is not supported in underlying multi-calendar-dates logic
+    const [lng, region] = unparsedLocale?.replaceAll('_', '-').split('-')
+
+    return region ? `${lng}-${region}` : lng
+}
+
 export const useUserInfo = () => {
     const userInfoQuery = useQuery(queryKey, queryOpts)
+    // this maps java format locales like `pt_BR` to JS style `pt-BR`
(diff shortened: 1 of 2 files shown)


src/shared/use-user-info/use-user-info.js
```diff
@@ -8,8 +8,25 @@ const queryOpts = {
     staleTime: 24 * 60 * 60 * 1000,
 }
 
+const parseLocale = (unparsedLocale) => {
+    if (typeof unparsedLocale !== 'string') {
+        return 'en'
+    }
+
+    // while script is possible in DHIS2 locales, it is not supported in underlying multi-calendar-dates logic
+    const [lng, region] = unparsedLocale?.replace(/_/g, '-').split('-')
```
tomzemp, line 17: the language codes returned from the backend should use `_` instead of `-`, but I'm doing a replace before split in case any start being returned in the ISO format

tomzemp:
Closing because the bug of the app crashing was indirectly fixed by https://github.com/dhis2/aggregate-data-entry-app/pull/397 (and the supporting updates to the multi-calendar library)

PR #401, opened 2024-10-03
ci: update workflows

ghost:
https://dhis2.atlassian.net/browse/CLI-71

This app had a unique cypress setup. I've moved it to use the same setup as our other platform apps. That meant:

- removing the custom reporters (https://docs.cypress.io/guides/tooling/reporters#Reporter-Options, https://github.com/reportportal/agent-js-cypress). See https://github.com/dhis2/aggregate-data-entry-app/pull/349 and https://github.com/dhis2/aggregate-data-entry-app/pull/352
- removing the cypress cloud integration
- removing the custom matrix logic

I can imagine that the removal of the custom reporter would be controversial to some. We currently don't use it in any of our (team platform) other apps, from what I've seen thus far. We could add it back, but since our workflows logic is now centralized, when we add it back we should probably add it back for all apps instead of just this one. Or add it conditionally at the cost of extra complexity. Since the workflows were blocking Tom's work I propose we remove it for now and add it back properly if we all feel it's worth the added complexity (and if we're not blocking any requirements from any other teams, etc.).

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,34 +1,3 @@
-const reportPortalConfig = [
-    '@reportportal/agent-js-jest',
-    {
-        apiKey: process.env.REPORTPORTAL_API_KEY,
-        endpoint: process.env.REPORTPORTAL_ENDPOINT,
-        project: process.env.REPORTPORTAL_PROJECT,
-        launch: 'aggregate_data_entry_app_master',
-        attributes: [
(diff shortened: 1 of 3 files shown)


tomzemp (approved):
thanks 🙏

Mohammer5 (approved):

PR #418, opened 2024-11-05
feat: add title and subtitle too all form types

flaminic:


tomzemp (approved):
Looks good to me. I don't know if you want to update any of the descriptions (e.g. like styles.sectionsCustomText) (both the name and the location of the styles import) to not mention sections? It doesn't make a functional difference but it just seems like the reference to sections is now not relevant.

PR #421, opened 2024-11-12
feat: when we have multiple sections, default to first section 

flaminic:
This Pr is meant to address the performance problems that occur when forms with many fields are rendered.
In this PR, we mitigate the problem by rendering by default only one section. This is a temporary fix until a proper fix will be in place.

### DIFF
--- a/src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
+++ b/src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
@@ -28,8 +28,7 @@ export default function SectionFilterSelectorBarItem() {
 
     useEffect(() => {
         const sections = dataSet?.sections
-        // select first section if renderAsTabs === true
-        if (dataSet?.renderAsTabs && !sectionFilter && sections?.length) {
+        if (sectionFilter === undefined && sections?.length) {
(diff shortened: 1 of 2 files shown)


src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.js
```diff
@@ -55,9 +54,7 @@ export default function SectionFilterSelectorBarItem() {
 
     const selectableOptions = dataSet?.renderAsTabs
         ? sectionOptions
-        : [{ value: undefined, label: i18n.t('All sections') }].concat(
-              ...sectionOptions
-          )
+        : [...sectionOptions, { value: null, label: i18n.t('All sections') }]
```
Birkbjo, line 57: Here you're having null as value, while checking strictly for undefined above. Is that intentional?

flaminic: Ah yes, i decided to take advantage of the philosophical differentiation between underlined and null, where null in this case is an intentional absence, where undefined is not set yet. Does it make sense to you?

Birkbjo (approved):

PR #425, opened 2024-11-25
fix: remove FinalForm dependency in main form [DHIS2-18373]

tomzemp:
This addresses performance issues associated with Final Form by removing the FinalForm dependency from the main form.

Instead of using form/field state, the field inputs have been refactored 
- to keep values as state variables
- to set a state variable `valueSynced: true` when the cell's current state value matches the last synced value
- active state

To simplify things, I've also used useHighlightedField in some places where we had looked for active field. This could be updated, but apart from the difference in border color of the cell, I wasn't sure if the differentiating between active/highlighted was so relevant. (It does slightly change the calculation of indicators/totals, but I think this is pretty minor and some of our active state management seemed a bit buggy on some input types)


src/data-workspace/custom-form/replace-input-node.js
```diff
@@ -29,7 +29,7 @@ export const replaceInputNode = (domNode, metadata) => {
     // TODO: This was already there when I started on this branch
     // Need to check with Kai what his intentions were with it.
     // const cellProps = getCellPropsByInputType(inputType)
-
+    // console.log(domNode)
```
Birkbjo, line 32: Remove?

src/data-workspace/data-entry-cell/data-entry-field.js
```diff
@@ -33,6 +36,9 @@ export const DataEntryField = React.memo(function DataEntryField({
             disabled={disabled}
             locked={locked}
             highlighted={highlighted}
+            valueSynced={valueSynced}
+            dirty={true}
```
Birkbjo, line 40: Remove `dirty`?

src/data-workspace/indicators-table-body/compute-indicator-value.js
```diff
@@ -83,13 +82,14 @@ const parseExpressionTemplate = (expression, values) => {
...
     return matches.reduce((substitudedExpression, match) => {
         const operand = match.replace(/[#{}]/g, '')
-        const { categoryOptionComboId } = parseFieldOperand(operand)
+        const { dataElementId, categoryOptionComboId } =
+            parseFieldOperand(operand)
         const value =
             (categoryOptionComboId
                 ? // For COCs we read the data directly from the form values
-                  getIn(values, operand)
+                  dataValues?.[dataElementId]?.[categoryOptionComboId]?.value
```
Birkbjo, line 90: Would it make sense to have a getter for this? There is already in the store, but that is "bound", but we could potentially extract it to a function, there's at least two places we do this deep "getter".

src/data-workspace/inputs/boolean-radios.js
```diff
@@ -1,103 +1,72 @@
...
-    const clearField = useFieldWithPatchedOnFocus(fieldname, {
-        type: 'radio',
-        value: '',
-        subscription: { value: true },
-    })
+    const [value, setValue] = useState(() => convertBooleanValue(initialValue))
+    const [lastSyncedValue, setLastSyncedValue] = useState(() =>
+        convertBooleanValue(initialValue)
+    )
+    const [syncTouched, setSyncTouched] = useState(false)
```
Birkbjo, line 24: In "genric-input" we just do `setValueSynced(value === newValue)`. What is the reason that this and other fields need the effect and `syncTouched`? Could we not calculate this? Or should the `generic-input` also use this way of doing it?

src/shared/use-blurred-field.js
```diff
@@ -1,13 +1,14 @@
...
-    const { active } = useFormState({ subscription: { active: true } })
 
-    return useMemo(() => {
-        const blurredField = previouslyActiveFieldRef.current
-        previouslyActiveFieldRef.current = active
-        return blurredField
-    }, [active])
+    const { dataElementId, categoryOptionComboId } =
+        useHighlightedFieldStore((state) => state.getHighlightedField()) || {}
+    const fieldId = `${dataElementId}.${categoryOptionComboId}`
```
Birkbjo, line 9: Could we use `getFieldId` here?

Birkbjo (commented):
This looks great! Thanks a lot for getting into this so quickly!

src/data-workspace/form-wrapper.js
```diff
@@ -21,29 +20,34 @@ function createInitialValues(dataValueSet) {
...
-            mutators={{ setFieldData }}
-        >
-            {() => children}
-        </Form>
+
+    const setInitialDataValues = useValueStore(
+        (state) => state.setInitialDataValues
     )
+
+    const initialValuesInStore =
```
tomzemp, line 34: This is the logic to check that the values in the zustand store are for the given data form, so as to prevent rendering the input cells before the initial values are present

src/data-workspace/indicators-table-body/compute-indicator-value.js
tomzemp: I'm not sure if having a getter would improve things here as this is a function outside of a component/hook?

src/data-workspace/inputs/boolean-radios.js
tomzemp: I think it's more correct to do the side effect with useEffect. In the generic input, which is the first one I wrote, this `setValueSynced(value === newValue)` was working, but in the other components, it does not work, I suppose because the code is executing before the new value has actually been set.

Birkbjo: Alright, I just in general try to not use `useEffect`, and always try to look for other solutions. It adds additional re-renders and complexity. But maybe in this case it's warranted - because we're actually syncing states here. 

But I do feel like the need for "syncTouched", "lastSyncedValue", and "setValueSynced" is somewhat complex. 
Ideally this is just something we could calculate from mutation state, and potentially the query-cache. But it's hard to argue for such changes when this seems to work great.

src/data-workspace/form-wrapper.js
Birkbjo: I found a small (but scary) bug that I think is related to this, but also the service worker somehow.

If you edit a cell - blur and let it save - then refresh. The value will not be shown in the form. It seems like we're getting data from the service worker that is not updated.

src/data-workspace/form-wrapper.js
```diff
@@ -21,29 +20,34 @@ function createInitialValues(dataValueSet) {
...
+        useValueStore((state) => state.getInitialDataValues())?.formKey ===
+        validFormKey
+
+    useLayoutEffect(() => {
```
Birkbjo, line 38: I don't think it matters too much here, but I think normal `useEffect` should suffice.

src/data-workspace/inputs/date-input.js
```diff
@@ -15,47 +14,49 @@ export const DateInput = ({
...
 }) => {
+    const [value, setValue] = useState(initialValue)
+    const [lastSyncedValue, setLastSyncedValue] = useState(initialValue)
+    const [syncTouched, setSyncTouched] = useState(false)
+
+    useEffect(() => {
+        if (syncTouched) {
+            setValueSynced(value === lastSyncedValue)
+        }
+    }, [value, lastSyncedValue, syncTouched])
```
Birkbjo, line 32: If exhaustive deps?

Birkbjo (commented):
Thanks a lot for fixing this! This is a lot of work and very impressive!

I've mainly a few comments about lint-issues and a bug that needs to be fixed. 

I do feel like some of the sync logic could be hoisted to "DataEntryField". Eg. we could have a `lastSyncedValue`-state there. Then we could update that value, instead of a "synced"-boolean. And compare `value` (from the store) with lastSyncedValue. Right now we're comparing the state in the field, but it's the data in the dataValueStore that is actually "the source of truth", right? And in theory these could diverge? 
But it might be that would complicate things - because I do see some logic to compare the values for eg. the `file`-field. So I won't block this PR because of this.

src/data-workspace/inputs/date-input.js
tomzemp: 🤦‍♂️ sorry for that. I'm impressed(?) with myself for managing to miss multiple lint warnings both locally and in github 😬

src/data-workspace/form-wrapper.js
tomzemp: 😱 thanks for that!

I'm too be honest a little confused with the react-query nomenclature here. I would have thought that since we had `refetchOnMount: false` for this query that `isFetchedAfterMount` would remain false, but I guess it's 'isFetchedAfterMount` and maybe I'm imaging a variable `isRefetchedAfterMount`?

tomzemp:
Thanks for the review (x2)!  I've addressed the lint + bug issue 🙏

> I do feel like some of the sync logic could be hoisted to "DataEntryField". Eg. we could have a `lastSyncedValue`-state there. Then we could update that value, instead of a "synced"-boolean. And compare `value` (from the store) with lastSyncedValue. Right now we're comparing the state in the field, but it's the data in the dataValueStore that is actually "the source of truth", right? And in theory these could diverge?

I think of this PR as a bug fix for the immediate issue of the forms being non-performant because of the react-final-form dependency (which we probably went with for convenience/familiarity/fact that it's called a data entry form even though it's not really a typical form), but it would be good to make a follow-up ticket (feature) to revisit some of the other logic to see if we can improve it further/leverage some of react-query's functionality for checking synced state.

I do wonder though on the specific issue of whether it's more ideal to indicate sync status based on comparison with the input state or the dataValueStore. I would actually think that insofar as the sync state is being recorded to provide a visual cue in the cell, we'd probably want it to reflect what the user is seeing in the cell? Like if for some reason cell shows `1`, dataValueStore has `2`, and the value that was sent to the server is `1`, we'd actually prefer to show that as synced? But this is a point I think it would be good to discuss / revisit the logic. I guess if you went offline you might trigger a resync based on the dataValueStore, but if you're online it seems you'll need to enter a value in the cell to get a POST to update the data value on the server? 🤔

src/data-workspace/form-wrapper.js
Birkbjo: Hmm... Yeah you may be right that this could cause a bug if there's an entry in the cache, and it's not refetched (eg. switching forms). But yeah I believe `isFetchedAfterMount` also applies to any refetches.
We might want to have a blocker another way. But we have to somehow prevent the form from rendering with cached (hydrated) values, before the initial fetch has completed.

Birkbjo (approved):

PR #426, opened 2024-11-29
fix: handle compulsory data element operands [DHIS2-17647]

tomzemp:
This PR addresses the referenced ticket ([DHIS2-17647](https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-17647)) and in general

(Note when I was testing this, I just made a couple of new data elements and assigned them to a new data set and then marked some as compulsory data element operands)

**Before:**
- no handling of compulsory data element operands

**After:**

- compulsory data element operands are marked with an `*` when they are empty. They then will get other applicable styling as values are entered (e.g. synced, error, warning)
image

- if the data set has `compulsoryFieldsCompleteOnly: true` ("Complete allowed only if compulsory fields are filled in" in the maintenance app), a snack bar message will appear saying that the form cannot be completed due to compulsory data element operands and the styling on the cells will be set to red/invalid/not saved (this styling will update as values are entered)
image

### Testing
- I've added automated tests on the complete-button to test that the checks preventing the completion based on the values/compulsoryDataElementOperands is working (i.e. this an integration level test which tests `useHasCompulsoryDataElementOperandsToFillOut` hook and logic preventing completion in `useOnCompleteCallback`)
- I have not added any automated tests for the styling (and hence implicit test of `useIsCompulsoryDataElementOperand` hook) partly because we didn't have existing tests and more because the InnerWrapper component where this can be tested is being refactored due to the performance PR (https://github.com/dhis2/aggregate-data-entry-app/pull/425) so the mocking would have to be redone if I added some now. I find these tests less crucial in any case because the styling doesn't block user from entering data.

I've also done some manual testing (see note above). 

[DHIS2-17647]: https://dhis2.atlassian.net/browse/DHIS2-17647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/bottom-bar/use-on-complete-callback.js
+++ b/src/bottom-bar/use-on-complete-callback.js
@@ -13,6 +13,7 @@ import {
     useSetFormCompletionMutationKey,
     validationResultsSidebarId,
     useHasCompulsoryDataElementOperandsToFillOut,
+    useEntryFormStore,
 } from '../shared/index.js'
 
 const validationFailedMessage = i18n.t(
@@ -122,7 +123,9 @@ export default function useOnCompleteCallback() {
     const { offline } = useConnectionStatus()
     const { data: metadata } = useMetadata()
     const [dataSetId] = useDataSetId()
-    const dataSet = selectors.getDataSetById(metadata, dataSetId)
+    const dataSet = dataSetId
+        ? selectors.getDataSetById(metadata, dataSetId)
+        : {}
     const { validCompleteOnly } = dataSet
     const { show: showErrorAlert } = useAlert((message) => message, {
         critical: true,
@@ -138,19 +141,24 @@ export default function useOnCompleteCallback() {
         useOnCompleteWithoutValidationClick()
     const hasCompulsoryDataElementOperandsToFillOut =
         useHasCompulsoryDataElementOperandsToFillOut()
+    const setCompleteAttempted = useEntryFormStore(
+        (state) => state.setCompleteAttempted
+    )
 
     return () => {
         let promise
-        // showErrorAlert('Complete compulsory data element operands')
-        // return Promise.resolve()
+
         if (hasCompulsoryDataElementOperandsToFillOut) {
+            setCompleteAttempted(true)
             cancelCompletionMutation()
-            showErrorAlert(
-                'Compulsory fields must be filled out before completing form'
+            promise = Promise.reject(
+                new Error(
+                    i18n.t(
+                        'Compulsory fields must be filled out before completing the form'
+                    )
+                )
             )
-            return Promise.resolve()
-        }
-        if (isLoading && offline) {
+        } else if (isLoading && offline) {
             cancelCompletionMutation()
             // No need to complete when the completion request
             // hasn't been sent yet due to being offline.
--- a/src/bottom-bar/use-on-complete-callback.js
+++ b/src/bottom-bar/use-on-complete-callback.js
@@ -12,6 +12,7 @@ import {
     useSetFormCompletionMutation,
     useSetFormCompletionMutationKey,
     validationResultsSidebarId,
+    useHasCompulsoryDataElementOperandsToFillOut,
 } from '../shared/index.js'
 
 const validationFailedMessage = i18n.t(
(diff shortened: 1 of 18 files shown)


src/data-workspace/category-combo-table-body/category-combo-table-body.test.js
```diff
@@ -453,6 +464,7 @@ const metadata = {
 
 describe('<CategoryComboTableBody />', () => {
     useMetadata.mockReturnValue({ data: metadata })
+    useDataSetId.mockReturnValue(['dataSet1'])
```
tomzemp, line 467: (FYI: edit was necessary because of thenew  call to the `useIsCompulsoryDataElementOperand` hook

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -135,11 +139,26 @@ export default function useOnCompleteCallback() {
...
+    const setCompleteAttempted = useEntryFormStore(
+        (state) => state.setCompleteAttempted
+    )
 
     return () => {
         let promise
 
-        if (isLoading && offline) {
+        if (hasCompulsoryDataElementOperandsToFillOut) {
+            setCompleteAttempted(true)
```
tomzemp, line 152: this is only set if there is a compulsory data element operand to fill out, so that invalid styling only gets applied if the compulsory data element operands need to be filled out (otherwise, they remain with just the `*`)

src/shared/stores/entry-form-store.js
```diff
@@ -4,6 +4,7 @@ import create from 'zustand'
 const inititalState = {
     errors: {},
     warnings: {},
+    completeAttempted: false,
```
tomzemp, line 7: I couldn't totally decide where to put this, but thought that since it was going to apply error styling to cells, I would put it in the store with the errors information.

src/shared/use-is-compulsory-data-element-operand.js
```diff
@@ -0,0 +1,57 @@
+import { useMemo } from 'react'
+import { useMetadata, selectors } from './metadata/index.js'
+import { useDataSetId } from './use-context-selection/use-context-selection.js'
+import { useDataValueSet } from './use-data-value-set/use-data-value-set.js'
+
+export const useIsCompulsoryDataElementOperand = ({
```
tomzemp, line 6: I colocated these hooks because I didn't really want to create a new folder in src/shared directory. I'd also vote for moving the miscellaneous hooks in src/shared into src/shared/hooks, and then I could create a folder and two separate hooks for these, but I didn't want to create two separate files with the the existing file structure.

src/shared/validation/use-imperative-validate.js
```diff
@@ -12,7 +12,7 @@ import {
  * @params {Object} options
  * @params {Boolean} options.enabled - Defaults to `true`
  **/
-export default function useImperativeValidate() {
+export const useImperativeValidate = () => {
```
tomzemp, line 15: this is not necessary FYI, but I was having some trouble with jest mocking when it was a default, so I changed to be a named export.

src/bottom-bar/use-on-complete-callback.js
```diff
@@ -135,11 +139,26 @@ export default function useOnCompleteCallback() {
...
+            cancelCompletionMutation()
```
flaminic, line 153: I think that doing ```       completeStatus: {
                    ...previousDataValueSet.completeStatus,
                    complete: !completed,
                },``` in cancelCompletionMutation messes up the complete status a bit. In you try to mark as complete successfully, then mark as incomplete successfully and then finally mark as complete unsuccessfully you will see the complete status = true even if the completion did fail. maybe we just want to set it as ```       completeStatus: {
                    ...previousDataValueSet.completeStatus,
                    complete: false,
                },``` ?

flaminic (approved):
Look good thanks @tomzemp . Just one comment on a weird behaviour i found for the complete status

PR #448, opened 2025-02-28
fix: add warning about noncalculable indicators [DHIS2-14221]

tomzemp:
See https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-14221, including notes.

This changes the presentation of indicators where the calculation is invalid. Previously, we showed either blank or '0' if the expression came back as invalid. This change updates to show

1. For expressions that cannot generally be calculated client-side (e.g. periodOffset):
  InfoIcon / Tooltip that says `This value cannot be calculated in this app`

2. For expressions that cannot be calculated because they evaluate to NaN/Infinity (e.g. 1/0):
WarningIcon / Tooltip that says `This expression is not mathematically calculable {{numeratorValue}}/{{denominatorValue}}`

**Testing**
- I've manually tested by creating an indicator with periodOffset (I think this is only out-of-the-box expression in maintenance app that presents problems in data entry app) and creating indicators that calculate to non-finite values (e.g. 1/0 because of missing data on form, or log10(0) / 1).
- I've added automated tests to cover these scenarios, cover the indicator cell updates. I've also updated the existing tests because I've changed the indicator calculations to send numeratorValue,denominatorValue back for display purposes.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2024-11-29T16:17:15.582Z\n"
-"PO-Revision-Date: 2024-11-29T16:17:15.582Z\n"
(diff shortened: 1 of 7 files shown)


flaminic (approved):
Thanks Tom. Looks good to me

PR #452, opened 2025-03-12
fix: when sanitizing text and target to links so they work w global s…

flaminic:
After the bug bash, we discovered an issue with the text users can add in the Maintenance app for dataset's sections, and datasets (titles, subtitles, pretext, etc.), and the global shell.
As a quick reminder,  currently, users can add some text to datasets, and within this text, they are allowed to include basic HTML elements (a, b, strong, u, em). These elements are then rendered accordingly in the Data Entry app.
The issue we’ve found is that if a user include a link to external URLs and has the the global shell enabled, these are not working.
After a discussion, we decided that the easiest solution would be to change the behaviour so that if a user adds a link in the text, we will programmatically modify it to open in a new tab by adding target="_blank" and rel="noopener" before rendering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
         "@tanstack/react-query-persist-client": "4.24.10",
         "chart.js": "3.9.1",
         "classnames": "2.3.2",
-        "dompurify": "^3.0.9",
+        "dompurify": "^3.2.4",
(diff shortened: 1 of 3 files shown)


kabaros (approved):

tomzemp (approved):
lgtm

PR #454, opened 2025-03-14
fix(transpose): fix transpose rendering of more than 2 categories

Birkbjo:
The `transpose`-logic for rendering the headers was a bit off - and with the comment asking for a refactor I took the opportunity to try to fix it.

The `span`-calculation was a bit optimistic, and did not take into account the number of options before it. This needs to be an `iterative` calculation - so I just reused the same logic as non-transposed headers. 

The logic for when to render the headers through `alreadyAdded` also does not work correctly with more than 2 categories - because you want the options to repeat for each disaggregation above it. After trying a few different things, it seems as simple as `(rowIndex % span === 0)`, which will render the header at the `rowIndex` it's currently not "spanning" for. 

Also fixed a bug when using `coc.categoryOptions`. The `categoryOptions`-list of a `categoryOptionCombo` is unorded because this is a set on the backend ([see](https://debug.dhis2.org/dev/api/schemas/categoryOptionCombo)). In the rest of the app we're always using `category.categoryOptions` which has the correct order - but I updated the selector to use this instead.

Also fixes the filter logic as per [BETA-355](https://dhis2.atlassian.net/browse/BETA-355)

Previous:
image

After:
image

[BETA-355]: https://dhis2.atlassian.net/browse/BETA-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


flaminic (approved):

kabaros (approved):

src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js
```diff
@@ -55,16 +56,49 @@ export const PivotedCategoryComboTableBody = React.memo(
...
+            return localSearch || globalSearch
+                ? (localSearch &&
+                      elementName.toLowerCase().includes(localSearch)) ||
+                      (globalSearch &&
+                          elementName.toLowerCase().includes(globalSearch))
+                : true
+        }
+
+        const filteredDataElements = dataElements.filter((de) =>
+            filterDataElementsBySearch(de.displayName)
```
Birkbjo, line 80:
```suggestion
            filterDataElementsBySearch(de.displayFormName)
```

src/data-workspace/filter-field.js
```diff
@@ -17,7 +17,7 @@ export default function FilterField({ value, setFilterText, formType }) {
                 dense
                 placeholder={
                     formType === FORM_TYPES.SECTION
-                        ? i18n.t('Filter fields in all sections')
+                        ? i18n.t('Filter data sets in all sections')
```
Birkbjo, line 20:
```suggestion
                        ? i18n.t('Filter data elements in all sections')
```

src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js
```diff
@@ -55,16 +56,49 @@ export const PivotedCategoryComboTableBody = React.memo(
...
+                filterText.toLowerCase()
+            return search ? elementName.toLowerCase().includes(search) : true
+        }
+
+        const filterDataElementsBySearch = (elementName) => {
+            const localSearch =
+                displayOptions.pivotMode === 'move_categories' &&
+                filterText.toLowerCase()
+            const globalSearch = globalFilterText.toLowerCase()
+            return localSearch || globalSearch
```
Birkbjo, line 71: Is the `ternary` needed here? return `localSearch && elementName.toLowerCase().includes(localSearch) || globalSearch && elementName.toLowerCase().includes(globalSearch)`?

This is also a case where I would add a comment to quickly describe the result
eg something like
`// local search should always work on rows - so disable localSearch when transposed (filterCOCBySearch would do the filter)

flaminic: mm i think that rewrite if localSearch and globalSearch are both empty strings will return false, while we want them to return true?

PR #456, opened 2025-03-17
Fix/filtering when transposed

flaminic:


src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.js
```diff
@@ -55,16 +55,35 @@ export const PivotedCategoryComboTableBody = React.memo(
             })
             .flat()
 
+        const filteredDataElements =
+            filterText.toLowerCase() === '' &&
+            globalFilterText.toLowerCase() === ''
```
Birkbjo, line 60: I would prefer to extract this to a function. Could keep it in component scope, but it would cleanup the nested and duplicated code here. 

eg. 

```js

        const filterDataElements = (de) => {
            const search = filterText.toLowerCase() || globalFilterText.toLowerCase()
            if(!search) {
                return true
            }
            return de.displayFormName.toLowerCase().includes(search)
        }
    ```


```
 
        const options = {
            metadata,
            categoryOptionsDetails,
            sortedCOCs,
            categories,
            dataElements: dataElements.filter(filterDataElements),
        }
```

PR #469, opened 2025-05-07
fix: number only totals [DHIS2-18640]

tomzemp:
See https://dhis2.atlassian.net/browse/DHIS2-18640.

This ticket removes row/column totals if data element is of a non-numeric type (such that the total cannot be calculated).

**Before**
image

**After**
image

_note: if there is a mix of numeric and non-numeric value types, the row totals (but not column totals) will display_
image

**Testing:**
tested manually plus added some manual tests

### DIFF
--- a/src/data-workspace/category-combo-table-body/category-combo-table-body-header.js
+++ b/src/data-workspace/category-combo-table-body/category-combo-table-body-header.js
@@ -55,6 +55,7 @@ export const CategoryComboTableBodyHeader = ({
     categoryOptionCombos,
     categories,
     checkTableActive,
+    nonNumberValueType,
 }) => {
     const { deId: activeDeId, cocId: activeCocId } = useActiveCell()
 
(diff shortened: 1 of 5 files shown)


src/data-workspace/custom-form/custom-form.module.css
```diff
@@ -1,5 +1,5 @@
 .customForm {
-    width: 100%;
+    inline-size: 100%;
```
tomzemp, line 2: these css updates are just formatting fixes (because of upgraded d2-style). They just got picked up when I ran `yarn format` but are not related to the ticket

src/data-workspace/category-combo-table-body/category-combo-table-body-header.js
```diff
@@ -55,6 +55,7 @@ export const CategoryComboTableBodyHeader = ({
     categoryOptionCombos,
     categories,
     checkTableActive,
+    allNonNumberValueType,
```
Birkbjo, line 58: Minor, but I don't think this is very descriptive as to what it does? Maybe it could just be called `showTotalHeader` or something similiar, that describes the functionality?

tomzemp: I've updated to very dry `hideColumnTotalsDueToNonNumberValueType` and `hideRowTotalsDueToNonNumberValueTypes`

PR #474, opened 2025-06-16
fix: use gist api on offline organisationUnits query

tomzemp:
see https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-19654

Long story short, our query for retrieving the organisation units that we prefetch is inefficient because having `includeDescendants` parameter retrieved every descendant, then filtered. Additionally, since we retrieved from api/organisationUnits, the backend was retrieving all fields related to the organisation units before returning back the ones that we specified in the `fields` parameter. This resulted for example in large amounts of data for org unit geography being retrieved then filtered, despite the fact that we did not want this, and as a result the back end crashed.

After talking with Jan, we decided to move to use the gist api with filtering by path (more efficient filtering and the SQL is simpler and won't include unnecessary fields like geography). Since we needed a modification to get organisation units without paging, there was a backend change [https://github.com/dhis2/dhis2-core/pull/21283, note: this PR has some other changes required for org unit tree rewrite]

Since the backend change is backported, but since the new format with gist/api is unusable, we need to feature toggle based on patch version (so there is now logic for that introduced in the app).

### testing
In addition to manual tests (it seemed to work) (testing against current debug/dev), I have added some automated tests for the feature toggling logic and to confirm that the new/old query format is used based on the result of the logical feature toggling for this feature.

[**NOTE:** This can be reviewed, but I'm keeping it in draft/don't want to mergre until BE team, e.g. Jan, confirms that the minor/patch versions I'm using here are correct]


flaminic (approved):

Birkbjo (approved):
The new params for orgUnits looks amazing!

LGTM!

tomzemp:
4 July: Based on ticket information and current release schedule, I updated the compatible version to 2.42.2 (from 2.42.1)

PR #483, opened 2025-08-05
fix: update indicator recalculation logic [DHIS2-19832]

tomzemp:
This is a follow up to https://dhis2.atlassian.net/browse/DHIS2-19832

The original ticket mentioned delays in totals calculation, but after fixing that, I noticed that there were also delays with indicator calculations in custom forms (I had my overly optimistic developer mindset of only paying attention to what I had been told was a bug  🙃

This PR updates the logic to check for changes in affected values relevant to an indicator, rather than relying on blurred field. As with the PR for totals (https://github.com/dhis2/aggregate-data-entry-app/pull/482), the issue appears to be an apparent race condition where dataValues state is updated after the blurred field is set, so when we use the data values associated with the blurred field we have stale data values (at least for custom forms).

### testing
did manual testing with both custom form (the one provided in the original ticket) and section form (I created an indicator that I added to Child Health form).
I've updated the existing automated tests to mimic an update to value returned from useDataValues and check that the calculation is occurring only when relevant value changes.

### DIFF
--- a/src/data-workspace/indicators-table-body/use-indicator-value.js
+++ b/src/data-workspace/indicators-table-body/use-indicator-value.js
@@ -18,7 +18,6 @@ import {
  * @returns {number} Indicator value
  */
 export const useIndicatorValue = ({
-    blurredField,
     denominator,
     factor,
     numerator,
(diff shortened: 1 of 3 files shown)


src/data-workspace/indicators-table-body/use-indicator-value.js
```diff
@@ -43,33 +42,35 @@ export const useIndicatorValue = ({
...
-        affectedDataElementsLookup,
-        blurredField,
-        denominator,
-        factor,
-        dataValues,
-        numerator,
-        decimals,
-    ])
+    }, [stringifiedValues, denominator, numerator, factor, decimals])
+    // dependency array is non-exhaustive insofar as dataValues is excluded
```
tomzemp, line 74: I went with getting a simple representation of the values and putting this in the dependency array (which makes the dependency array non-exhaustive). An alternative would be to stringify/parse a longer string representation of dataValues, but I feel like that's less efficient, so I don't think it's worth it to get rid of dependency array warning. I also don't want to turn off the eslint warning for the non-exhaustive dependencies, so we don't edit it and potentially get even less exhaustive in the future.

flaminic (approved):

PR #484, opened 2025-09-03
fix: default to all sections [DHIS2-19909]

tomzemp:
See https://dhis2.atlassian.net/browse/DHIS2-19909

This PR essentially reverses this work https://dhis2.atlassian.net/browse/DHIS2-18405. We had implemented a change to display first section by default as a workaround for rendering performance issues for large forms. We have now improved app performance by eliminating react-final-form dependency, so we should be able to safely go back to the old approach of showing "All sections" by default (https://dhis2.atlassian.net/browse/DHIS2-18373)

### DIFF
--- a/src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.jsx
+++ b/src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.jsx
@@ -28,7 +28,11 @@ export default function SectionFilterSelectorBarItem() {
 
     useEffect(() => {
         const sections = dataSet?.sections
-        if (sectionFilter === undefined && sections?.length) {
+        if (
+            dataSet?.renderAsTabs &&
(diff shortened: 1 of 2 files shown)


src/context-selection/section-filter-selector-bar-item/section-filter-selector-bar-item.jsx
```diff
@@ -67,7 +71,7 @@ export default function SectionFilterSelectorBarItem() {
             >
                 <MenuSelect
                     values={selectableOptions}
-                    selected={sectionFilter}
+                    selected={sectionFilter ?? null}
```
tomzemp, line 74: I've kept the undefined/null distinction for the useEffect dependency, but I'm mapping undefined to null here in order for the default "All sections" to show up as highlighted when nothing is selected.

PR #562, opened 2026-06-08
fix: update packages

tomzemp:
This updates packages. So this addresses:

- updates select bar (https://dhis2.atlassian.net/browse/DHIS2-21582). Note that this update means that tab functionality within the select bar item does not work. Henrik is creating a follow up ticket for this (according to Joe this is correct in terms of accessibility; the preference would be for items to be accessible with arrow keys within select bar item)
- removes patch-package of multi-calendar-dates (https://dhis2.atlassian.net/browse/DHIS2-21532): no longer necessary because underlying bug in chrome is fixed


package.json
```diff
@@ -24,9 +24,7 @@
         "cypress:prepare": "start-server-and-test 'yarn cypress:start' http://localhost:3000",
         "cypress:live": "yarn cypress:prepare 'yarn cypress:open:live'",
         "cypress:capture": "yarn cypress:prepare 'yarn cypress:run:capture'",
-        "cypress:stub": "yarn cypress:prepare 'yarn cypress:run:stub'",
-        "jgs:start": "BROWSER=none yarn start --proxy https://debug.dhis2.org/dev",
```
tomzemp, line 28: getting rid of this as Jan-Gerke had added it for dev

flaminic (approved):

PR #563, opened 2026-06-08
fix: replace standalone indicators custom forms [DHIS2-21607]

tomzemp:
Similar to https://github.com/dhis2/aggregate-data-entry-app/pull/471

We had logic for custom forms that replaced indicator cells if (and only if) they were inside `<td>` elements.

This PR updates the logic to replace standalone indicator cells as well.

### DIFF
--- a/src/data-workspace/custom-form/custom-form-total-cell.jsx
+++ b/src/data-workspace/custom-form/custom-form-total-cell.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types'
-import React, { useState, useEffect } from 'react'
+import React, { useMemo, useState, useEffect } from 'react'
 import { useValueStore } from '../../shared/index.js'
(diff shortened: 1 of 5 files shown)


src/data-workspace/custom-form/custom-form-total-cell.jsx
```diff
@@ -18,7 +18,10 @@ const computeTotalValues = (dataElementValues) => {
 
 export const CustomFormTotalCell = ({ dataElementId }) => {
     const dataValues = useValueStore((state) => state.getDataValues())
-    const dataElementValues = dataValues?.[dataElementId] || {}
+    const dataElementValues = useMemo(
```
tomzemp, line 21: this is not related, but is in the same directory, and the linter pointed this error out to me (this just guarantees the object reference is stable, but if it's defined, it wouldn't be an issue)

Sharmyn28 (approved):

PR #573, opened 2026-07-02
fix:  percentage decimals

flaminic:


kabaros (approved):

tomzemp (approved):
I still don't totally understand why we restricted percentages to between 0 - 100 😐, but this validation matches the backend 🙏

PR #576, opened 2026-07-14
fix: add description for indicators

tomzemp:
See https://dhis2.atlassian.net/browse/DHIS2-21674

This PR lets you highlight indicator cells and open the details panel for indicators. The only thing that will show in details panel for indicators is the displayFormName and description (if available) for the indicator, but this accomplishes the goal of having descriptions visible for indicators (in a way that is consistent with how we show description for data elements)

image


Chisomchima (approved):

Chisomchima (commented):
Thanks Tom, looks good! unfortunately, I couldn't see any description for ANC visits total, even when item?.isIndicator was true and item.description had a description.

PR #578, opened 2026-07-20
feat: totals for pivoted layout [DHIS2-19079]

tomzemp:
See https://dhis2.atlassian.net/browse/DHIS2-19079

Adds totals for pivoted sections (e.g. DEs/CoCs are pivoted, or a specific category is pivoted).

From child health (I've removed some DEs for ease of view)

_normal_
image

_pivoted_
image

_1 category pivoted_
image

Note: I noticed that there is a general issue when we pivot a category in a section but the data elements have different category combos: https://dhis2.atlassian.net/browse/DHIS2-21869. As such, I have not made the totals work for this case, but think we can address as part of the general fix to that bug.

### DIFF
--- a/src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.jsx
+++ b/src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.jsx
@@ -149,8 +149,8 @@ export const PivotedCategoryComboTableBody = React.memo(
                                                         'dataElement',
                                                 }
                                             )}
-                                            colSpan={fieldInRow.colSpan}
-                                            rowSpan={fieldInRow.rowSpan}
+                                            colSpan={fieldInRow.colSpan?.toString()}
+                                            rowSpan={fieldInRow.rowSpan?.toString()}
                                         >
                                             {fieldInRow.name !== 'default' &&
                                                 fieldInRow.displayFormName}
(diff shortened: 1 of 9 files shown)


src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.jsx
```diff
@@ -1,14 +1,70 @@
...
+
+const TotalRow = ({
+    dataElements,
+    categoryOptionCombos,
+    index,
+    pivotMode,
+    pivotedCategory,
+    categories,
+}) => {
+    const indexAdjustment = pivotMode === 'move_categories' ? 2 : 1
```
flaminic, line 26: I think for move_categories we need to get the number of cat in the cat combo no? instead of hardcoding it to 2?

src/data-workspace/category-combo-table-body-pivoted/category-combo-table-body-pivoted.jsx
```diff
@@ -179,9 +239,38 @@ export const PivotedCategoryComboTableBody = React.memo(
...
+                                    categories={categories}
+                                />
+                            )}
                         </TableRow>
                     )
                 })}
+                {renderColumnTotals && (
+                    <ColumnTotals
+                        paddingCells={paddingCells}
+                        initialColumns={categories.length}
```
flaminic, line 258: should this be dataElements.length if pivot mode is move_categories?

src/data-workspace/category-combo-table-body/total-cells.jsx
```diff
@@ -7,6 +7,23 @@ import styles from '../table-body.module.css'
 import { calculateColumnTotals, calculateRowTotal } from './calculate-totals.js'
 import { useValueMatrix } from './use-value-matrix.js'
 
+const defaultEmptyArray = []
+
+export const PaddingCell = ({ children, colSpan }) => (
```
flaminic: i think we never use this?

flaminic (approved):

flaminic:
Thanks @tomzemp looks great. A couple of comments but once those are addressed i think it can be merged

PR #579, opened 2026-07-28
fix: constrain option set dropdown menu width [DHIS2-13634]

arina-gutuboe:
See [DHIS2-13634](https://dhis2.atlassian.net/browse/DHIS2-13634)

This PR improves the readability of option set dropdowns by constraining the dropdown menu width.

**Changes**

- Updated @dhis2/ui from ^10.16.1 to ^10.17.0 to use the new menuMinWidth and menuMaxWidth props on the Select component.
- Set the option set dropdown menu width to:
   - Minimum: 350px
   - Maximum: 450px

[DHIS2-13634]: https://dhis2.atlassian.net/browse/DHIS2-13634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
         "@dhis2/app-runtime": "^3.14.5",
         "@dhis2/data-engine": "^3.17.0",
         "@dhis2/multi-calendar-dates": "^2.1.2",
-        "@dhis2/ui": "^10.16.1",
+        "@dhis2/ui": "^10.17.0",
         "@dhis2/ui-forms": "7.16.3",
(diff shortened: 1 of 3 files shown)


tomzemp (approved):
Thanks! 🙏

I think this looks like a good compromise. We end up with a pop-up menu that's a bit wide if the options are very short, but I assume this would not be much particularly annoying for users

Image

